PR #6, opened 2023-11-16
Fix Windows electron app build

nop33:


mvaivre (approved):
Great work! 👏 
The actions look good to me. I haven't tested the workflows, but you probably did ;)

.github/workflows/release-desktop.yml
```diff
@@ -53,12 +52,76 @@ jobs:
...
+          value: yarn@1
+
+      - name: Install NPM dependencies
+        run: yarn install --network-timeout 100000
+
+      - name: Compile shared package
+        run: cd packages/shared && yarn run compile
+
+      - name: Build & release Electron app
+        uses: samuelmeuli/action-electron-builder@v1
```
mvaivre, line 119: Should we use the action from `coparse-inc` instead?

Issue #123, opened 2023-11-23
[SHRD] Clean up API error parsing


nop33:
- [ ] 1. There is no point [exporting an `APIError`](https://github.com/alephium/alephium-frontend/blob/master/packages/shared/lib/errors.ts#L22-L29) since all API responses are converted and if there's an error it is [thrown as a string](https://github.com/alephium/alephium-web3/blob/master/packages/web3/src/api/utils.ts#L23-L25).
- [ ] 2. There's [no need for `isHTTPError`](https://github.com/alephium/alephium-frontend/blob/master/packages/shared/lib/errors.ts#L31) since (once again), the thrown API error will never have the `error` property.
- [ ] 3. I am not sure it makes sense to always [prefix the error message with the default error message](https://github.com/alephium/alephium-frontend/blob/master/packages/shared/lib/errors.ts#L39). In the following example, the default error message is "Error while building transaction"
    <img width="381" alt="image" src="https://github.com/alephium/alephium-frontend/assets/1579899/b01e778e-ea46-4cff-9d7d-0b1ea9beb875">

Issue #122, opened 2023-11-23
Do not close send modal when receiving an error "Not enough balance for fee, maybe transfer a smaller amount"


nop33:
This is not cool. Let's brainstorm:
- [x] 1. I think the error message gets hidden too fast. What about letting the user close it with an "X" button (or increasing the time it's visible by a lot so that the user has enough time to take a screenshot, or copy the error, etc)?
- [x] 2. The send modal should not close when a build tx error is received. It should give the chance to the user to update their inputs and try to build again.
- [x] 3. I don't like that we display the word "Error" 3 times!!! It's so scary, especially since it's inside a red box. How about making this a bit more "less scary"?
   1. The 1st "Error" word is on the [desktop wallet side](https://github.com/alephium/alephium-frontend/blob/master/apps/desktop-wallet/src/modals/SendModals/SendModal.tsx#L162)
   1. The 2nd comes from the [shared package](https://github.com/alephium/alephium-frontend/blob/master/packages/shared/lib/errors.ts#L36)
   1. The 3rd comes from the [web3 SDK](https://github.com/alephium/alephium-web3/blob/master/packages/web3/src/api/utils.ts#L25)
- [x] 4. If the error message is HUGE, it should not overflow, like it does here: https://github.com/alephium/alephium-frontend/issues/119

https://github.com/alephium/alephium-frontend/assets/1579899/ca7d3844-a15d-4cef-a924-ae06a71a3740

nop33:
Closing, since it was done.

Issue #119, opened 2023-11-21
[DW] Fix UTXO consolidation error handling.


nop33:
A user on Discord was looking for help because they received an error:

![image](https://github.com/alephium/alephium-frontend/assets/1579899/f5be00b3-77a4-4175-87db-3cb07fd5c4e1)

First of all, it is not acceptable to display errors like this, they provide 0 value. Thankfully, some of our Discord superhero helped the user to figure out their problem. The user needed to do a UTXO consolidation first.

Secondly, it seems that our error handling for UTXO consolidation does not work on v2.1.5.

nop33:
It appears that the consolidation modal is properly displayed.

I tested it using the following code. In summary:
1. I built 5 transaction, each containing 135 UTXOs with 0.01 ALPH in each
2. I sent all 5 transactions and waited until they got confirmed
3. Now, with 675 available to my address, I used the normal transfer send flow to send some funds to another address and I correctly got the popup

```ts
for (let i = 0; i < 5; i++) {
  const unsignedTxData = await client.node.transactions.postTransactionsBuild({
    fromPublicKey: address.publicKey,
    destinations: Array.from({ length: 135 }, () => ({
      address: '18xPARoYPLBGF7CPKZCaT6yyvfZ2hwujT3eY6haRhomvq',
      attoAlphAmount: '1000000000000000'
    }))
  })

  const signature = await signAndSendTransaction(address, unsignedTxData.txId, unsignedTxData.unsignedTx)
  console.log('signature', signature)
}
```

nop33:
This is the commit that fixes the bug: https://github.com/alephium/desktop-wallet/commit/7851994e8

It will be included in the next release.

nop33:
It's merged in master, so closing.

PR #124, opened 2023-11-23
Improve errors

nop33:
This relates to the following issues:
- #119 
- #122
- #123

### DIFF
--- a/apps/desktop-wallet/src/components/SnackbarManager.tsx
+++ b/apps/desktop-wallet/src/components/SnackbarManager.tsx
@@ -74,6 +74,7 @@ const SnackbarManagerContainer = styled.div`
   display: flex;
   justify-content: flex-end;
   z-index: 2;
+  max-height: 20vh;
 
   @media ${deviceBreakPoints.mobile} {
     justify-content: center;
@@ -90,6 +91,7 @@ const SnackbarPopup = styled(motion.div)`
   backdrop-filter: blur(10px);
   max-width: 800px;
   word-wrap: break-word;
+  overflow: scroll;
 
   &.alert {
     ${({ theme }) => getSnackbarStyling(theme.global.alert)}
--- a/apps/desktop-wallet/src/modals/SendModals/SendModal.tsx
+++ b/apps/desktop-wallet/src/modals/SendModals/SendModal.tsx
@@ -170,8 +170,6 @@ function SendModal<PT extends { fromAddress: Address }, T extends PT>({
 
           if (isRequestToApproveContractCall && onTransactionBuildFail) {
             onTransactionBuildFail(errorMessage)
-          } else {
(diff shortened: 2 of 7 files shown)


apps/desktop-wallet/src/components/SnackbarManager.tsx
```diff
@@ -90,6 +91,7 @@ const SnackbarPopup = styled(motion.div)`
   backdrop-filter: blur(10px);
   max-width: 800px;
   word-wrap: break-word;
+  overflow: scroll;
```
nop33, line 94: The above 2 changes make sure that when a very long error gets displayed, the height of the snackbar doesn't grow too much and that the content can be scrolled. Imporant that the content is scrolled to top so that when the user takes a screenshot, we can see the top of the error message.

https://github.com/alephium/alephium-frontend/assets/1579899/874e714b-6565-4c6d-b850-4befeb165d99

apps/desktop-wallet/src/modals/SendModals/SendModal.tsx
```diff
@@ -170,8 +170,6 @@ function SendModal<PT extends { fromAddress: Address }, T extends PT>({
 
           if (isRequestToApproveContractCall && onTransactionBuildFail) {
             onTransactionBuildFail(errorMessage)
-          } else {
-            onClose()
```
nop33, line 174: Addresses the main issue at #122

packages/shared/lib/errors.ts
```diff
@@ -16,25 +16,8 @@ You should have received a copy of the GNU Lesser General Public License
...
-  const stringifiedError = tryGetString(e)
-  const serverError = isHTTPError(e)
-    ? e.error.detail || `(${e.status}${e.statusText && `: ${e.statusText}`}) ${e.error.message || ''}`
-    : stringifiedError
-
-  return `${defaultErrorMessage}${serverError && `: ${serverError}`}`
-}
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const tryGetString = (obj: any): string | null =>
+  typeof obj.toString === 'function' ? obj.toString().replace('Error: [API Error] - ', '') : null
```
nop33, line 23: I thought I'd clean up the error message before displaying to the user, WDYT?

nop33: | Before | After |
| ----- | ----- |
| image | image |

mvaivre: Nice, but what about naming the function `tryGetErrorString`?

nop33: I simplified it even further.

nop33: Ready for another review :)

mvaivre (approved):
Nice! Tight!

PR #125, opened 2023-11-23
Register WC deep link schema, show toast message on success, update WC project ID

nop33:


apps/mobile-wallet/src/contexts/walletConnect/WalletConnectContext.tsx
```diff
@@ -606,10 +610,11 @@ export const WalletConnectContextProvider = ({ children }: { children: ReactNode
 
       setSessionProposalEvent(undefined)
     } catch (e) {
-      console.error('❌ WC: Error while approving and acknowledging', e)
+      console.error('❌ WC: Error while rejecting', e)
     } finally {
       setLoading('')
       setIsSessionProposalModalOpen(false)
+      showToast('👍 Rejected, you can go back to your browser.')
```
mvaivre, line 617: I find it a bit weird to use a thumb up emoji when rejecting a proposal. What about using 🙅 or🫸 instead? :)

apps/mobile-wallet/src/utils/layout.ts
```diff
@@ -36,10 +36,16 @@ export const scrollScreenTo = (
   }
 }
 
+export enum ToastDuration {
```
mvaivre, line 39: 👍

apps/mobile-wallet/src/contexts/walletConnect/WalletConnectContext.tsx
nop33: haha! I don't know. I don't like any of them because the ones you proposed look as if there is a problem. But I get your point. Maybe we just remove the emoji completely. WDYT?

nop33: Or maybe remove all "Reject/Approved" and just say "You can now go back to your browser".

mvaivre (approved):

PR #131, opened 2023-11-27
Use tanstack to fetch alph price

mvaivre:


apps/explorer/src/api/assets/assetsApi.ts
```diff
@@ -109,5 +110,17 @@ export const assetsQueries = createQueriesCollection({
...
+    assetPrice: (coinGeckoTokenId: string, currency = 'usd') => ({
+      queryKey: ['assetPrice', coinGeckoTokenId, currency],
+      queryFn: async (): Promise<number> => {
+        const res = (await (
+          await fetch(`https://api.coingecko.com/api/v3/simple/price?ids=${coinGeckoTokenId}&vs_currencies=${currency}`)
+        ).json()) as AssetPriceResponse
+
+        return res[coinGeckoTokenId][currency]
+      }
+    })
```
nop33, line 124: Interesting to see the implementation using tanstack! Just FYI, here's how it'd be done with rtkQuery: https://github.com/alephium/alephium-frontend/blob/master/apps/desktop-wallet/src/storage/assets/priceApiSlice.ts#L37-L47

I like how you made sure that we can query other tokens apart from ALPH.

apps/explorer/src/types/assets.ts
```diff
@@ -44,6 +44,8 @@ export type NFTFile = {
   image?: string
 }
 
+export type AssetPriceResponse = { [tokenId: string]: { [currency: string]: number } }
```
nop33, line 47: This could be added in the shared package, so that we can reuse it in our wallets!

nop33 (commented):
Nice stuff!

Let's clean up the unused `apps/explorer/src/api/priceApi.ts` file, we don't need it anymore, right?

Also, could you please add a changeset? 🙏

mvaivre:
> Let's clean up the unused apps/explorer/src/api/priceApi.ts file, we don't need it anymore, right?
It was already deleted if I'm not mistaken.

apps/explorer/src/types/assets.ts
mvaivre: Let's deal with moving all the types in another PR?

mvaivre: Cf. #134

nop33 (commented):
Nice!

I can see the network request to coingeko succeeded (status code 200), but I don't see the fiat price being calculated when I click on a random address. Why could that be? Is it working for you?

image

http://localhost:5173/addresses/1EyW3cXwYeP9BZd4AYngiPfXHFqzUziHmYaq9sBAMv68h

nop33 (approved):

PR #152, opened 2023-12-06
Replace comma with full-stop in amount input

### LINKED ISSUE #151: Bug - Not able to enter decimals
**Describe the bug**
Mobile Wallet, IOS Iphone 10. In some IOS versions or most like certain language set-up,  the keyboard to input the amount when sending assets is a "," and not a "." as a result it is not possible to specify a non-integer number of coins to send. When you click on the comma, the input field is cleared. Could this be because the keyboard suggests a comma instead of a period? Attached a video with the bug.

**To Reproduce**
I'm assuming this is language specific, so one would need to set his iphone in the language in question (or any language that results the keyboard to show "," instead of "." 
Asking the user for his language set-up and IOS version.

**Expected behavior**
User expected to be able to enter decimals.

**Screenshots**

Uploading RPReplay_Final1701682546.mp4…

**Environment**

- OS: IOS 
- Iphone 10
- Version [e.g. 1.0.1]

nop33:
Closes #151

### DIFF
--- a/.changeset/mean-knives-bake.md
+++ b/.changeset/mean-knives-bake.md
@@ -1,0 +1,5 @@
+---
+'@alephium/mobile-wallet': patch
+---
+
+Fix amount input when numerical keyboard has only comma available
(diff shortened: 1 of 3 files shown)


mvaivre (approved):
Now that we have a nice shared lib, we should definitely add some common tests for cases like this one. WDYT?

nop33:
> Now that we have a nice shared lib, we should definitely add some common tests for cases like this one. WDYT?

https://github.com/alephium/alephium-frontend/issues/70

Issue #128, opened 2023-11-27
Separate groups of thousands in amount display based on region


nop33:
A user reported:

> those of us in the US do not use ' as the 1000 separator, we use a comma as in 10,000 and not 10'000 can we make this an option as it's totally unnatural to see it like 10'000 in the US thanks much

nop33:
This is interesting: https://byby.dev/js-format-numbers-commas

PR #183, opened 2023-12-13
Use locale number separators

nop33:
Related to #128

Here is an initial approach to this problem. This [article](https://byby.dev/js-format-numbers-commas) was insightful.

We need to write tests.


nop33:
Superseded by https://github.com/alephium/alephium-frontend/pull/383. Closing

PR #184, opened 2023-12-13
Allow searching by address public key in explorer

nop33:
@polto will be happy ✨

### DIFF
--- a/apps/explorer/src/components/SearchBar.tsx
+++ b/apps/explorer/src/components/SearchBar.tsx
@@ -16,6 +16,8 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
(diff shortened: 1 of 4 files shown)


apps/explorer/src/components/SearchBar.tsx
```diff
@@ -84,11 +86,19 @@ const SearchBar = ({ className }: SearchBarProps) => {
         } else {
           redirect(`/transactions/${word}`)
         }
+      } else if (word.length === 66) {
```
nop33, line 89: I checked at least 10 addresses and they all had public keys of `66` characters. I assume that's the standard length.

polarker:
Can we also support searching by contract id as well? Some community members would be happy as I was asked for this feature a while back )

mvaivre (approved):
👍

PR #205, opened 2023-12-19
Fix iOS WC crash, show version, capture WC client init error

nop33:
I wanted to see some analytics of how many installations we have in each version of our wallet, basically group by version number. Even though we track the dw version in every event, we don't capture it as a person property. This PR fixes that.

~~I am also addressing the issue at #206~~

### DIFF
--- a/.changeset/green-cheetahs-press.md
+++ b/.changeset/green-cheetahs-press.md
@@ -1,0 +1,5 @@
+---
+'@alephium/mobile-wallet': patch
+---
+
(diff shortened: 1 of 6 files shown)


nop33:
I've used the https://docs.expo.dev/versions/latest/sdk/application package to add the version:

image

image

nop33:
Superseded by https://github.com/alephium/alephium-frontend/pull/220

Closing.

Issue #215, opened 2023-12-20
Send modal "Your addresses" doesn't list all addresses of the wallet


nop33:
<img width="1168" alt="image" src="https://github.com/alephium/alephium-frontend/assets/1579899/020f9159-8730-4878-addb-a56bf7d9e6b6">

<img width="1157" alt="image" src="https://github.com/alephium/alephium-frontend/assets/1579899/6aecb3b1-4825-4685-864b-714f7ca42203">

PR #220, opened 2023-12-20
Fix MW & DW releases

nop33:
This PR attempts to solve the following issues:

- iOS crash due to WalletConnect: https://github.com/alephium/alephium-frontend/issues/213
  - If for some reason an exception is thrown while initializing the WC client, we enter an infinite loop. This PR introduces a mechanism for retrying to initialize every 3 seconds instead. Hopefully this was it.
- Fixes https://github.com/alephium/alephium-frontend/issues/215
- Fixes https://github.com/alephium/alephium-frontend/issues/219
- Changes WC project ID, maybe related to https://github.com/alephium/alephium-frontend/issues/221?
- Improves WC client initialization in DW to better understand https://github.com/alephium/alephium-frontend/issues/221
- Shows version number in settings screen: https://github.com/alephium/alephium-frontend/issues/211
   > I've used the https://docs.expo.dev/versions/latest/sdk/application package to add the version:
   > 
   > | iOS | Android |
   > | ----- | ----- |
   > | image | image |

### DIFF
--- a/apps/desktop-wallet/src/contexts/walletconnect.tsx
+++ b/apps/desktop-wallet/src/contexts/walletconnect.tsx
@@ -113,7 +113,7 @@ export const WalletConnectContextProvider: FC = ({ children }) => {
         metadata: {
           name: 'Alephium desktop wallet',
           description: 'Alephium desktop wallet',
-          url: 'https://github.com/alephium/desktop-wallet',
+          url: 'https://github.com/alephium/alephium-frontend',
           icons: ['https://alephium.org/favicon-32x32.png']
         }
       })
--- a/apps/desktop-wallet/src/contexts/walletconnect.tsx
+++ b/apps/desktop-wallet/src/contexts/walletconnect.tsx
@@ -126,7 +126,7 @@ export const WalletConnectContextProvider: FC = ({ children }) => {
       setWalletConnectClientStatus('initialized')
       setActiveSessions(getActiveWalletConnectSessions(client))
     } catch (e) {
-      setWalletConnectClientStatus('initialization-failed')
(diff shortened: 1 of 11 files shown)


apps/desktop-wallet/src/contexts/walletconnect.tsx
```diff
@@ -406,10 +414,13 @@ export const WalletConnectContextProvider: FC = ({ children }) => {
     console.log('👉 ARGS:', args)
   }, [])
 
+  const shouldInitialize = walletConnectClientStatus === 'initialization-failed'
+  useInterval(initializeWalletConnectClient, 3000, !shouldInitialize)
```
polarker, line 418: Same here. This seems not used.

nop33: I simplified the init code to only call `initializeWalletConnectClient` from one place.

I'll work on now to add a 10 retry limit.

apps/mobile-wallet/src/contexts/walletConnect/WalletConnectContext.tsx
```diff
@@ -413,6 +444,9 @@ export const WalletConnectContextProvider = ({ children }: { children: ReactNode
     console.log('👉 ARGS:', args)
   }, [])
 
+  const shouldInitialize = walletConnectClientStatus === 'initialization-failed'
+  useInterval(initializeWalletConnectClient, 3000, !shouldInitialize)
```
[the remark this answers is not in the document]
nop33: Nope

mvaivre (approved):
LGTM

PR #226, opened 2023-12-22
Fix wc

polarker:


Lbqds:
We should put the clean storage code into the `shared` package, but this error will appear when running Android:

```
Android Bundling complete 3493ms
 ERROR  ReferenceError: Property 'TextEncoder' doesn't exist, js engine: hermes
 ERROR  Invariant Violation: "main" has not been registered. This can happen if:
* Metro (the local dev server) is run from the wrong folder. Check if Metro is running, stop it and restart it in the current project.
* A module failed to load due to an error and `AppRegistry.registerComponent` wasn't called., js engine: hermes
```

So I copied the code twice.

polarker:
@mvaivre @nop33 We have added all of the required changes to fix all of the reproducible WC issues.

Please take over the PR and merge it at your convenience. Note this comment by @Lbqds : https://github.com/alephium/alephium-frontend/pull/226#issuecomment-1867668187

nop33 (commented):
Code LGTM ✅ 

- [x] iOS simulator works
- [x] iOS internal build works on real device
- [x] Android simulator works
- [ ] Windows VM works

I am getting errors when I try to pair with ayin.app

image

mvaivre:
I don't see the error mentioned above and it works for me, but it may be because I don't have the same "trash" in my WC local storage.

nop33:
I cleaned all things related to WC from the local storage and index db, same problem

nop33 (approved):
I also tested with an Android physical device, I could do a swap with ayin.app on mainnet.

and as we discussed on Discord, the Windows seems to be an isolated case in my VM.

Issue #188, opened 2023-12-14
Review RC and release workflow with `changeset`


nop33:
We need to revisit our workflow of building RC's using `bunx changeset`. A few issues I noticed:

1. I cannot easily only run `bunx changeset version` for one specific app. There is the `--ignore` parameter which allows me to ignore each other package/app individually, but if a changeset affects all packages/apps then it doesn't work.
https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#version
2. Similar issue with `bunx changset tag`, but that one doesn't even have an `--ignore` parameter. We can easily delete generated tags with `git tag -d [TAG_NAME]`, however.
https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#tag
3. I noticed that `bunx changeset version` will replace the `"*"` version of the `@alephium/shared` dependency in our apps with the actual package version. We don't want that.

PR #232, opened 2024-01-04
New price endpoints, better API mgmt and NFT display (superseded)

mvaivre:
⚠️ Need new web3 release cf. https://github.com/alephium/alephium-web3/pull/295

Simple replacement of endpoint for fetch price.

I'll add up all token prices in a follow up PR, once the explorer backend will be updated with endpoints based on token IDs instead of symbols.


nop33 (commented):
Code looks good ✅ I'll do the functional tests as soon as web3 is released.

.changeset/four-otters-study.md
nop33, line 1: Because of https://github.com/alephium/alephium-frontend/issues/188 it's better to not include multiple apps/packages in the same changeset, otherwise it gets ugly when trying to release the apps/packages separately. It works better to have 1 changeset per app/package.

nop33:
Superseded by https://github.com/alephium/alephium-frontend/pull/324. Feel free to close.

PR #233, opened 2024-01-04
Fix AssetBadge decimals

### LINKED ISSUE #230: [DW] Wrong decimal display for token-list assets (in AssetBadge)
The assets' value seems to be wrongly displayed for assets with `decimals` equal to 0.

mvaivre:
Fixes #230

### DIFF
--- a/apps/desktop-wallet/src/components/AssetBadge.tsx
+++ b/apps/desktop-wallet/src/components/AssetBadge.tsx
@@ -53,7 +53,7 @@ const AssetBadge = ({ assetId, amount, simple, className }: AssetBadgeProps) => 
       {nftInfo?.name ? (
         <AssetSymbol>{nftInfo?.name}</AssetSymbol>
       ) : amount !== undefined ? (
-        <Amount value={amount} suffix={assetInfo?.symbol} />
+        <Amount value={amount} suffix={assetInfo?.symbol} decimals={assetInfo?.decimals} />
       ) : (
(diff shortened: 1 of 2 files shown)


apps/desktop-wallet/src/components/AssetBadge.tsx
```diff
@@ -53,7 +53,7 @@ const AssetBadge = ({ assetId, amount, simple, className }: AssetBadgeProps) =>
       {nftInfo?.name ? (
         <AssetSymbol>{nftInfo?.name}</AssetSymbol>
       ) : amount !== undefined ? (
-        <Amount value={amount} suffix={assetInfo?.symbol} />
+        <Amount value={amount} suffix={assetInfo?.symbol} decimals={assetInfo?.decimals} />
```
polarker, line 56: Optional parameters backfire quite often. Could use explicit parameters for important ones, in this case, we could introduce `<AssetAmount mandatoryDecimals=.../>`

polarker (approved):
The fix LGTM 👍 Just a remark on coding practice

apps/desktop-wallet/src/components/AssetBadge.tsx
mvaivre: Agreed. The `Amount` component is due for a cleanup. I think we can do better with clever types. Just an idea :
- Expecting a mandatory amount type prop to be either `fiat` or `asset`
- If it's asset, then the `decimals` field is mandatory.

nop33 (approved):

PR #256, opened 2024-01-09
Detect biometrics change and fallback to pin

### LINKED ISSUE #236: Cannot find wallet after enabling Face ID on Android

nop33:
Closes #236


mvaivre (approved):
Code looks good! 
Will test on iOS as soon as I can add a new face to my device ;)

Issue #240, opened 2024-01-08
macOS & Windows crash after login in v2.2.0


nop33:
At least 3 users report this on Discord:

> Once logged in, nothing clickable and after some seconds all the app becomes white (also after reboot and osx update - Sonoma 14.2.1). As stated on the download page, the integrated update wasn't working due to certificates changes. So I downloaded and replaced the app. The wallet works to log in, but after that, as soon as I click a button of the interface or the wallet makes some update in background, it freezes 1 second and all the app becomes white.
>
> I made the test to install and import the wallet in another user profile (another session on the computer) and the problem didn't occur.
So I think that the problem is not about the system <-> app integration, but more something like v2.2.0 crashing in interpreting earlier versions' user data 

> All I get is white screen of death when using new version on Mac after login- is there any reason I cannot continue to use the previous version as it still works

Another reported it for Windows:

> v2.2.0 freezes directly after password entry and never recovers. Force quit required to end operation.
> Windows 10

nop33:
Merged to master, closing.

PR #257, opened 2024-01-10
Add clear cache button in general settings

nop33:
Related to #240

~~We still need to hook the button with a reset function.~~


mvaivre (approved):
LGTM

PR #283, opened 2024-01-18
Sign messages through WalletConnect

nop33:
Also closes https://github.com/alephium/alephium-frontend/issues/286 through 9d2eea5


apps/desktop-wallet/scripts/syncTranslationFiles.js
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env node
+
```
nop33, line 2: For some reason now this seems to be needed otherwise I can't run the script

apps/desktop-wallet/src/contexts/walletconnect.tsx
```diff
@@ -237,6 +249,9 @@ export const WalletConnectContextProvider: FC = ({ children }) => {
         params: { request }
       } = event
 
+      console.log('📣 RECEIVED EVENT TO PROCESS A SESSION REQUEST FROM THE DAPP.')
+      console.log('👉 REQUESTED METHOD:', request.method)
+
```
nop33, line 254: Intended to keep 😉

nop33 (commented):
I have self-reviewed my code (since @mvaivre is on holidays) and I have tested it by running the desktop wallet and the mobile wallet in both Android and iOS simulators. 

Also, our Discord user who was interested in this feature has used the open beta to test it and gave positive feedback:

> Works great, thank you 🫡

I'll merge this.

PR #285, opened 2024-01-18
EAS GitHub workflows

### LINKED ISSUE #190: Create EAS build/submit CI workflows for mobile wallet
Get inspired by:
- https://github.com/byCedric/expo-monorepo-example/blob/main/.github/actions/setup-monorepo/action.yml
- https://github.com/byCedric/expo-monorepo-example/blob/main/.github/workflows/build.yml

Useful:
- https://docs.expo.dev/build/building-on-ci/

nop33:
Closes #190

We can now trigger EAS builds from our CI!

Once this PR is merged, the following options will be available in our GitHub Actions:

- `android`: This triggers an `eas build` and `submit` of a draft internal release. The release can then be later promoted to an open-beta and production through the Google Play Console website. No need to build another one
- `ios-registered-devices`: This triggers an `eas build` for the registered iOS devices. Once the build passes, a URL will be printed in the logs that the registered iOS devices can open to get the new version.
- `ios-testflight`: This is similar to `android` but for iOS: It triggers an `eas build` and `submit`. The TestFlight version can be promoted to a production release through the Apple Store Connect website.

### DIFF
--- a/.github/actions/setup-monorepo/action.yml
+++ b/.github/actions/setup-monorepo/action.yml
@@ -1,0 +1,54 @@
+name: Setup Monorepo
+description: Prepare and install everything for the monorepo
+
+inputs:
+  node-version:
+    description: Version of Node to use
+    default: 18.x
+
+  pnpm-version:
+    description: Version of pnpm to use
+    default: 8.x
+
+  eas-version:
+    description: Version of EAS CLI to use
+    default: latest
+
+  expo-token:
+    description: Expo token to authenticate with
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - name: 🏗 Setup pnpm
+      uses: pnpm/action-setup@v2
+      with:
+        version: ${{ inputs.pnpm-version }}
+
+    - name: 🏗 Setup Node
+      uses: actions/setup-node@v3
+      with:
+        node-version: ${{ inputs.node-version }}
+        cache: pnpm
+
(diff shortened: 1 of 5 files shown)


.github/actions/setup-monorepo/action.yml
```diff
@@ -0,0 +1,54 @@
...
+      uses: expo/expo-github-action@v8
+      with:
+        eas-version: ${{ inputs.eas-version }}
+        token: ${{ inputs.expo-token }}
+
+    - name: 📦 Install dependencies
+      run: pnpm install
+      shell: bash
+
+    - name: ♻️ Restore cache
```
mvaivre, line 45: Noob question: is it normal that we restore the cache *after* the other steps? Also, should we use `cache@v4`?

nop33: The cache we are talking about is the Turbo cache (see the paths that are being restored below. So it doesn't matter where it is. I consulted this: https://github.com/byCedric/expo-monorepo-example/blob/main/.github/actions/setup-monorepo/action.yml

v4 uses node 20, I would say we go for v3 now and update our whole stack slowly to v20 (we still use v16 in some cases: https://github.com/alephium/alephium-frontend/issues/343

mvaivre: Alrighty then! Let's try it live for test builds :)

mvaivre (approved):

PR #293, opened 2024-01-23
Fix list of suggested words in import screen of desktop wallet

nop33:
A bug has been found in the library we depend upon. A fix was implemented as well and a PR was opened. I am patching the library with the same fix, until the PR gets merged and a new version gets released.

https://github.com/yairEO/tagify/pull/1285


nop33 (commented):
I have self-reviewed my code (since @mvaivre is on holidays) and I have tested it by running the desktop wallet. I'll merge this to proceed with an RC.

Issue #193, opened 2023-12-14
Mobile wallet does not respond to WC requests when inactive (DEX shows 0 balances)


nop33:
It seems that the mobile wallet *sometimes* cannot process the WC session requests when it's inactive, in the background. This means that when the dApp sends the `alph_requestNodeApi` or `alph_requestExplorerApi` request, it does not receive a response.

nop33:
Let's have a look into https://github.com/Rapsssito/react-native-background-actions

nop33:
- [Metamask uses `react-native-background-timer`](https://github.com/MetaMask/metamask-mobile/blob/main/app/core/LockManagerService/index.ts#L7) but for a different reason (locking app)
- Rainbow doesn't use anything background-related.

nop33:
The callback of the `session_request` WC event is this: https://github.com/alephium/alephium-frontend/blob/master/apps/mobile-wallet/src/contexts/walletConnect/WalletConnectContext.tsx#L147-L355

We register a listener to the `session_request` event in this `useEffect`: https://github.com/alephium/alephium-frontend/blob/master/apps/mobile-wallet/src/contexts/walletConnect/WalletConnectContext.tsx#L417

We also unsubscribe the listener in the same `useEffect`: https://github.com/alephium/alephium-frontend/blob/master/apps/mobile-wallet/src/contexts/walletConnect/WalletConnectContext.tsx#L428

@mika, this made me think: What if the unsubscription happens when the app becomes inactive, leading to our app not listening to such events anymore? It seemed like a very plausible theory. But I cannot prove this on the Android simulator. The component seems to remain mounted, so the `useEffect` function does not execute, so we should still be listening to WC events.

nop33:
Another interesting find is this:

> We allow developers to take full control of the WebSocket connection by specifying the ConnectionType.MANUAL and ConnectionType.AUTOMATIC connection type. The default type(ConnectionType.AUTOMATIC) disconnects the web-socket connection when app enters the background and reconnects when app is brought back to the foreground. The ConnectionType.MANUAL allows developers to control when to open the WebSocket connection and when to close it. This can be done using CoreClient.Relay.connect() and CoreClient.Relay.close(), respectively
https://medium.com/walletconnect/beginner-guide-to-walletconnect-v2-0-sign-protocol-for-android-developers-936293e30700

but I can't find a `connectionType` prop in the SDK...

nop33:
I've asked the WalletConnect Discord a few times:
[...]

PR #295, opened 2024-01-23
Respond to WalletConnect while app in background

nop33:
Related to #193 

So, it seems that simply having a timer to keep the app active while in the background prevents React Native from going to "sleep" mode, and all WalletConnect requests seem to receive responses! I've tested this on Android simulator and physical device as well as iOS simulator and it works! I look forward to seeing if it also works on an iOS physical device.

### DIFF
--- a/.changeset/clean-bikes-thank.md
+++ b/.changeset/clean-bikes-thank.md
@@ -1,5 +1,0 @@
----
-"@alephium/mobile-wallet": patch
----
-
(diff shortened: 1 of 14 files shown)


nop33:
I am creating internal releases on both platforms to test this with our internal testers.

nop33 (commented):
Self-reviewed since @mvaivre's on holiday. Merging.

PR #316, opened 2024-01-26
Handle WC request when wallet locks

### LINKED ISSUE #313: Unknown signer address: Desktop wallet responds to WC requests with error after it locks
**Describe the bug**
When the desktop wallet app locks, it clears all data. We chose this approach since the app can support multiple wallets. In the mobile wallet we follow a different approach where when you lock the app, we only clear the sensitive info (like the mnemonic) but keep all the rest (addresses, transactions, etc), so that when you unlock everything is already there.

The (reproducible) scenario is that if the wallet has a short lock time and the user has connected their desktop wallet to a dApp, the wallet will respond with an error when the dApp tries to request something from the wallet and the wallet has locked.

**To Reproduce**
Steps to reproduce:

1. Select 2 min lock time in settings
2. Open ayin.app
3. Connect with the desktop wallet
4. Wait 2 minutes
5. Try to do a swap
6. You'll get the error _"Error: Error: Unknown signer address: [YOUR ADDRESS]"_

**Expected behavior**
The wallet should not respond with an error. It should come to the foreground and ask the user to unlock, just like it happens when you try to pair with a dApp for the first time. After unlock, it should respond to the dApp with the request it asked for. To achieve this, we could follow a similar approach like the mobile wallet: not deleting all data on lock. Maybe we can just delete the sensitive data and then delete all data if the user wants to unlock a different one than the last one.

nop33:
Closes #313

### DIFF
--- a/.changeset/hungry-bats-smell.md
+++ b/.changeset/hungry-bats-smell.md
@@ -1,0 +1,5 @@
+---
+"alephium-desktop-wallet": patch
+---
+
+Handle WalletConnect requests when app locks
--- a/apps/desktop-wallet/locales/bg-BG/translation.json
+++ b/apps/desktop-wallet/locales/bg-BG/translation.json
@@ -477,5 +477,6 @@
   "Transaction ID": "Transaction ID",
   "Unsigned transaction": "Unsigned transaction",
   "Sign Message": "Sign Message",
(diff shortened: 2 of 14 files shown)


mvaivre (approved):
LGTM 👌

Issue #243, opened 2024-01-08
Display fiat prices of tokens in desktop wallet


nop33:
## Address and wallet total balances
- [x] Fetch latest price of ALPH from explorer backend instead of CoinGecko
- [x] Fetch latest price of every other verified token from explorer backend

## Chart
- [x] Fetch ALPH historical balances of each address from explorer backend
- ~~Fetch token historical balances of each address from explorer backend~~ https://github.com/alephium/alephium-frontend/issues/426
- [x] Fetch ALPH historical prices from explorer backend instead of CoinGecko
- [x] Fetch historical prices of each token from explorer backend

PR #109, opened 2023-11-17
Fetch token prices & history using new Explorer BE endpoints

mvaivre:
**⚠️ Is waiting for a new web3 release.** cf https://github.com/alephium/alephium-web3/pull/295

Related to https://github.com/alephium/alephium-frontend/issues/88

PR #322, opened 2024-01-29
Fetch token prices & history using new Explorer BE endpoints (rebased and fixed)

nop33:
This is @mvaivre's [initial PR](https://github.com/alephium/alephium-frontend/pull/109), rebased.

Fixes most subtasks of https://github.com/alephium/alephium-frontend/issues/243.

@nop33's subtasks:
- [x] Rebase to latest master
- [x] Use latest web3
- [x] Update code to use latest explorer API
- [x] Fix: Loading indication of total worth and chart is buggy
- [x] Fix: Switching currency does not update chart
- [x] Re-implement price polling
- [x] Fix greeting message animation
- [x] Compare loading speed with `master` and improve if needed


apps/desktop-wallet/src/App.tsx
```diff
@@ -189,26 +194,76 @@ const App = () => {
...
+      }
+    }
+  }, [
+    dispatch,
+    fungibleTokens.status,
+    isLoadingVerifiedFungibleTokens,
+    network.status,
+    settings.fiatCurrency,
+    verifiedFungibleTokenSymbols.uninitialized
+  ])
```
nop33, line 237: @mvaivre I think it makes sense to only fetch prices for **verified** fungible tokens, not just tokens that have a **symbol**. Don't you agree?

nop33: The important use-cases for this `useEffect` is that it will run:
1. when the **network status** changes to `online` (when for example the network settings change)
1. when **tokens are discovered** in addresses data that belong to the verified token list and we have not yet fetched the historic prices for
1. when the **currency settings** are changed

apps/desktop-wallet/src/App.tsx
```diff
@@ -189,26 +194,76 @@ const App = () => {
...
+      network.status === 'online' &&
+      !isLoadingVerifiedFungibleTokens &&
+      verifiedFungibleTokenSymbols.uninitialized.length > 1
+    ) {
+      console.log(
+        'TODO: Sync address verified tokens balance histories for',
+        verifiedFungibleTokenSymbols.uninitialized.filter((symbol) => symbol !== ALPH.symbol)
+      )
+    }
+  }, [isLoadingVerifiedFungibleTokens, network.status, verifiedFungibleTokenSymbols.uninitialized])
```
nop33, line 250: This is a placeholder.

apps/desktop-wallet/src/App.tsx
```diff
@@ -189,26 +194,76 @@ const App = () => {
...
+        currency: settings.fiatCurrency
+      })
+    )
+  }, [dispatch, settings.fiatCurrency, verifiedFungibleTokenSymbols.withPriceHistory])
+
+  useInterval(
+    refreshTokensLatestPrice,
+    60000,
+    network.status !== 'online' || verifiedFungibleTokenSymbols.withPriceHistory.length === 0
+  )
```
nop33, line 265: This is how I re-implemented the polling mechanism that was in place using **RTK Query** before.

apps/desktop-wallet/src/App.tsx
```diff
@@ -34,15 +35,20 @@ import { WalletConnectContextProvider } from '@/contexts/walletconnect'
...
-import { syncNetworkTokensInfo, syncUnknownTokensInfo } from '@/storage/assets/assetsActions'
-import { selectIsTokensMetadataUninitialized } from '@/storage/assets/assetsSelectors'
+import { syncAddressesAlphHistoricBalances, syncAddressesData } from '@/storage/addresses/addressesActions'
+import {
+  makeSelectAddressesUnknownTokens,
+  selectAddressIds,
+  selectAllAddressVerifiedFungibleTokenSymbols
+} from '@/storage/addresses/addressesSelectors'
+import { syncUnknownTokensInfo, syncVerifiedFungibleTokens } from '@/storage/assets/assetsActions'
+import { selectAreVerifiedFungibleTokensInitialized } from '@/storage/assets/assetsSelectors'
```
nop33, line 45: FYI, you'll notice **many** renames. It took me half a day to wrap my head around the code and another full day to make this PR feature complete. Renaming things helped me clarify our code.

apps/desktop-wallet/src/pages/UnlockedWallet/OverviewPage/AmountsOverviewPanel.tsx
```diff
@@ -69,9 +70,9 @@ const AmountsOverviewPanel: FC<AmountsOverviewPanelProps> = ({ className, addres
   const selectAddressesHaveHistoricBalances = useMemo(makeSelectAddressesHaveHistoricBalances, [])
   const hasHistoricBalances = useAppSelector((s) => selectAddressesHaveHistoricBalances(s, addressHashes))
   const fiatCurrency = useAppSelector((s) => s.settings.fiatCurrency)
-  const { data: price, isLoading: isPriceLoading } = useGetPriceQuery(currencies[fiatCurrency].ticker, {
-    pollingInterval: 60000
-  })
```
nop33, line 74: Using **RTK Query** for polling was super cool... Very comparable experience to **TanStack Query** I suppose.

I discovered that with RTK Query [we can also provide a custom `queryFn` function](https://redux-toolkit.js.org/rtk-query/usage/customizing-queries#customizing-queries-with-queryfn), just like with TanStack, where we can use our `client` to make the requests. The only problem I faced is on our side: We need to catch for exceptions and return the HTTP status code, but [web3 is not returning it](https://github.com/alephium/alephium-web3/blob/master/packages/web3/src/api/utils.ts#L22-L29).

Another thing I did not look for answers but was in my mind is: How do we ensure that the `client` object that is referenced inside the `queryFn`s is always up to date when the network settings change?

apps/desktop-wallet/src/App.tsx
nop33: See also https://github.com/alephium/alephium-frontend/pull/322/files#r1470716937

apps/desktop-wallet/src/App.tsx
```diff
@@ -189,26 +193,80 @@ const App = () => {
...
+      }
+    }
+  }, [
+    dispatch,
+    isLoadingVerifiedFungibleTokens,
+    network.status,
+    settings.fiatCurrency,
+    verifiedFungibleTokenSymbols.uninitialized,
+    verifiedFungibleTokensNeedInitialization
+  ])
```
nop33, line 240: @mvaivre I think it makes sense to only fetch prices for **verified** fungible tokens, not just tokens that have a **symbol**. Don't you agree?

The important use-cases for this `useEffect` is that it will run:
1. when the **network status** changes to `online` (when for example the network settings change)
1. when **tokens are discovered** in addresses data that belong to the verified token list and we have not yet fetched the historic prices for
1. when the **currency settings** are changed

See also https://github.com/alephium/alephium-frontend/pull/322/files#r1470716937

apps/desktop-wallet/src/storage/assets/fungibleTokensSlice.ts
```diff
@@ -16,23 +16,25 @@ You should have received a copy of the GNU Lesser General Public License
...
+  checkedUnknownTokenIds: FungibleToken['id'][]
 }
 
-const initialState: AssetsInfoState = assetsInfoAdapter.addOne(
-  assetsInfoAdapter.getInitialState({
-    loading: false,
+const initialState: FungibleTokensState = fungibleTokensAdapter.addOne(
+  fungibleTokensAdapter.getInitialState({
+    loadingVerified: false,
+    loadingUnverified: false,
```
nop33, line 37: Splitting this into 2 allows us to display data faster, which is basically replicating the experience in our latest release.

apps/desktop-wallet/src/storage/prices/pricesActions.ts
```diff
@@ -0,0 +1,102 @@
...
+import client from '@/api/client'
+import { TokenHistoricalPrice, TokenPriceEntity, TokenPriceHistoryEntity } from '@/types/price'
+import { isFulfilled } from '@/utils/misc'
+
+export const syncTokenCurrentPrices = createAsyncThunk(
+  'assets/syncTokenCurrentPrices',
+  async ({ verifiedFungibleTokenSymbols, currency }: { verifiedFungibleTokenSymbols: string[]; currency: string }) => {
+    let tokenPrices = [] as TokenPriceEntity[]
+
+    const promiseResults = await Promise.allSettled(
```
nop33, line 33: We cannot use `Promise.all` here because if one call for a token price fails, then all will fail.

apps/explorer/package.json
```diff
@@ -32,7 +32,7 @@
   "dependencies": {
     "@alephium/shared": "workspace:*",
     "@alephium/token-list": "^0.0.12",
-    "@alephium/web3": "0.28.1",
+    "@alephium/web3": "0.29.1",
```
nop33, line 35: To avoid the bugs we experienced with the mobile wallet in the past, we need to keep the dependencies of our projects always in sync.

apps/desktop-wallet/src/api/addresses.ts
```diff
@@ -30,7 +30,9 @@ import {
 
 const PAGE_LIMIT = 100
```
mvaivre, line 31: Not directly linked to this PR but this could be extracted as well (I actually thought it was already). WDYT?

mvaivre: To be discussed later:
- For maintainability sake, we could use comments to describe what each constant does. 
- Some constants would be in SHRD, those ones should IMO be imported and re-declared in each sub-package for clarity, instead of being imported directly in the code.

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressGridRow.tsx
```diff
@@ -36,8 +36,8 @@ import {
   selectAddressByHash,
   selectIsStateUninitialized
 } from '@/storage/addresses/addressesSelectors'
-import { selectIsTokensMetadataUninitialized } from '@/storage/assets/assetsSelectors'
-import { useGetPriceQuery } from '@/storage/assets/priceApiSlice'
+import { selectDoVerifiedFungibleTokensNeedInitialization } from '@/storage/assets/assetsSelectors'
```
mvaivre, line 39: 😆

apps/desktop-wallet/src/pages/UnlockedWallet/OverviewPage/AmountsOverviewPanel.tsx
mvaivre: I feel you. Because I started to created the Redux boilerplate, it indeed made sense going that way fully.
We may try to improve this once we'll implement fungible tokens historical balances.

apps/desktop-wallet/src/storage/addresses/addressesActions.ts
```diff
@@ -181,14 +181,16 @@ export const syncAddressesHistoricBalances = createAsyncThunk(
 
     for (const addressHash of addresses) {
       const balances = []
-      const data = await client.explorer.addresses.getAddressesAddressAmountHistory(
+
+      // TODO: Do not use getAddressesAddressAmountHistoryDeprecated when the new delta endpoints are released
```
mvaivre, line 185: 👍

apps/desktop-wallet/src/storage/prices/pricesActions.ts
```diff
@@ -0,0 +1,102 @@
...
+      )
+    )
+
+    tokenPrices = promiseResults.filter(isFulfilled).flatMap((r) => r.value)
+
+    return tokenPrices
+  }
+)
+
+export const syncTokenPriceHistories = createAsyncThunk(
```
mvaivre, line 57: Nice one!

apps/desktop-wallet/src/storage/prices/pricesActions.ts
```diff
@@ -0,0 +1,102 @@
...
+          .then((rawHistory) => {
+            const today = dayjs().format(CHART_DATE_FORMAT)
+            let history = [] as TokenHistoricalPrice[]
+
+            if (rawHistory.timestamps && rawHistory.prices) {
+              const pricesHistoryArray = rawHistory.prices
+
+              history = rawHistory.timestamps.reduce((acc, v, index) => {
+                const itemDate = dayjs(v).format(CHART_DATE_FORMAT)
+                const isDuplicatedItem = !!acc.find(({ date }) => dayjs(date).format(CHART_DATE_FORMAT) === itemDate)
```
mvaivre, line 77: Are we sure this still happens (duplicated dates)? Asking because this is quite heavy, using `find` for every item in the timestamps list.

apps/explorer/package.json
mvaivre: What about writing a simple script ran with turbo, which could check this for us? 🤔

mvaivre (commented):
Awesome job! 👏

apps/desktop-wallet/src/App.tsx
nop33: Good idea 👍

apps/explorer/package.json
nop33: Yeah, I had the following npm script set up in the monorepo, but it seems to not work with the internal packages:

```
"postinstall": "manypkg check"
```

source: https://turbo.build/repo/docs/handbook/troubleshooting

nop33: The original problem is this:
- https://github.com/Thinkmill/manypkg/issues/72

Other references:
- https://github.com/Thinkmill/manypkg/issues/147
- https://github.com/Thinkmill/manypkg/issues/182

The workaround I found (and I will push in another PR) is this: https://github.com/Thinkmill/manypkg/issues/72#issuecomment-1108801027

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressGridRow.tsx
nop33: 🤷 haha

apps/desktop-wallet/src/storage/prices/pricesActions.ts
nop33: We'd need to ask Thomas or inspect it ourselves.

apps/desktop-wallet/src/api/addresses.ts
nop33: Hm, why should we re-declare them? We can easily click on the constant in our IDEs and navigate to the source where we could have the comments that you mentioned.

apps/explorer/package.json
nop33: Pushed a fix here: https://github.com/alephium/alephium-frontend/pull/344/commits/5af039482ef6543a83fbb4317c2c918e72939947

apps/desktop-wallet/src/storage/prices/pricesActions.ts
nop33: Yep, it's still needed

image

apps/desktop-wallet/src/App.tsx
mvaivre: Yeah I like it!

mvaivre:
Testing, everything looks pretty good. As expected, I'm not really happy with the amount / price history situation (lack of), but it's a temporary thing :)

mvaivre (approved):

PR #339, opened 2024-02-01
Include tokens in "Value today" calculation

nop33:
Related to #243


nop33:
Closing because code was already merged (apparently in a previous PR: https://github.com/alephium/alephium-frontend/pull/322/commits)

Issue #288, opened 2024-01-20
Received tx shows up as "moved"


nop33:
Steps to reproduce:
1. Create a new wallet
2. Send funds to devnet main address
3. Swap funds from devnet main address back to wallet
4. Look at the last transaction of the wallet

PR #341, opened 2024-02-01
Fix tx display

nop33:
Closes https://github.com/alephium/alephium-frontend/issues/288

### DIFF
--- a/packages/shared/lib/transactions.ts
+++ b/packages/shared/lib/transactions.ts
@@ -18,7 +18,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import { TokenInfo } from '@alephium/token-list'
 import { explorer, Optional } from '@alephium/web3'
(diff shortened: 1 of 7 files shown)


packages/shared/lib/transactions.ts
```diff
@@ -132,6 +131,9 @@ export const isMempoolTx = (
   transaction: explorer.Transaction | explorer.MempoolTransaction
 ): transaction is explorer.MempoolTransaction => !('blockHash' in transaction)
 
+export const isInternalTx = (tx: explorer.Transaction, internalAddresses: AddressHash[]): boolean =>
+  [...(tx.outputs ?? []), ...(tx.inputs ?? [])].every((io) => io?.address && internalAddresses.indexOf(io.address) >= 0)
```
mvaivre, line 135: Nit: I know it was already done like this before, but what about using `includes` instead of `indexOf` for readability sake?

mvaivre (approved):
LGTM!

PR #344, opened 2024-02-02
Create React shared package

nop33:
Don't be intimidated by the number of lines in this PR, it's mainly moving code around, barely any new code.

### 1. Created a scaffolding of a React shared package

We can now move many of common React context logic (see WC) between desktop and mobile wallet into one place. We can also share React components between the desktop wallet and explorer. Win 💪 

### 2. Split the shared package into `@alephium/shared` and `@alephium/shared-crypto`

I moved all code of `@alephium/shared` that required polyfilling with webpack into its own package called `@alephium/shared-crypto`. This allows for a much simpler build flow for the remaining code of  `@alephium/shared` which has now become a simple TypeScript library. Super win 💪💪 

### 3. Simple shared code development

By running `pnpm watch` from the monorepo root before starting the dev server of an app, we can watch for changes in the shared libraries without having to manually compile. Mega win 💪💪💪 

### 4. Updated docs and development workflow

The developer should not need to have `turbo` installed globally in their system. All our development workflows should be able to run with just `pnpm`. Thus, I updated the scripts to simplify them and updated our docs as well. Now, for example, to start working on the desktop wallet, a developer needs to run:

```shell
pnpm i
pnpm watch # optional, only useful if working on the shared packages as well
cd apps/desktop-wallet
pnpm start
```

I also added a simple command for doing all typechecking, lining and formatting:

```
pnpm check
```

💪💪💪💪

---

Other smol wins:
- Utilized `turbo` for cleaning up the dist/build and node_modules folders
- Simplified our `turbo` config by getting rid of `compile` task


apps/explorer/.env.example
```diff
@@ -1,3 +1,3 @@
 VITE_BACKEND_URL=
 VITE_NODE_URL=
-VITE_NETWORK_TYPE=
+VITE_NETWORK_TYPE=testnet
```
nop33, line 3: It's hard for a developer to know what to input there. Prefilling it with the string `testnet` helps, I believe.

apps/mobile-wallet/turbo.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": ["//"],
+  "pipeline": {
+    "build:deps": {
+      "dependsOn": ["^build"]
+    }
+  }
+}
```
nop33, line 8: This is needed so that EAS builds the internal packages when building the apps

mvaivre (approved):
Everything I tested worked smoothly for me! Great job!

PR #347, opened 2024-02-05
Simplify GitHub api throttle hook

nop33:
@mvaivre I think this code is more readable, what do you think?


mvaivre (approved):
That's indeed much better! Thanks :)

PR #356, opened 2024-02-13
Clean up XPLR error messages (remove details)

mvaivre:


mvaivre:
Test is failing because of an audit problem, see #357.
We may need to ignore this one (cf. discussion linked in aforementioned issue).

apps/explorer/src/components/InlineErrorMessage.tsx
```diff
@@ -26,7 +26,9 @@ const InlineErrorMessage = ({ message, code }: { message?: string; code?: number
   if (!message) {
     shownMessage = "Something's wrong"
   } else {
-    shownMessage = message
+    // Clean the error message by removing details that could be used to trick users
+    const regex = /\(([^()]|\([^()]*\))*\)/g
```
nop33, line 30: Can you please give me an explanation of what this regex does?

nop33: Also, shouldn't we do the clean up in `getHumanReadableError` so that none of our apps get affected?

nop33: Also, we should use `getHumanReadableError` in the `TransactionInfoPage`, like we do in the `BlockInfoPage` before displaying the message

nop33 (commented):
Left some comments

apps/explorer/src/components/InlineErrorMessage.tsx
mvaivre: Let's have a quick call to go through this!

nop33 (approved):

PR #357, opened 2024-02-13
Update deps (audit)

mvaivre:
~Problem: no fix yet for IP package: https://github.com/react-native-community/cli/issues/2294~

Added overrides to fix problematic packages.


apps/desktop-wallet/package.json
```diff
@@ -91,7 +91,7 @@
     "concurrently": "^7.4.0",
     "cross-env": "^7.0.3",
     "dayjs": "^1.10.7",
-    "electron": "22.3.5",
```
nop33, line 94: I would be careful upgrading electron... There was a reason I fixed the version, but I don't remember why... Let's verify that everything in the desktop wallet works, building, running dev, etc.

nop33: Maybe that's why the CI fails

nop33:
Nice, looks like the PR that fixes this was approved 1h ago :D https://github.com/react-native-community/cli/pull/2299#pullrequestreview-1887496489

nop33 (approved):

nop33 (approved):

PR #359, opened 2024-02-15
Restart electron app on wallet lock

nop33:


apps/desktop-wallet/public/electron.js
```diff
@@ -312,6 +315,20 @@ app.on('ready', async function () {
...
+  ipcMain.handle('app:restart', () => {
+    app.relaunch()
+    app.exit()
+  })
+
+  ipcMain.handle('app:clearCache', async () => {
+    const session = mainWindow.webContents.session
+
+    await session.clearCache()
+    await session.clearStorageData({
```
mvaivre, line 327: Should we keep this if this doesn't solve our issue? I'm a bit scared of potential unforeseen side effects. WDYT?

mvaivre:
Tested, works well (manual lock, auto lock...).

apps/desktop-wallet/public/electron.js
nop33: ok, we can remove it.

nop33: pushed

mvaivre (approved):
Looks good!

PR #370, opened 2024-02-20
Improve accessibility

### LINKED ISSUE #92: Increase tappable size of buttons
Google recommends:

> Consider making this clickable item larger.
> 
> This item's size is 40dp x 40dp. Consider making this touch target 48dp wide and 48dp high or larger. A parent container may be handling touch events for this item. If selecting the larger container performs the same action as selecting this item, consider defining this item as not clickable. If a different action is performed, consider increasing the size of this item. [Learn more](https://support.google.com/accessibility/android/answer/7101858)
>
> image

nop33:
Closes #92

This PR increases the tappable area around buttons without affecting their position. It has been something that was bothering me when my taps are not registered because I didn't tap right on the button. Google also warned us that we should increase the tappable areas by at least 8px in the smallest of buttons.

image

### DIFF
--- a/.changeset/witty-flies-tan.md
+++ b/.changeset/witty-flies-tan.md
@@ -1,0 +1,5 @@
+---
+"@alephium/mobile-wallet": patch
+---
+
+Improve accessibility with larger tappable areas around buttons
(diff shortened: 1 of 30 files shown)


mvaivre (approved):
LGTM! Tested pretty much all screens I think & made a tiny change in the landing screen.

PR #371, opened 2024-02-20
Fix mobile wallet assets loading

nop33:


packages/shared/src/store/assets/fungibleTokensSlice.ts
```diff
@@ -46,6 +47,12 @@ const fungibleTokensSlice = createSlice({
       .addCase(syncVerifiedFungibleTokens.pending, (state) => {
         state.loadingVerified = true
       })
+      .addCase(syncFungibleTokensInfo.pending, (state) => {
+        state.loadingUnverified = true
+      })
+      .addCase(syncUnknownTokensInfo.pending, (state) => {
```
mvaivre, line 53: Nit: Should we rename `syncUnknownTokensInfo` into something like `syncTokenTypes` to better match the `loadingTokenTypes` bool?

mvaivre (approved):
One tiny nit suggestion, but LGTM!

apps/mobile-wallet/src/components/AddressesTokensList.tsx
```diff
@@ -152,6 +154,12 @@ const AddressesTokensList = ({ addressHash, isRefreshing, style }: AddressesToke
   )
 }
 
+const AssetNumberLoader = () => (
```
mvaivre, line 157: FYI: Added a spinner in the asset tabs when loading.

packages/shared/src/store/assets/fungibleTokensSlice.ts
nop33: `syncUnknownTokensInfo` is doing 3 things:
1. Gets the types
2. Calls `syncFungibleTokensInfo`
3. Calls `syncNFTsInfo`

Issue #373, opened 2024-02-21
Make sure wallet always knows about the latest address balances before calling the sweep endpoint


nop33:
If the wallet does not have the latest address data before sending a tx (this is the case where the same seed is imported in multiple wallets and the second wallet sends a tx that the first wallet is not aware of), then sending a tx might result in calling the sweep endpoint. 

The condition to call sweep is:
1. Did the user select all tokens of this address?
2. Did the user select an amount of each token that is equal to thethe available balance of each token

If the answer is true, then it calls sweep

Suggested solution: 

I suggest this:
1. if all of the tokens are selected as MAX, we do a sweep query
2. when the unsigned transaction is received, we check if there are new tokens there (by decoding the unsigned tx using the upcoming web3 TS function). If not, we sign and submit the tx. 

Otherwise, we handle the corner case

PR #374, opened 2024-02-21
Sync origin address data when sending

nop33:
Related to #373


nop33:
With the auto-polling that https://github.com/alephium/alephium-frontend/pull/414 implements, this PR is now obsolete.

PR #383, opened 2024-02-22
Use locale number separators in wallets & explorer

### LINKED ISSUE #128: Separate groups of thousands in amount display based on region

nop33:
Closes #128

### DIFF
--- a/.changeset/lovely-jeans-reply.md
+++ b/.changeset/lovely-jeans-reply.md
@@ -1,0 +1,5 @@
+---
+"alephium-desktop-wallet": patch
+---
+
+Display numbers using locale separators
--- a/.changeset/neat-camels-invite.md
+++ b/.changeset/neat-camels-invite.md
@@ -1,0 +1,5 @@
+---
+"@alephium/mobile-wallet": patch
+---
+
+Display numbers using locale separators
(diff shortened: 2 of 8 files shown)


nop33:
Another complaint regarding this, this time on Reddit: https://www.reddit.com/r/Alephium/comments/1bezpf5/alephium_explorer_formatting_improvement/

cc @mvaivre

mvaivre:
I've been testing it and IMO the new format can be more confusing than before:

**NEW**
image

instead of 
OLD
image

nop33:
> Today, simply switch from using apostrophes to comas for thousand separators.

Is this more common, globally? 

Let's keep these docs in mind when we implement the proper solution: 
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

mvaivre:
Here's what ChatGPT has to say. Switching to coma makes sense IMO.

- In many English-speaking countries and most of Europe, the comma is used as the thousand separator, and the period is used as the decimal separator (e.g., 1,000.00).
- In some European countries and other parts of the world, the roles are reversed: the period is used as the thousand separator, and the comma is used as the decimal separator (e.g., 1.000,00).
- The apostrophe as a thousand separator is less common but is used in some countries, such as Switzerland for Italian-speaking regions (e.g., 1'000.00).

nop33:
Closing PR since we'll never merge this.

Issue #244, opened 2024-01-08
Display fiat prices of tokens in mobile wallet


nop33:

ElvisGAN777:
This is much needed at this point. I am onboarding normies and they are confounded as to why the value of their alternate tokens are not showing up in the total fiat balance.

diomark:
agreed definately good to have this. guessing the prices need to come from alph backend somehow though instead of being fetched from coingecko everytime

hopsai:
It would be nice to have the valuation of the tokens in the wallet as well as the valuation of the ALPH.

nop33:
Thanks for the feedback people. It is our highest priority task. It has been a tricky feature that required work from many teams. It will be released asap.

ElvisGAN777:
Thank you @nop33

PR #405, opened 2024-03-04
Display fiat worth of each token in mobile wallet and sort assets of both wallets based on worth

### LINKED ISSUE #403: Order tokens based on their fiat worth in mobile wallet

### LINKED ISSUE #404: Order tokens based on their fiat worth in desktop wallet

nop33:
Related to #244
Closes #403
Closes #404


mvaivre:
Tested on mobile and desktop, works well! I'm wondering if we should maybe order tokens by fiat worth when it's known.
It seems to be the standard in many wallet. WDYT @VDAODAO?

nop33:
> I'm wondering if we should maybe order tokens by fiat worth when it's known.

I am not sure what you mean. This is what was implemented, with exception of ALPH that is always shown first.

VDAODAO:
I've checked in a few different wallets I have and it always works like this: ranked by fiat value (trust, metamask, keplr, cobo, live, phantom). Those who don't have a fiat value can then be ranked below by either name or numerical value of units.

mvaivre:
> I am not sure what you mean. This is what was implemented, with exception of ALPH that is always shown first.

VDAODAO:
@mvaivre i agree with that 100%

nop33:
@mvaivre @VDAODAO done, pushed

mvaivre (approved):

PR #418, opened 2024-03-07
Fix GitHub API rate issue

nop33:
Here is a quick solution of 2 problems:
1. Infinite loop when GitHub API request fails: If the retry strategy fails, we stop querying until the user re-launches the app
2. Announcement & last version check conflict: We were reusing the same timestamp to throttle these GitHub API requests. The quickest way to solve this that I thought of is to store a serate timestamp for each.

### DIFF
--- a/apps/desktop-wallet/src/hooks/useThrottledGitHubApi.ts
+++ b/apps/desktop-wallet/src/hooks/useThrottledGitHubApi.ts
@@ -18,22 +18,27 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import { useState } from 'react'
 
-import { AppMetaData, isRcVersion, KEY_APPMETADATA, toAppMetaData } from '@/utils/app-data'
+import { AppMetaData, AppMetadataGitHub, isRcVersion, KEY_APPMETADATA, toAppMetaData } from '@/utils/app-data'
 import { useTimeout } from '@/utils/hooks'
 
 // TODO: Move to shared
 const ONE_HOUR_IN_MS = 1000 * 60 * 60
 
-const useThrottledGitHubApi = (githubApiCallback: (appMetadata: AppMetaData) => Promise<void>) => {
+interface ThrottledGitHubApiProps {
+  key: keyof AppMetadataGitHub
+  githubApiCallback: (appMetadata: AppMetaData) => Promise<void>
+}
+
+const useThrottledGitHubApi = ({ key, githubApiCallback }: ThrottledGitHubApiProps) => {
   const [timeoutDelay, setTimeoutDelay] = useState(0)
 
   const timeoutCallback = () => {
     const now = new Date()
-    const lastTimeGitHubApiWasCalled = getLastTimeGitHubApiWasCalled()
+    const lastTimeGitHubApiWasCalled = getLastTimeGitHubApiWasCalled(key)
     const timePassedSinceGitHubApiWasCalledInMs = now.getTime() - lastTimeGitHubApiWasCalled.getTime()
 
     if (timePassedSinceGitHubApiWasCalledInMs > ONE_HOUR_IN_MS) {
-      const updatedAppMetadata = storeAppMetadata({ lastTimeGitHubApiWasCalled: now })
+      const updatedAppMetadata = storeAppMetadata({ [key]: now })
 
       setTimeoutDelay(ONE_HOUR_IN_MS)
       githubApiCallback(updatedAppMetadata)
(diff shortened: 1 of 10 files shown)


apps/desktop-wallet/src/utils/app-data.ts
```diff
@@ -21,15 +21,20 @@ export const KEY_APPMETADATA = 'alephium/desktop-wallet/appmetadata'
 export const currentVersion: string = import.meta.env.VITE_VERSION
 export const isRcVersion: boolean = currentVersion.includes('-rc.')
 
-export interface AppMetaData {
-  lastTimeGitHubApiWasCalled: Date
+export interface AppMetadataGitHub {
+  lastTimeGitHubApiWasCalledForLatestVersion: Date
```
mvaivre, line 25: `lastGithubCallForLatestVersion: Date` ?

apps/desktop-wallet/src/utils/app-data.ts
```diff
@@ -21,15 +21,20 @@ export const KEY_APPMETADATA = 'alephium/desktop-wallet/appmetadata'
...
+  lastTimeGitHubApiWasCalledForAnnoumcenent: Date
```
mvaivre, line 26: Little typo here.

`lastGithubCallForAnnouncement: Date` ?

apps/desktop-wallet/src/utils/app-data.ts
```diff
@@ -21,15 +21,20 @@ export const KEY_APPMETADATA = 'alephium/desktop-wallet/appmetadata'
...
+}
+
+export interface AppMetaData extends AppMetadataGitHub {
```
mvaivre, line 29: It's kinda weird that the more generically named `AppMetadata` extends `AppMetadataGitHub` IMO. Should be do it the other way around?

```ts
export type AppMetadataGitHub = {
  lastGithubCallForLatestVersion: Date
  lastGithubCallForAnnouncement: Date
}

export type AppMetaData = {
  lastAnnouncementHashChecked: string
} & AppMetadataGitHub
```

packages/shared/src/store/assets/fungibleTokensSlice.ts
```diff
@@ -47,6 +47,10 @@ const fungibleTokensSlice = createSlice({
       .addCase(syncVerifiedFungibleTokens.pending, (state) => {
         state.loadingVerified = true
       })
+      .addCase(syncVerifiedFungibleTokens.rejected, (state) => {
+        state.loadingVerified = false
+        state.status = 'initialization-failed'
```
mvaivre, line 52: The `initialization-failed` isn't used anywhere for now, right? I guess you added it so that we can show something in the UI at one point?

apps/desktop-wallet/src/hooks/useThrottledGitHubApi.ts
```diff
@@ -18,22 +18,27 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
 // TODO: Move to shared
 const ONE_HOUR_IN_MS = 1000 * 60 * 60
 
-const useThrottledGitHubApi = (githubApiCallback: (appMetadata: AppMetaData) => Promise<void>) => {
+interface ThrottledGitHubApiProps {
+  key: keyof AppMetadataGitHub
+  githubApiCallback: (appMetadata: AppMetaData) => Promise<void>
+}
+
+const useThrottledGitHubApi = ({ key, githubApiCallback }: ThrottledGitHubApiProps) => {
```
mvaivre, line 32: Nit: what about renaming `key` to a more specific: `lastGithubCallTimestampKey`

apps/desktop-wallet/src/utils/app-data.ts
nop33: Don't you think we should be as clear as possible to avoid the readability problems we faced recently?

apps/desktop-wallet/src/utils/app-data.ts
nop33: That's the same thing, no?

packages/shared/src/store/assets/fungibleTokensSlice.ts
nop33: https://github.com/alephium/alephium-frontend/blob/master/packages/shared/src/store/assets/assetsSelectors.ts#L28-L37

Before `initialization-failed` was introduced, the `status` was remaining `uninitialized`, which was causing the above selector to return `true`. By setting the `status` to `initialization-failed`, the selector returns now `false`.

apps/desktop-wallet/src/utils/app-data.ts
nop33: Typo fixed

apps/desktop-wallet/src/hooks/useThrottledGitHubApi.ts
nop33: Fixed

apps/desktop-wallet/src/utils/app-data.ts
mvaivre: Mhhh in my logic, you **extend** from a more generic, more global type.

In my version, it's an intersection, you don't have any hierarchy involved. Maybe it's just a matter of personal pref?

packages/shared/src/store/assets/fungibleTokensSlice.ts
mvaivre: Ah!! Alright, makes sense.

apps/desktop-wallet/src/utils/app-data.ts
nop33: fixed.

mvaivre (approved):

PR #423, opened 2024-03-08
Fix all fetches

nop33:
I did a global search for all `fetch` calls and I replaced them with the retry fetch and handle their potential failures.


nop33:
We don't need a changeset for this one, doesn't need to be in the changelog

mvaivre (approved):
LGTM

Issue #352, opened 2024-02-09
Android get stuck on startup and restarting doesn't show biometrics login (samsung, xiaomi)


nop33:
Multiple reports on this, all from Samsung or Xiaomi users.

<img src="https://github.com/alephium/alephium-frontend/assets/1579899/af5a1391-7177-4a89-9251-ed8d59fa4ce0" width="300" />

mvaivre:
More info: apparently, a device restart solved the issue 🤔

nop33:
Not always. We got an email:

> ** Cannot open wallet app on Samsung S20**
>
> Hey guys, i just got a new to me S20fe and i installed the app from the Play Store but it won't open. I get the welcome screen but it doesn't get past there and ask for a seed phrase or make a new account. My old phone (s10) is useable and i can access the wallet from there. I have uninstalled and reinstalled tge app. Any ideas?

nop33:
This indicates that there is a problem with accessing the secure storage on some devices...

nop33:
We have the same issue with Xiaomi phones

nop33:
This might be related: https://stackoverflow.com/a/67195378/1644591

Let's disable this flag https://docs.expo.dev/versions/latest/config/app/#allowbackup

nop33:
@mvaivre did you have a chance in reproducing this on the Samsung phone?

mvaivre:
No, I didn't get to reproduce it. Will try again today with the latest version to make sure.

mvaivre:
Can't reproduce on my Samsung phone.

mvaivre:
Didn't get any new reports. Closing, but feel free to reopen if this happens on your device.

PR #429, opened 2024-03-08
Set allowBackup to false & fix crash

nop33:
Metamask does the same: https://github.com/MetaMask/metamask-mobile/blob/910313340e43c24473d057825a5a810d32a55a6e/android/app/src/main/AndroidManifest.xml#L28

Same as Rainbow: https://github.com/rainbow-me/rainbow/blob/74d054fa48e7649c7c18376e4da9e78df80d295d/android/app/src/main/AndroidManifest.xml#L25

Possibly related to #352

- [x] Confirm that the change of `WHEN_UNLOCKED_THIS_DEVICE_ONLY` does not break existing wallets

### DIFF
--- a/.changeset/orange-beans-lick.md
+++ b/.changeset/orange-beans-lick.md
@@ -1,0 +1,5 @@
+---
+"@alephium/mobile-wallet": patch
+---
+
+Disable app backup on Android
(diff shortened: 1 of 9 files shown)


mvaivre (approved):
👏

PR #446, opened 2024-03-12
Handle WC request to sign unsigned tx in mobile wallet

nop33:


nop33:
> because the toolkit wasn't working well for me

@mvaivre do you mean alephium.github.io/wallet-test-app?

It only works on devnet atm.

Did you follow Cheng's instructions on Slack?

mvaivre:
Great job 👏 

A few things I discovered after testing this, one small one related to this PR + a few things related to the message signing (may be fixed it in a subsequent PR).

image

Plus one general NIT: in all tables I saw, we can define the last row as being the last, so that the bottom border isn't set.

nop33:
Thanks for your review @mvaivre !!!

Regarding the spacings, I'll work on it, but would you mind if I don't hold back the beta/prod release and fix them in a follow up PR? :/

As for the rejection message: If you close the modal without selecting an option we need to inform WalletConnect somehow of what the user did, otherwise the WC request will remain pending, blocking all other requests. That's why we reject it when the user closes the modal, same as if they press "Reject".

As for the error, I don't know why it happens but it doesn't seem to affect anything. Been seeing it every now and again.

mvaivre:
Thanks for your answer @nop33! Sounds good to me.

apps/mobile-wallet/src/components/Row.tsx
```diff
@@ -39,14 +39,26 @@ export interface RowProps {
   isLast?: boolean
   style?: StyleProp<ViewStyle>
   layout?: AnimatedProps<ViewProps>['layout']
+  isHorizontal?: boolean
```
mvaivre, line 42: Wouldn't it be more logical to call this `isVertical` when we're stacking the first cell on top of the second cell?

apps/mobile-wallet/src/contexts/walletConnect/WalletConnectSessionRequestModal.tsx
```diff
@@ -287,12 +310,22 @@ const WalletConnectSessionRequestModal = <T extends SessionRequestData>({
...
+            <Row isHorizontal title="Bytecode" titleColor="secondary">
               <AppText>{requestData.wcData.bytecode}</AppText>
             </Row>
           )}
+          {requestData.type === 'sign-unsigned-tx' && (
+            <>
+              <Row isHorizontal title="Unsigned TX ID" titleColor="secondary">
+                <AppText>{requestData.unsignedTxData.unsignedTx.txId}</AppText>
+              </Row>
+              <Row isHorizontal title="Unsigned TX" titleColor="secondary">
```
mvaivre, line 322: NIT: `isLast`

apps/mobile-wallet/src/contexts/walletConnect/WalletConnectSessionRequestModal.tsx
```diff
@@ -287,12 +310,22 @@ const WalletConnectSessionRequestModal = <T extends SessionRequestData>({
...
+                <AppText>{requestData.wcData.unsignedTx}</AppText>
+              </Row>
+            </>
+          )}
           {requestData.type === 'sign-message' && (
-            <Row title="Message" titleColor="secondary">
+            <Row isHorizontal title="Message" titleColor="secondary">
```
mvaivre, line 328: NIT: `isLast`

apps/mobile-wallet/src/components/Row.tsx
nop33: That's right

nop33:
@mvaivre all your comments were addressed! Re-requested review.

mvaivre (approved):
🚀

PR #448, opened 2024-03-12
Remove error analytics

nop33:
I want to avoid the issue of floading the analytics with errors because of a `useEffect`, like it happened with the GH API rate limits. I went through both our wallets in search of the places we report things to PostHog and removed some. @mvaivre I would like to review them with you and think of a way to ensure that we won't end up with the same problem in the future.

### DIFF
--- a/apps/mobile-wallet/src/components/WalletSwitchButton.tsx
+++ b/apps/mobile-wallet/src/components/WalletSwitchButton.tsx
@@ -64,6 +64,7 @@ const WalletSwitchButton = ({ style }: WalletSwitchButtonProps) => {
       setNbOfTaps(0)
     }
 
+    // TODO: Shouldn't this be 69?
     if (nbOfTaps === 68) {
       setIsDoingMagic(true)
       sendAnalytics('Activated magic')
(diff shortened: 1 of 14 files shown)


mvaivre:
Should we plan a little call to go through this?

nop33:
Yes

nop33:
@mvaivre as we are preparing to fix the analytics issue, I would like to merge this. As the description of the PR says, I went through all places where we call the Posthog API to see where there could be a potential issue in the future. I removed a few API requests in places that I deemed necessary.

I also posted a question in Posthog support here: https://posthog.com/questions/how-to-limit-the-amount-of-requests-sent-from-my-app-to-posthog

Let me know how you'd like to proceed.

nop33:
@mvaivre when you have the time, please check my last comment here 🙏

nop33:
kind ping @mvaivre for my previous comments 🙏

apps/mobile-wallet/src/components/WalletSwitchButton.tsx
```diff
@@ -64,6 +64,7 @@ const WalletSwitchButton = ({ style }: WalletSwitchButtonProps) => {
       setNbOfTaps(0)
     }
 
+    // TODO: Shouldn't this be 69?
```
mvaivre, line 67: 😨 you're right. I wanted to start counting from -1 to make it less obvious, but apparently I considered this as non vital and made a mistake 🙃

mvaivre (approved):

PR #451, opened 2024-03-14
[1.0.8] Fix bottom modal padding

nop33:
The trick was to move `inset.bottom` from the container to the content


mvaivre (approved):
LGTM!

Issue #425, opened 2024-03-08
Fix display of NFT attributes in mobile wallet


nop33:
<img width="374" alt="image" src="https://github.com/alephium/alephium-frontend/assets/1579899/5e7bc553-de7a-48f0-a4c1-f9d4d4d4c503">

PR #455, opened 2024-03-15
[1.0.8] Fix image loading

nop33:


mvaivre:
The changeset wan't detected 🤔

mvaivre:
Thanks for the nice work :) 

Testing : NFTs take a while to load for me (and the initial loading time is long a well), here's a few things I noticed:

image

We shouldn't show the "No NFTs yet" while NFTs are loading. When loading, a tap on the "NFTs" button should open the modal and show a loading spinner, and a message "Your NFTs are loading" IMO

image

Tab shouldn't be accessible, otherwise we get to situation like the one shown in the screenshot above.

nop33:
> We shouldn't show the "No NFTs yet" while NFTs are loading. When loading, a tap on the "NFTs" button should open the modal and show a loading spinner, and a message "Your NFTs are loading" IMO.

👍 

> Also, to avoid displaying a wrong number of tokens (cf. screenshot above), we should maybe use a spinner in the token tab as well. WDYT?

👍 

> Tab shouldn't be accessible, otherwise we get to situation like the one shown in the screenshot above.

👍 

> Last point, IPFS loading is really hit or miss. Don't know exactly what can be done to mitigate this, but it gets to the point where I have literally 0 images showing, for like minutes:

That is indeed a pain.

> Should we display a little spinner in the NFTs placeholders to show that something is happening?

We can provide a `placeholder`, see here: https://docs.expo.dev/versions/latest/sdk/image/#usage. I would avoid spinners (imagine a whole screen of spinners 😵). How could we build such a `blurhash` placeholder?

> Last thing I noticed, not linked to this PR, the boolean attributes aren't shown properly:

I have an open issue for that: https://github.com/alephium/alephium-frontend/issues/425


---

Question: What do we do when there are no NFTs, after NFTs have finished loading?

mvaivre (approved):
🚀 
Let's ask the team to focus their attention on the loading experience.

PR #467, opened 2024-03-20
[1.0.8] Public keys screen

### LINKED ISSUE #435: Expose public key in mobile wallet
The public key is available for desktop wallet and extension wallet. Would be good to have it in mobile wallet, as requested by Linx team.

mvaivre:
Fixes #435


mvaivre:
@nop33 Chengeset bot is acting weird. It didn't find a changeset when we added one (in other PR), and it finds one here when there's none.

apps/mobile-wallet/src/screens/PublicKeysScreen.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { selectAllAddresses } from '~/store/addressesSlice'
+import { showToast, ToastDuration } from '~/utils/layout'
+
+interface ScreenProps extends StackScreenProps<RootStackParamList, 'PublicKeysScreen'>, ScrollScreenProps {}
+
+const PublicKeyScreen = ({ navigation, ...props }: ScreenProps) => {
+  const addresses = useAppSelector(selectAllAddresses)
+
+  const handleAddressPress = async (publicKey: string) => {
+    await Clipboard.setStringAsync(publicKey)
```
nop33, line 40: try/catch needed to avoid crashing the app if an exception is thrown

apps/mobile-wallet/src/screens/PublicKeysScreen.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <ScreenSection>
+        <AppText>Click on an address to copy its public key to the clipboard.</AppText>
+      </ScreenSection>
+      <ScreenSection>
+        <BoxSurface>
+          {addresses.map((address) => (
+            <Row key={address.hash} onPress={() => handleAddressPress(address.publicKey)}>
+              <AddressBadge addressHash={address.hash} />
+            </Row>
+          ))}
```
nop33, line 57: It should be a flatlist. Check out how `FlatListScreen` is used in our app. That takes care of performance and of not displaying the bottom border for the last item,

apps/mobile-wallet/src/screens/PublicKeysScreen.tsx
```diff
@@ -0,0 +1,64 @@
...
+    await Clipboard.setStringAsync(publicKey)
+
+    sendAnalytics('Copied public key')
+    showToast({ text1: 'Public key copied!', visibilityTime: ToastDuration.SHORT })
+  }
+
+  return (
+    <ScrollScreen verticalGap fill screenTitle="Public keys" headerOptions={{ type: 'stack' }} {...props}>
+      <ScreenSection>
+        <AppText>Click on an address to copy its public key to the clipboard.</AppText>
```
nop33, line 49: nit: replace "Click" with "Tap"

nop33 (commented):
:clap:

mvaivre:
Note to myself: current implementation of Row's onTouchEnd clashes with screen scrolling.

apps/mobile-wallet/src/components/Row.tsx
```diff
@@ -77,7 +82,12 @@ const Row = ({
   )
 
   return onPress ? (
-    <AnimatedPressable onPress={onPress} style={style}>
+    <AnimatedPressable
+      onPress={onPress}
+      onTouchStart={handleTouchStart}
+      onTouchEnd={handleTouchEnd}
+      style={[style, { opacity: isPressed ? 0.8 : 1 }]}
```
nop33, line 89: just curious, shouldn't this be a shared value to avoid re-rendering? Maybe I'm wrong

apps/mobile-wallet/src/components/layout/ScrollScreen.tsx
```diff
@@ -43,6 +44,7 @@ export interface ScrollScreenProps extends ScrollScreenBaseProps, ScrollViewProp
   scrollViewRef?: RefObject<ScrollView>
   verticalGap?: number | boolean
   contentPaddingTop?: number | boolean
+  screenIntroPaddingBottom?: boolean
```
nop33, line 47: Is there a case where we have `screenIntro` but we don't want padding bottom? If there is maybe it's worth it to refactor those cases so that the UI is consistent. It's a weird prop to have imo.

apps/mobile-wallet/src/screens/PublicKeysScreen.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { BackButton } from '~/components/buttons/Button'
+import FlatListScreen from '~/components/layout/FlatListScreen'
+import { ScrollScreenProps } from '~/components/layout/ScrollScreen'
+import Row from '~/components/Row'
+import { useAppSelector } from '~/hooks/redux'
+import RootStackParamList from '~/navigation/rootStackRoutes'
+import { selectAllAddresses } from '~/store/addressesSlice'
+import { showToast, ToastDuration } from '~/utils/layout'
+
+interface PublicKeysScreenProps extends StackScreenProps<RootStackParamList, 'PublicKeysScreen'>, ScrollScreenProps {}
```
nop33, line 33: It should extend `FlatListScreenProps` (see `AddressFlatListScreenProps` as an example)

apps/mobile-wallet/src/screens/PublicKeysScreen.tsx
```diff
@@ -0,0 +1,70 @@
...
+      showToast({ text1: 'Public key copied!', visibilityTime: ToastDuration.SHORT })
+      sendAnalytics('Copied public key')
+    } catch (error) {
+      console.log(error)
+      showToast({ text1: 'Error while copying ', visibilityTime: ToastDuration.SHORT, type: 'error' })
+    }
+  }
+
+  return (
+    <FlatListScreen
```
nop33, line 51: Should include `{...props}` in its arguments (see `AddressFlatListScreen` as example)

apps/mobile-wallet/src/screens/new-wallet/ImportWalletSeedScreen.tsx
```diff
@@ -143,9 +142,10 @@ const ImportWalletSeedScreen = ({ navigation, ...props }: ImportWalletSeedScreen
           headerRight: () => <ContinueButton onPress={handleWalletImport} disabled={!isImportButtonEnabled} />
         }}
         keyboardShouldPersistTaps="always"
+        screenTitle="Secret phrase"
+        screenIntro={`Enter the secret phrase for the ${name} wallet.`}
```
nop33, line 146: We lost the wrapping of the wallet name with `"` with this change.

apps/mobile-wallet/ios/Podfile.lock
```diff
@@ -1577,7 +1577,7 @@ EXTERNAL SOURCES:
 
 SPEC CHECKSUMS:
   boost: d3f49c53809116a5d38da093a8aa78bf551aed09
-  DoubleConversion: fea03f2699887d960129cc54bba7e52542b6f953
```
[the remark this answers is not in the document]
nop33:
```
$ pnpm --version

8.15.4

$ pnpm dlx pod-install

Scanning for pods...
1.15.2
```

This is my diff:

```diff
diff --git a/apps/mobile-wallet/ios/Alephium.xcodeproj/project.pbxproj b/apps/mobile-wallet/ios/Alephium.xcodeproj/project.pbxproj
index 3897f2d27..6a662ef37 100644
--- a/apps/mobile-wallet/ios/Alephium.xcodeproj/project.pbxproj
+++ b/apps/mobile-wallet/ios/Alephium.xcodeproj/project.pbxproj
@@ -444,11 +444,7 @@
                                ONLY_ACTIVE_ARCH = YES;
                                OTHER_CFLAGS = "$(inherited)";
                                OTHER_CPLUSPLUSFLAGS = "$(inherited)";
-                               OTHER_LDFLAGS = (
-                                       "$(inherited)",
-                                       "-Wl",
-                                       "-ld_classic",
-                               );
+                               OTHER_LDFLAGS = "$(inherited)";
                                REACT_NATIVE_PATH = "${PODS_ROOT}/../../../../node_modules/react-native";
                                SDKROOT = iphoneos;
                                USE_HERMES = true;
@@ -502,11 +498,7 @@
                                MTL_ENABLE_DEBUG_INFO = NO;
                                OTHER_CFLAGS = "$(inherited)";
                                OTHER_CPLUSPLUSFLAGS = "$(inherited)";
-                               OTHER_LDFLAGS = (
-                                       "$(inherited)",
-                                       "-Wl",
-                                       "-ld_classic",
-                               );
+                               OTHER_LDFLAGS = "$(inherited)";
                                REACT_NATIVE_PATH = "${PODS_ROOT}/../../../../node_modules/react-native";
                                SDKROOT = iphoneos;
                                USE_HERMES = true;
diff --git a/apps/mobile-wallet/ios/Podfile.lock b/apps/mobile-wallet/ios/Podfile.lock
index 827118315..f11a26889 100644
--- a/apps/mobile-wallet/ios/Podfile.lock
+++ b/apps/mobile-wallet/ios/Podfile.lock
@@ -1577,7 +1577,7 @@ EXTERNAL SOURCES:

 SPEC CHECKSUMS:
   boost: d3f49c53809116a5d38da093a8aa78bf551aed09
-  DoubleConversion: 5189b271737e1565bdce30deb4a08d647e3f5f54
+  DoubleConversion: fea03f2699887d960129cc54bba7e52542b6f953
   EXApplication: 137189a3f149b4e8e546884629392c3efc94cbd3
   EXBarCodeScanner: d59fd943cebee3f913ebf4ffde0d05d344da8b78
   EXConstants: 988aa430ca0f76b43cd46b66e7fae3287f9cc2fc
@@ -1611,7 +1611,7 @@ SPEC CHECKSUMS:
   FBLazyVector: 84f6edbe225f38aebd9deaf1540a4160b1f087d7
   FBReactNativeSpec: dd7ff1f4ea433fb1820e5ea80da25a0e65ea9de2
   fmt: ff9d55029c625d3757ed641535fd4a75fedc7ce9
-  glog: 04b94705f318337d7ead9e6d17c019bd9b1f6b1b
+  glog: c5d68082e772fa1c511173d6b30a9de2c05a69a2
   hermes-engine: b2669ce35fc4ac14f523b307aff8896799829fe2
   libaom: 144606b1da4b5915a1054383c3a4459ccdb3c661
   libavif: 84bbb62fb232c3018d6f1bab79beea87e35de7b7
```

apps/mobile-wallet/src/components/layout/ScrollScreen.tsx
mvaivre: You're right, I've tried multiple approaches but I actually can do easily without it with the current solution I believe.

apps/mobile-wallet/src/components/Row.tsx
mvaivre: I've used a shareValue in doubt

apps/mobile-wallet/src/screens/PublicKeysScreen.tsx
mvaivre: What I did initially, but then the `RootStackNavigation` complained.
I realise now that the `AddressFlatListScreen` is always used inside a Screen declared as a `ScrollScreen` (cf. `OriginScreen`, `AddressScreen`) which is why the compiler isn't screaming. But it's wrong too..

apps/mobile-wallet/src/screens/new-wallet/ImportWalletSeedScreen.tsx
mvaivre: Compiler was complaining at one point. But you're right, adding them back now (no complains anymore).

apps/mobile-wallet/src/screens/PublicKeysScreen.tsx
```diff
@@ -0,0 +1,71 @@
...
+      headerOptions={{
+        headerTitle: 'Public keys',
+        type: 'stack',
+        headerLeft: () => <BackButton onPress={() => navigation.goBack()} />
+      }}
+      screenTitle="Public keys"
+      screenIntro="Tap on an address to copy its public key to the clipboard."
+      keyExtractor={(item) => item.hash}
+      {...props}
```
nop33, line 60: Unless this is at the bottom of the props list, it won't be able to override `data` or `renderItem`.

apps/mobile-wallet/src/screens/PublicKeysScreen.tsx
nop33: If you don't want to figure out how we should properly fix this in all our screens, please create an issue on GH and reference it in a TODO 🙏

PR #469, opened 2024-03-22
Simplify app metadata

nop33:


mvaivre:
Looks like it's working well! I've added a line to hide the banner when the action button is clicked.

mvaivre (approved):

PR #474, opened 2024-03-26
Improve secrets management in desktop wallet

nop33:
1. Removes `mnemonic` string of words, `private keys` hex strings, and `passphrase` string from Redux state.
2. Inspired by [Metamask](https://github.com/MetaMask/eth-hd-keyring), it introduces the `Keyring` singleton in a new `@alephium/keyring` package to cache:
   1. the `root` HD key (used for address derivation) and
   2. the `private keys` of each derived address.
4. Persists the mnemonic as an encrypted array of indexes (of the mnemonic words in the bip39 word list), instead of persisting it as an encrypted string of words.
5. To match Metamask's implementation, it replaces following `bip` packages:
   1. [bip39](https://www.npmjs.com/package/bip39) is replaced with [@metamask/scure-bip39](https://www.npmjs.com/package/@metamask/scure-bip39) which is a fork of and [@scure/bip39](https://www.npmjs.com/package/@scure/bip39)
   2. [bip32](https://www.npmjs.com/package/bip32) is replaced with [ethereum-cryptography](https://www.npmjs.com/package/ethereum-cryptography) which is a re-exported [@scure/bip32](https://www.npmjs.com/package/@scure/bip32)
 6. It [patches](https://github.com/alephium/alephium-frontend/pull/474/files#diff-6359c5557bc07f7b0455de8ce9664cb20df9fea72e6c240b10b577fed3f7ef6a) `@metamask/scure-bip39` to reset the intermittent mnemonic `Uint8Array` used for seed derivation, so that ProcessHacker2 cannot find the mnemonic string after wallet unlock.
 7. [Replaces](https://github.com/alephium/alephium-frontend/pull/474/files#diff-5a08b60d82bc8dff7cd8abea7815aaf3bdecc84565c7b314d757d44fc9bda6eaR251-R258) the use of `Buffer` with `Uint8Array`


apps/desktop-wallet/src/contexts/global.tsx
```diff
@@ -82,58 +82,85 @@ export const GlobalContextProvider: FC<{ overrideContextValue?: PartialDeep<Glob
...
+    // In version 1 the encrypted mnemonic used to be stored as a string before we started using Buffer. This migrates
+    // the encrypted wallet from StoredStateV1 to StoredStateV2.
+    if (version === 1) {
+      try {
+        WalletStorage.update(walletId, { encrypted: keyring.encryptMnemonicForStorageAndCachePassword(password) })
+        console.log('Migrated stored mnemonic from version 1 to version 2')
+      } catch (e) {
+        console.error('Failed migrating stored mnemonic from version 1 to version 2', e)
+      }
+    }
```
nop33, line 113: I purposefully didn't want to create a migration script for this one because I want to avoid passing the password around.

apps/desktop-wallet/src/api/transactions.ts
```diff
@@ -17,7 +17,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
 import { AddressHash, client } from '@alephium/shared'
-import { transactionSign } from '@alephium/web3'
+import { keyring } from '@alephium/shared-crypto'
```
nop33, line 20: I named the class that holds all secrets `Keyring`, in accordance to Metamask's package: https://github.com/MetaMask/eth-hd-keyring/blob/main/index.js#L28

apps/desktop-wallet/src/api/transactions.ts
```diff
@@ -35,7 +35,7 @@ export const buildSweepTransactions = async (fromAddress: Address, toAddressHash
 }
 
 export const signAndSendTransaction = async (fromAddress: Address, txId: string, unsignedTx: string) => {
-  const signature = transactionSign(txId, fromAddress.privateKey)
+  const signature = keyring.signTransaction(txId, fromAddress.hash)
```
nop33, line 38: Since we are not storing private keys on our side anymore, the signing of a transaction is delegated to the keyring, like on Metamask: https://github.com/MetaMask/eth-hd-keyring/blob/main/index.js#L170

apps/desktop-wallet/src/components/AddressColorIndicator.tsx
```diff
@@ -38,7 +38,7 @@ const AddressColorIndicator = ({
   className
 }: AddressColorIndicatorProps) => {
   const address = useAppSelector((s) => selectAddressByHash(s, addressHash))
-  const isPassphraseUsed = useAppSelector((s) => s.activeWallet.passphrase)
+  const isPassphraseUsed = useAppSelector((s) => s.activeWallet.isPassphraseUsed)
```
nop33, line 41: Since we are not storing the passphrase anymore, I replaced it with a boolean flag.

apps/desktop-wallet/src/components/AppHeader.tsx
```diff
@@ -52,16 +52,15 @@ const AppHeader: FC<AppHeader> = ({ children, title, className, invisible }) =>
   const theme = useTheme()
   const dispatch = useAppDispatch()
   const defaultAddress = useAppSelector(selectDefaultAddress)
-  const mnemonic = useAppSelector((s) => s.activeWallet.mnemonic)
-  const isPassphraseUsed = useAppSelector((s) => s.activeWallet.passphrase)
+  const isAuthenticated = useAppSelector((s) => !!s.activeWallet.id)
```
nop33, line 55: Since we are not storing the mnemonic anymore, we now compute if a wallet is unlocked or not by checking of there is an active wallet id in redux.

apps/desktop-wallet/src/components/Buttons/ShortcutButtons.tsx
```diff
@@ -71,6 +72,7 @@ const ShortcutButtons = ({
   const [isAddressOptionsModalOpen, setIsAddressOptionsModalOpen] = useState(false)
 
   const lockWallet = () => {
+    keyring.clearCachedSecrets()
```
nop33, line 75: It deletes all cached secrets, see the implementation later.

apps/desktop-wallet/src/components/WalletSwitcher.tsx
```diff
@@ -71,7 +71,9 @@ const WalletSwitcher = ({ onUnlock }: WalletSwitcherProps) => {
         if (location.pathname !== nextPageLocation) navigate(nextPageLocation)
       }
     })
-    if (passphrase) setPassphrase('')
+
+    setPassphrase('')
+    password = ''
```
nop33, line 76: Reassigning empty strings to function parameters is an approach Metamask is using: https://github.com/MetaMask/metamask-mobile/blob/c575f53/app/core/Authentication/Authentication.ts#L156-L164

apps/desktop-wallet/src/contexts/wallet.tsx
```diff
@@ -16,53 +16,27 @@ You should have received a copy of the GNU Lesser General Public License
...
-  plainWallet?: Wallet | undefined
-  setPlainWallet: (w: Wallet | undefined) => void
-  mnemonic: string
-  setMnemonic: (m: string) => void
-  walletName: string
-  setWalletName: (w: string) => void
-  password: string
-  setPassword: (password: string) => void
+  mnemonic: Buffer | null
+  setMnemonic: (mnemonic: Buffer | null) => void
```
nop33, line 25: The wallet creation flow is the only place we are caching the mnemonic in the desktop wallet. This is because we have multiple screens to see the mnemonic, verify it, and give a password to encrypt and store it.

apps/desktop-wallet/src/hooks/useAddressGeneration.tsx
```diff
@@ -52,33 +51,26 @@ interface GenerateOneAddressPerGroupProps {
...
+// const addressDiscoveryWorker = new Worker(new URL('../workers/addressDiscovery.ts', import.meta.url), {
+//   type: 'module'
+// })
+// const deriveAddressesInGroupsWorker = new Worker(new URL('../workers/deriveAddressesInGroups.ts', import.meta.url), {
+//   type: 'module'
+// })
+// const deriveAddressesFromIndexesWorker = new Worker(
+//   new URL('../workers/deriveAddressesFromIndexes.ts', import.meta.url),
+//   { type: 'module' }
+// )
```
nop33, line 63: Web workers are disabled because they don't have access to the same `keyring` instance as the rest of the app.

apps/desktop-wallet/src/modals/SecretPhraseModal.tsx
```diff
@@ -26,18 +27,37 @@ import { Section } from '@/components/PageComponents/PageContainers'
...
 import { useAppSelector } from '@/hooks/redux'
 import CenteredModal from '@/modals/CenteredModal'
+import { walletStorage } from '@/storage/wallets/walletPersistentStorage'
 
 const SecretPhraseModal = ({ onClose }: { onClose: () => void }) => {
   const { t } = useTranslation()
-  const activeWalletMnemonic = useAppSelector((state) => state.activeWallet.mnemonic)
+  const activeWalletId = useAppSelector((s) => s.activeWallet.id)
   const [isDisplayingPhrase, setIsDisplayingPhrase] = useState(false)
+  const [mnemonic, setMnemonic] = useState<string>()
```
nop33, line 36: I lied, this is the second place where we cache the mnemonic. We need to display it in the UI somehow.

apps/desktop-wallet/src/modals/SettingsModal/DevToolsSettingsSection.tsx
```diff
@@ -73,7 +74,7 @@ const DevToolsSettingsSection = () => {
     if (!selectedAddress) return
 
     try {
-      await navigator.clipboard.writeText(selectedAddress.privateKey)
+      await navigator.clipboard.writeText(keyring.exportPrivateKeyOfAddress(selectedAddress.hash).toString())
```
nop33, line 77: This is the only place that a private key leaves the keyring

apps/desktop-wallet/src/modals/WalletQRCodeExportModal.tsx
```diff
@@ -70,21 +71,42 @@ const WalletQRCodeExportModal = ({ onClose }: { onClose: () => void }) => {
...
-      addresses: addresses.map(({ index, label, color, isDefault }) => ({ index, label, color, isDefault })),
-      contacts: contacts.map(({ name, address }) => ({ name, address }))
+    try {
+      const encryptedData = encrypt(
+        password,
+        JSON.stringify({
+          mnemonic: decryptMnemonic(
+            walletStorage.load(activeWalletId).encrypted,
+            password
+          ).decryptedMnemonic.toString(),
```
nop33, line 84: The mnemonic gets decrypted before it gets encrypted again (see like 78) in a way that the mobile wallet can decrypt it.

apps/desktop-wallet/src/storage/statelessEncryptedPersistentStorage.ts
nop33, line 1: Same as above, not needed since we don't encrypt

apps/desktop-wallet/src/tests/migration.test.ts
```diff
@@ -450,17 +437,11 @@ describe('_20230209_124300', () => {
...
-    localStorage.setItem(
-      `addresses-metadata-${activeWallet.id}`,
-      JSON.stringify({
-        version: '2022-05-27T12:00:00Z',
-        encrypted: encrypt(activeWallet.mnemonic, JSON.stringify(deprecatedAddressMetadata))
-      })
-    )
+    localStorage.setItem(`addresses-metadata-${activeWallet.id}`, JSON.stringify(deprecatedAddressMetadata))
 
-    migrate._20230209_124300()
```
nop33, line 461: I hate these weird function names, so I renamed the ones I touched

packages/shared-crypto/lib/keyring.ts
nop33, line 1: Here is the juice 🍹

packages/shared-crypto/lib/keyring.ts
```diff
@@ -0,0 +1,314 @@
...
+
+type GenerateAddressProps = {
+  group?: number
+  addressIndex?: number
+  skipAddressIndexes?: number[]
+}
+
+class Keyring {
+  private hdPath = "m/44'/1234'/0'/0"
+  private root: bip32.BIP32Interface | null
```
nop33, line 46: This corresponds to the same thing as in Metamask, see below where it gets assigned: https://github.com/MetaMask/eth-hd-keyring/blob/main/index.js#L300

packages/shared-crypto/lib/keyring.ts
```diff
@@ -0,0 +1,314 @@
...
+  private addresses: SensitiveAddressData[]
```
nop33, line 47: This corresponds to this: https://github.com/MetaMask/eth-hd-keyring/blob/main/index.js#L32

packages/shared-crypto/lib/keyring.ts
```diff
@@ -0,0 +1,314 @@
...
+  public signTransaction = (txId: string, addressHash: AddressHash): string =>
+    transactionSign(txId, this.exportPrivateKeyOfAddress(addressHash).toString('hex'))
+
+  public signMessage = (message: string, addressHash: AddressHash): string =>
+    sign(message, this.exportPrivateKeyOfAddress(addressHash).toString('hex'))
+
+  public exportPrivateKeyOfAddress = (addressHash: AddressHash): SensitiveAddressData['privateKey'] =>
+    this._getAddress(addressHash).privateKey
+
+  public discoverAndCacheActiveAddresses = async (
```
nop33, line 102: No changes were made to this algorithm, it was just copied over so that it can use the secrets that should not leave the keyring

packages/shared-crypto/lib/keyring.ts
```diff
@@ -0,0 +1,314 @@
...
+
+    while (group !== undefined && addressToGroup(newAddressData.hash, TOTAL_NUMBER_OF_GROUPS) !== group) {
+      nextAddressIndex = findNextAvailableAddressIndex(newAddressData.index, skipAddressIndexes)
+      newAddressData = this._deriveAddressAndKeys(nextAddressIndex)
+    }
+
+    return newAddressData
+  }
+
+  private _deriveAddressesInGroup = (group: number, amount: number, skipIndexes: number[]): SensitiveAddressData[] => {
```
nop33, line 202: Same as above

packages/shared-crypto/lib/keyring.ts
```diff
@@ -0,0 +1,314 @@
...
+    passphrase = ''
+  }
+
+  private _getPath = (addressIndex: number = 0) => {
+    if (!isAddressIndexValid(addressIndex)) throw new Error('Invalid address index path level')
+
+    return `${this.hdPath}/${addressIndex}`
+  }
+
+  private _deriveAddressAndKeys = (addressIndex: number): SensitiveAddressData => {
```
nop33, line 239: This was copied over as well

packages/shared-crypto/lib/keyring.ts
```diff
@@ -0,0 +1,314 @@
...
+  }: SensitiveAddressData): NonSensitiveAddressData => ({ hash, index, publicKey })
+
+  private _initFromMnemonic = (mnemonic: Buffer, passphrase: string) => {
+    if (this.root) throw new Error('Keyring: Secret recovery phrase already provided')
+
+    const isValid = bip39.validateMnemonic(mnemonic.toString())
+    if (!isValid) throw new Error('Keyring: Invalid secret recovery phrase provided')
+
+    const seed = bip39.mnemonicToSeedSync(mnemonic.toString(), passphrase)
+    this.root = bip32.fromSeed(seed)
```
nop33, line 228: This is where we cache the `root` that is used for address generation, just like Metamask: https://github.com/MetaMask/eth-hd-keyring/blob/main/index.js#L300

apps/desktop-wallet/src/contexts/wallet.tsx
mvaivre: Side note for other potential reviewers: when we studied other wallets, the mnemonic was inspectable in memory during wallet creation in all of them. This seems to be inevitable.

apps/desktop-wallet/src/modals/SecretPhraseModal.tsx
mvaivre: Again, inevitable.

packages/keyring/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "clean:node-modules": "rm -rf node_modules",
+    "typecheck": "tsup"
+  },
+  "dependencies": {
+    "@alephium/shared": "workspace:*",
+    "@alephium/shared-crypto": "workspace:*",
+    "@alephium/web3": "0.29.1",
+    "@metamask/scure-bip39": "^2.0.3",
+    "blakejs": "^1.1.1",
+    "ethereum-cryptography": "^2.1.2"
```
nop33, line 26: To eliminate all possibilities I used the same package as Metamask extension does: 
- https://github.com/MetaMask/metamask-extension/blob/v11.13.1/package.json#L292
- https://github.com/MetaMask/eth-hd-keyring/blob/v7.0.1/package.json#L32

packages/keyring/src/keyring.ts
```diff
@@ -0,0 +1,333 @@
...
+
+  public clearCachedSecrets = () => {
+    this.addresses = []
+    this.hdWallet = null
+    this.root = null
+  }
+
+  public generateRandomMnemonic = (mnemonicLength: MnemonicLength = 24): Uint8Array => {
+    // const strength = mnemonicLength === 24 ? 256 : 128
+    const mnemonic = generateMnemonic(wordlist)
```
nop33, line 67: The `generateMnemonic` function of `@scure/bip39` does not accept mnemonic length as the one from `bip39` does.

packages/keyring/src/keyring.ts
```diff
@@ -0,0 +1,333 @@
...
+  }: SensitiveAddressData): NonSensitiveAddressData => ({ hash, index, publicKey })
+
+  private _initFromMnemonic = (mnemonic: Uint8Array | null, passphrase: string) => {
+    if (this.root) throw new Error('Keyring: Secret recovery phrase already provided')
+    if (!mnemonic) throw new Error('Keyring: Secret recovery phrase not provided')
+
+    const isValid = validateMnemonic(mnemonic, wordlist)
+    if (!isValid) throw new Error('Keyring: Invalid secret recovery phrase provided')
+
+    const seed = mnemonicToSeedSync(mnemonic, wordlist, passphrase)
```
nop33, line 234: This is where the mnemonic gets leaked (it is turned into a string in the implementation of this function: https://github.com/MetaMask/scure-bip39/blob/v2.1.0/src/index.ts#L142-L177

I don't understand why on Metamask it doesn't. It uses the same code: https://github.com/MetaMask/eth-hd-keyring/blob/v7.0.1/index.js#L298

packages/keyring/src/keyring.ts
```diff
@@ -0,0 +1,333 @@
...
+    if (!keyPair.privateKey) throw new Error('Keyring: Missing private key')
+
+    const publicKey = bytesToHex(keyPair.publicKey)
+    const privateKey = keyPair.privateKey
+    const hash = blake.blake2b(Uint8Array.from(keyPair.publicKey), undefined, 32)
+    const type = new Uint8Array([0])
+    const bytes = new Uint8Array(type.length + hash.length)
+    bytes.set(type, 0)
+    bytes.set(hash, type.length)
+    const address = bs58.encode(bytes)
```
nop33, line 258: This is an important part, we used to use `Buffer`, but to make it work with the bip packages that Metamask uses it had to (rightfully) be changed to `Uint8Array`.

packages/keyring/src/mnemonic.ts
```diff
@@ -0,0 +1,100 @@
...
+export type EncryptedMnemonicVersion = 1 | 2
+
+export type DecryptMnemonicResult = {
+  decryptedMnemonic: Uint8Array
+  version: EncryptedMnemonicVersion
+}
+
+export class EncryptedMnemonicStoredAsUint8Array {
+  readonly version: EncryptedMnemonicVersion = 2
+  readonly mnemonic: Uint8Array
```
nop33, line 33: Storing the mnemonic in local storage as a `Uint8Array` now, instead of a `string`.

packages/keyring/src/mnemonic.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+}
+
+// It will convert the mnemonic from Uint8Array to string, leaking it to the memory. Use only when absolutely needed,
+// ie: displaying the mnemonic for backup, etc
+export const dangerouslyConvertUint8ArrayMnemonicToString = (mnemonic: Uint8Array) => {
+  const recoveredIndices = Array.from(new Uint16Array(new Uint8Array(mnemonic).buffer))
+
+  return recoveredIndices.map((i) => wordlist[i]).join(' ')
+}
```
nop33, line 46: Same implementation as on Metamask: https://github.com/MetaMask/eth-hd-keyring/blob/v7.0.1/index.js#L40-L45

packages/keyring/src/mnemonic.ts
```diff
@@ -0,0 +1,100 @@
...
+  return new Uint8Array(new Uint16Array(indices).buffer)
+}
+
+// When JSON.stringify an Uint8Array it becomes a JS object that we need to cast back to an Uint8Array
+const mnemonicJsonStringifiedObjectToUint8Array = (mnemonic: unknown): Uint8Array => {
+  if (!(mnemonic instanceof Object))
+    throw new Error('Could not convert stringified Uint8Array mnemonic back to Uint8Array')
+
+  return Uint8Array.from(Object.values(mnemonic))
+}
```
nop33, line 100: Same as on Metamask: https://github.com/MetaMask/eth-hd-keyring/blob/v7.0.1/index.js#L76-L77

apps/desktop-wallet/src/hooks/useAddressGeneration.tsx
```diff
@@ -87,7 +79,7 @@ const useAddressGeneration = () => {
       (group) => !skipGroups.includes(group)
     )
 
-    deriveAddressesInGroupsWorker.onmessage = ({ data }: { data: (AddressKeyPair & { group: number })[] }) => {
+    const onAddressesDerivedCallback = ({ data }: { data: (NonSensitiveAddressData & { group: number })[] }) => {
```
mvaivre, line 82: Seems like the `group` prop isn't used.

mvaivre (commented):
Review part 1 of XX 😅 👏

apps/desktop-wallet/src/pages/NewWallet/CreateWalletPage.tsx
```diff
@@ -62,7 +77,8 @@ const CreateWalletPage = ({ isRestoring = false }: { isRestoring?: boolean }) =>
         passwordError = t`Insecure password`
```
mvaivre, line 77: Nit: not part of this PR but signalling because I'm seeing it :)  

```ts
t('Password is too weak')
t('Insecure password')
```

apps/desktop-wallet/src/workers/addressDiscovery.ts
```diff
@@ -16,27 +16,28 @@ You should have received a copy of the GNU Lesser General Public License
...
-  discover(masterKey, client, skipIndexes)
+  discover(client, skipIndexes)
 }
 
-const discover = async (masterKey: Wallet['masterKey'], client: ExplorerProvider, skipIndexes?: number[]) => {
-  const activeAddresses = await discoverActiveAddresses(masterKey, client, skipIndexes)
-  self.postMessage(activeAddresses)
+const discover = async (client: ExplorerProvider, skipIndexes?: number[]) => {
+  try {
+    const activeAddresses = await keyring.discoverAndCacheActiveAddresses(client, skipIndexes)
```
mvaivre, line 38: Q: So this works inside of the worker? Wasn't there an issue with the keyring object not being the same in this context?

apps/desktop-wallet/src/workers/deriveAddressesFromIndexes.ts
```diff
@@ -16,22 +16,22 @@ You should have received a copy of the GNU Lesser General Public License
...
-  derive(masterKey, indexesToDerive)
+self.onmessage = ({ data: { indexesToDerive } }: WorkerPayload) => {
+  derive(indexesToDerive)
 }
 
-const derive = (masterKey: Wallet['masterKey'], indexesToDerive: number[]) => {
-  self.postMessage(indexesToDerive.map((index) => deriveNewAddressData(masterKey, undefined, index)))
+const derive = (indexesToDerive: number[]) => {
+  try {
+    self.postMessage(indexesToDerive.map((index) => keyring.generateAndCacheAddress({ addressIndex: index })))
```
mvaivre, line 33: Same question.

mvaivre (commented):
Part 2 🌷

packages/keyring/src/keyring.ts
```diff
@@ -0,0 +1,319 @@
...
+    }
+  }
+
+  return {
+    gap,
+    activeAddresses
+  }
+}
+
+const getActiveAddressesResults = async (
```
mvaivre, line 303: Maybe it's just me but I find it weird to have a call to an endpoint in the keyring package... 🤔

packages/keyring/src/keyring.ts
```diff
@@ -0,0 +1,319 @@
...
+
+      let gapPerGroup = gap
+      activeAddresses.push(...newActiveAddresses)
+
+      while (gapPerGroup < minGap) {
+        const remainingGap = minGap - gapPerGroup
+        const newAddresses = this._deriveAddressesInGroup(group, remainingGap, skipIndexes)
+        skipIndexes.push(...newAddresses.map((address) => address.index))
+
+        const addressesToCheckIfActive = newAddresses.map((address) => address.hash)
```
mvaivre, line 150: NIT: I'd rename this to `newAddressesToCheckIfActive` to make sure it's not confused with the variable with the same name above.

packages/keyring/src/keyring.ts
```diff
@@ -0,0 +1,319 @@
...
+  public signTransaction = (txId: string, addressHash: AddressHash): string =>
+    transactionSign(txId, this.exportPrivateKeyOfAddress(addressHash))
+
+  public signMessage = (message: string, addressHash: AddressHash): string =>
+    sign(message, this.exportPrivateKeyOfAddress(addressHash))
+
+  public exportPrivateKeyOfAddress = (addressHash: AddressHash): string =>
+    bytesToHex(this._getAddress(addressHash).privateKey)
+
+  public discoverAndCacheActiveAddresses = async (
```
mvaivre, line 117: Linked to my previous comment about (not) calling endpoints in the keyring, maybe this method should be moved to somewhere else. But I see why you did this, it feels more confined and handy to have this available in the keyring package...
Is Metamask doing something similar?

mvaivre (commented):
🚀

apps/desktop-wallet/src/components/AppHeader.tsx
nop33: Hm, for such a small logic (`!!`) I don't think it's necessary, but I really don't have a strong opinion on this, I can do it.

apps/desktop-wallet/src/components/Buttons/ShortcutButtons.tsx
nop33: Good call. I'll look into it.

apps/desktop-wallet/src/hooks/useAddressGeneration.tsx
nop33: FWIW, the web workers were not making things go faster. They were just not blocking the UI thread from rendering.

apps/desktop-wallet/src/contexts/global.tsx
```diff
@@ -82,58 +82,73 @@ export const GlobalContextProvider: FC<{ overrideContextValue?: PartialDeep<Glob
...
+    const { event, walletId, afterUnlock } = props
+    let { password, passphrase } = props
+    let encryptedWallet: StoredEncryptedWallet | null
+    let initialAddress: NonSensitiveAddressData
+    let version: EncryptedMnemonicVersion
 
     try {
-      wallet = WalletStorage.load(walletId, password)
+      encryptedWallet = walletStorage.load(walletId)
+      version = keyring.initFromEncryptedMnemonic(encryptedWallet.encrypted, password, passphrase ?? '')
```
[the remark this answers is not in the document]
nop33: I agree that it's not clear what the structure of the JSON is. I also want to improve this. Not sure `encryptedMnemonic` is the way to go. I'll do some research.

apps/desktop-wallet/src/hooks/useAddressGeneration.tsx
```diff
@@ -52,33 +51,26 @@ interface GenerateOneAddressPerGroupProps {
...
   const currentAddressIndexes = addresses.map(({ index }) => index)
 
-  const generateAddress = ({ group }: GenerateAddressProps = {}): AddressKeyPair => {
-    if (!mnemonic) throw new Error('Could not generate address, mnemonic not found')
-
-    const { masterKey } = getWalletFromMnemonic(mnemonic, passphrase)
-
-    return deriveNewAddressData(masterKey, group, undefined, currentAddressIndexes)
-  }
+  const generateAddress = ({ group }: GenerateAddressProps = {}): NonSensitiveAddressData =>
```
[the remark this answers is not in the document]
nop33: I can't think of a reason why! Old code

apps/desktop-wallet/src/workers/addressDiscovery.ts
nop33: Even though the web workers code has been updated, the use of it is commented out.

packages/keyring/src/keyring.ts
```diff
@@ -0,0 +1,319 @@
...
+
+  public clearCachedSecrets = () => {
+    this.addresses.map((address) => resetArray(address.privateKey))
+    this.addresses = []
+    this.hdWallet = null
+    this.root = null
+  }
+
+  public generateRandomMnemonic = (mnemonicLength: MnemonicLength = 24): Uint8Array => {
+    const strength = mnemonicLength === 24 ? 256 : 128
```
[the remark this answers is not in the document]
nop33: But we are typing the `MnemonicLength`, it can't accept anything but `12` or `24`.

packages/keyring/src/keyring.ts
```diff
@@ -0,0 +1,319 @@
...
+    return mnemonic
+  }
+
+  public importMnemonicString = (mnemonicStr: string): Uint8Array => {
+    if (!mnemonicStr) throw new Error('Keyring: Cannot import mnemonic, mnemonic not provided')
+
+    if (!metamaskBip39.validateMnemonic(mnemonicStr, wordlist))
+      throw new Error('Keyring: Cannot import mnemonic, invalid mnemonic provided')
+
+    console.log('Mnemonic leaked to memory as a string while importing wallet. This is expected.')
```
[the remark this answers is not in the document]
nop33: The idea behind it was for me to not pull my hair out when checking the processes on Windows to see if the mnemonic there or not. I can remove it now.

packages/keyring/src/keyring.ts
nop33: Hm, yeah. This code is a simple adaptation of what we had before in the shared-crypto. I'll look into alternative solutions (maybe passing the keyring as a callback)

packages/keyring/src/keyring.ts
nop33: I did not see any address discovery method in the keyring in Metamask.

nop33: Addressed in 1ce954a11 in PR https://github.com/alephium/alephium-frontend/pull/504

packages/keyring/src/keyring.ts
nop33: Addressed in d39c6f560 in PR https://github.com/alephium/alephium-frontend/pull/504

packages/keyring/src/keyring.ts
nop33: Addressed in https://github.com/alephium/alephium-frontend/commit/1ce954a1147ac03ee987bfd9c7ac33149c1577c6 in PR https://github.com/alephium/alephium-frontend/pull/504

packages/keyring/src/keyring.ts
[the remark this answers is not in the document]
nop33: Addressed in 355c70f24 in PR #504

apps/desktop-wallet/src/workers/deriveAddressesFromIndexes.ts
nop33: Removed web workers completely in 0772f5c42 in PR #504

apps/desktop-wallet/src/workers/addressDiscovery.ts
nop33: Removed web workers completely in 0772f5c42 in PR #504

apps/desktop-wallet/src/pages/NewWallet/CreateWalletPage.tsx
nop33: Fixed in 153e081b1 in PR #504

apps/desktop-wallet/src/hooks/useAddressGeneration.tsx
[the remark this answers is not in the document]
nop33: Addressed in daa663dd1 in PR #504

apps/desktop-wallet/src/components/AppHeader.tsx
nop33: Addressed in 2662fbd3f in PR #504

apps/desktop-wallet/src/components/Buttons/ShortcutButtons.tsx
nop33: I found a much better solution that combines locking, unlocking, and `isAuthenticated`, all in one hook called `useWalletLock`. Check out 6e4cc0590 in PR #504

apps/desktop-wallet/src/contexts/global.tsx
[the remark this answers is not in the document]
nop33: I added a comment, let me know if it helps: 95cdbf6bf (in PR #504)

I'd prefer to avoid changing the structure of the stored wallet once again, to not have to write yet another migration.

Also, if in the future we want to store additional encrypted data alongside the mnemonic, then the name would be misleading and we'd need a migration again.

packages/keyring/src/keyring.ts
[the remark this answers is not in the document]
nop33: We could use TypeError for checking if the index number or a group number is valid when generating an address for example

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError

nop33:
@mvaivre I believe I addressed all you comments, I hope I didn't miss something!

packages/keyring/src/keyring.ts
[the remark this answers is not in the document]
mvaivre: exactly

apps/desktop-wallet/src/components/Buttons/ShortcutButtons.tsx
mvaivre: I like it a lot! 💯

apps/desktop-wallet/src/contexts/global.tsx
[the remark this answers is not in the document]
mvaivre: fine by me, thanks for adding the comment.

mvaivre (approved):
👏👏👏👏👏👏👏👏

PR #477, opened 2024-03-27
Simplify WC button rendering logic

nop33:


nop33:
@mvaivre since we agreed to move from internal to beta, and since you said the code looks good, I am merging this to build the iOS beta from `next`.

PR #502, opened 2024-04-07
Use @metamask/browser-passworder

nop33:
This PR introduces a new internal package called `@alephium/encryptor`. It is a proxy to `@metamask/browser-passworder` with `Buffer` polyfilled. 

Ref: https://github.com/alephium/js-sdk/pull/133#issuecomment-1328787203


packages/encryptor/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { decrypt as metamaskDecrypt, encrypt, EncryptionResult } from '@metamask/browser-passworder'
+import { Buffer } from 'buffer'
+
+// Polyfill Buffer if we are in a browser environment
+if (typeof window !== 'undefined') window.Buffer = Buffer
```
mvaivre, line 23: Q: Why do that here in the encryptor package?

mvaivre (commented):
Code LGTM 🚀  but I haven't tested it yet.

packages/encryptor/src/index.ts
nop33: That's a good question!

So, I tried using `@metamask/browser-passworder` directly on the desktop wallet. However, I got the error that `Buffer is undefined` (so much for "browser"). This means it needs polyfilling to be used on the desktop wallet. Potentially that'd also be the case for mobile wallet, or any other app we might wanna use it. Because of this (and because of the need for the small utilities that I wrote) I decided to go for our own package and polyfill it only once.

To more specifically answer your question: The above line detects if we are in a browser environment (like React for example). In that case, it polyfills Buffer with the one coming from the NPM package `buffer`. Otherwise (if we are running the test suite for example in a node environment) it won't polyfill it.

mvaivre (approved):

PR #508, opened 2024-04-09
Improve secrets management in desktop wallet

nop33:


packages/encryptor/src/index.ts
nop33, line 1: 👇 This package re-exports the `encrypt` function of `@metamask/browser-passworder` and wraps the `decrypt` one in order to handle decryption of the old format that was using `@alephium/shared-crypto` for encryption.

ref: https://github.com/alephium/js-sdk/pull/133#issuecomment-1328787203

packages/keyring/src/keyring.ts
```diff
@@ -0,0 +1,211 @@
...
+    const bytes = new Uint8Array(type.length + hash.length)
+    bytes.set(type, 0)
+    bytes.set(hash, type.length)
+    const address = bs58.encode(bytes)
+
+    return { hash: address, publicKey, privateKey, index: addressIndex }
+  }
+}
+
+export const keyring = new Keyring()
```
nop33, line 211: It is in this class instance that the secrets (`root`, `privateKey`) are cached, as per Metamask: https://github.com/MetaMask/eth-hd-keyring/blob/main/index.js

packages/keyring/src/keyring.ts
```diff
@@ -0,0 +1,211 @@
...
+    if (!keyPair.privateKey) throw new Error('Keyring: Missing private key')
+
+    const publicKey = bytesToHex(keyPair.publicKey)
+    const privateKey = keyPair.privateKey
+    const hash = blake.blake2b(Uint8Array.from(keyPair.publicKey), undefined, 32)
+    const type = new Uint8Array([0])
+    const bytes = new Uint8Array(type.length + hash.length)
+    bytes.set(type, 0)
+    bytes.set(hash, type.length)
+    const address = bs58.encode(bytes)
```
nop33, line 205: We used to generate this info using `Buffer`, now we use `Uint8Array`: https://github.com/alephium/alephium-frontend/blob/master/packages/shared-crypto/lib/wallet.ts#L111C1-L118C1

```ts
  const publicKey = keyPair.publicKey.toString('hex')
  const privateKey = keyPair.privateKey.toString('hex')
  const hash = blake.blake2b(Uint8Array.from(keyPair.publicKey), undefined, 32)
  const pkhash = Buffer.from(hash)
  const type = Buffer.from([0])
  const bytes = Buffer.concat([type, pkhash])
  const address = bs58.encode(bytes)
```

mvaivre (approved):
Alrighhhtttt... Let's start testing this beauty more widely 👏 🚀

nop33:
Since master was ahead of next (some commits were added to master specifically that did not exist on next), I had to merge master into next. After doing that there was a conflict in this PR so I merged next into next-secrets and pushed.

PR #514, opened 2024-04-12
Add devnet to networkTypes

h0ngcha0:


mvaivre (approved):

nop33:
@mvaivre did you check all places we use `networkTypes` in our apps to make sure nothing breaks? Our apps have special handling of `devnet` so I am not 100% sure that simply adding it to the array is sufficient. Just checking

nop33:
Ah, it's only explorer, sorry. Let's make use of the `@alephium/shared`: https://github.com/alephium/alephium-frontend/blob/master/packages/shared/src/types/network.ts#L34-L41

polarker:
@nop33 @mvaivre Please take over this PR and add the necessary changes based on @h0ngcha0 's work. It's more efficient in that way.

nop33:
I've adapted the PR to make use of `@alephium/shared` for both the DW and explorer. I've also fixed the display of the current network when the type is `devnet` (it used to display "Testnet" before). @mvaivre, ready for review.

apps/desktop-wallet/src/utils/walletConnect.ts
```diff
@@ -16,11 +16,10 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { NetworkSettings, networkSettingsPresets } from '@alephium/shared'
+import { NetworkPreset, NetworkSettings, networkSettingsPresets } from '@alephium/shared'
```
mvaivre, line 19: NIT: Not linked to this PR, but `NetworkNamePreset` would maybe be less confusing that `NetworkPreset` (as it's a subset of `NetworkName`).

package.json
```diff
@@ -40,7 +40,8 @@
     "overrides": {
       "ip@<=2.0.0": ">=2.0.1",
       "minimatch@<3.0.5": ">=3.0.5",
-      "undici@<=5.28.2": ">=5.28.3"
+      "undici@<=5.28.2": ">=5.28.3",
+      "tar@<6.2.1": ">=6.2.1"
```
mvaivre, line 44: Why was this necessary?

nop33: because of this: https://github.com/alephium/alephium-frontend/actions/runs/8777788452/job/24083242757

apps/desktop-wallet/src/utils/walletConnect.ts
nop33: Maybe `NetworkPresetName` since it's the name of a preset rather than a preset of a name?

nop33:
Since only "nit" remains, I'll merge this to proceed with the release plan.

PR #515, opened 2024-04-16
Fixes before 2.2.5 release & add mp4 NFT display

mvaivre:


apps/desktop-wallet/src/components/AssetLogo.tsx
```diff
@@ -18,28 +18,40 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
-      <Initials size={size}>{assetName.slice(0, 2)}</Initials>
-    ) : (
-      <HelpCircle size={size} />
-    )}
-  </div>
-)
+const AssetLogo = ({ assetImageUrl, size, assetName, className }: AssetLogoProps) => {
+  let VideoPlayer
+
+  if (assetImageUrl?.includes('.mp4')) {
```
nop33, line 35: I think `.endsWith` would be safer.

apps/desktop-wallet/src/components/AssetLogo.tsx
```diff
@@ -18,28 +18,40 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+          <LogoImage src={assetImageUrl} />
+        )
+      ) : assetName ? (
+        <Initials size={size}>{assetName.slice(0, 2)}</Initials>
+      ) : (
+        <HelpCircle size={size} />
+      )}
+    </div>
+  )
+}
```
nop33, line 54: What do you think of this implementation? I find it simpler.

```tsx
const AssetLogo = ({ assetImageUrl, size, assetName, className }: AssetLogoProps) => (
  <div className={className}>
    {assetImageUrl?.endsWith('.mp4') ? (
      <ReactPlayer url={assetImageUrl} muted width={size} height={size} />
    ) : assetImageUrl ? (
      <LogoImage src={assetImageUrl} />
    ) : assetName ? (
      <Initials size={size}>{assetName.slice(0, 2)}</Initials>
    ) : (
      <HelpCircle size={size} />
    )}
  </div>
)
```

apps/desktop-wallet/src/components/NFTThumbnail.tsx
```diff
@@ -38,7 +39,15 @@ const NFTThumbnail = ({ nftId, size = '100', ...props }: NFTThumbnailProps) => {
...
-  return nft.image && !error ? (
+  let VideoPlayer
+
+  if (nft.image && nft.image.includes('.mp4')) {
+    VideoPlayer = <ReactPlayerStyled url={nft.image} playing loop muted width={size} height={size} />
+  }
+
+  return VideoPlayer ? (
+    VideoPlayer
+  ) : nft.image && !error ? (
```
nop33, line 50: Similar with above, all these lines can be replaced with:

```tsx
  return nft.image?.endsWith('.mp4') ? (
    <ReactPlayerStyled url={nft.image} playing loop muted width={size} height={size} />
  ) : nft.image && !error ? (
```

nop33 (commented):
Nice improvements! I like what you did in the transaction list of the address details modal where you stacked the direction keyword on top of the list of addresses, smart move. 🧠 👏 

I've added some comments small code improvement comments.

A small issue I found was that the tooltip of the NFT ID is wrong, it says "Copy address", I guess it should say "Copy ID"?:

image


Another minor thing (which I am not sure if it's introduced here or not) is that the wallet addresses get truncated a lot when the window is in its default size. Only the first and last characters are visible:

image

nop33 (approved):

PR #516, opened 2024-04-22
[2/3] Improve secrets management in mobile wallet: Authentication flow

nop33:


apps/mobile-wallet/src/contexts/SendContext.tsx
```diff
@@ -72,7 +72,8 @@ const initialValues: SendContextValue = {
 const SendContext = createContext(initialValues)
 
 export const SendContextProvider = ({ children }: { children: ReactNode }) => {
-  const requiresAuth = useAppSelector((s) => s.settings.requireAuth)
+  const biometricsRequiredForTransactions = useAppSelector((s) => s.settings.requireAuth)
```
nop33, line 75: I have not yet changed the name of the following settings yet because it will create unnecessary additional code and migrations. I would rather keep the scope as minimal as possible for now.

```
requireAuth    => biometricsRequiredForTransactions
usesBiometrics => biometricsRequiredForAppAccess
```

apps/mobile-wallet/src/persistent-storage/wallet.ts
```diff
@@ -320,6 +299,7 @@ export const migrateDeprecatedMnemonic = async (deprecatedMnemonic: string) => {
   await SecureStore.setItemAsync('wallet-mnemonic-v2', mnemonicStringified, defaultSecureStoreConfig)
   await SecureStore.deleteItemAsync(PIN_WALLET_STORAGE_KEY, defaultSecureStoreConfig)
   await SecureStore.deleteItemAsync(BIOMETRICS_WALLET_STORAGE_KEY, defaultSecureStoreConfig)
+  await SecureStore.setItemAsync('is-mnemonic-migrated-to-v2', 'true')
```
nop33, line 302: This is not yet used, but might be useful.

apps/mobile-wallet/src/screens/Settings/SettingsScreen.tsx
```diff
@@ -308,16 +262,46 @@ const SettingsScreen = ({ navigation, ...props }: ScreenProps) => {
...
+          isOpen={isCurrencySelectModalOpen}
+          onClose={() => setIsCurrencySelectModalOpen(false)}
+          Content={(props) => <CurrencySelectModal onClose={() => setIsCurrencySelectModalOpen(false)} {...props} />}
+        />
+
+        <BottomModal
+          isOpen={isMnemonicModalVisible}
+          onClose={() => setIsMnemonicModalVisible(false)}
+          Content={(props) => <MnemonicModal {...props} />}
+        />
```
nop33, line 304: This is just the same code, but moved, to fix a glitch when opening a modal right after closing another one.

apps/mobile-wallet/src/contexts/SendContext.tsx
mvaivre: So when do you suggest we proceed to the rename? Following release?

nop33: Separate PR and yes, also following release.

PR #534, opened 2024-04-23
Privacy manifest

mvaivre:
~WIP: I get a white screen on iOS.~

~Series of errors in the native logs:~

~Invariant Violation: Failed to call into JavaScript module method RCTLog.logIfNoNativeHook(). Module has not been registered as callable. Bridgeless Mode: false. Registered callable JavaScript modules (n = 0): .
A frequent cause of the error is that the application entry file path is incorrect. This can also happen when the JS bundle is corrupt or there is an early initialization error when loading React Native., js engine: hermes~

Edit: reanimated was the culprit. I had to downgrade it, to be investigated later. Hopefully this package doesn't need any of the privacy reasons. 

--- 

EDIT ILIAS: Closes #489

PR #540, opened 2024-04-25
Mw privacy manifest ilias 2

nop33:
1. I first run `pnpm dlx expo prebuild --clean` which created this diff: https://github.com/alephium/alephium-frontend/commit/adca479d6efe905933b3d6935a76ffaecfbfb54e. Here I explain my concerns regarding this diff: https://github.com/alephium/alephium-frontend/pull/534#issuecomment-2076253419. So, I picked only the changes that I thought were necessary into the first commit of this PR.
2. Then I replaced the local path with a relative one, manually, into the second commit
3. Then I run pod install and I noticed that it removed the double quotes from the local path (that's a good sign?) into the 3rd commit.

---

My remaining concerns are still there:
1. How do we avoid re-adding by accident the local file when we run prebuild? We need to be careful until hopefully a future version of expo fixes it.
2. We should try updating our deps and testing to see whether the privacy manifest is still needed.


nop33:
This is what I get when I try to run the app...

image

nop33:
To make sure the above is not an issue, I uninstalled the app, checked out `@alephium/mobile-wallet@1.0.9`, installed the app, created a wallet, checked out this branch, and launched the app again. It seems to work, I don't get the blank screen anymore.

nop33:
I tried to submit an app on TestFlight to see if we get the email again, but I had to update the version number to 1.0.10 first. Tryign again...

mvaivre (approved):

PR #544, opened 2024-04-25
Secrets fixes

### LINKED ISSUE #543: Cannot copy private key after upgrading to 2.3.0
> Error: Invalid version: got undefined, expected: 1

nop33:
These changes were already included in the mobile secrets PRs, so I cherry-picked them to fix #543 

Closes #543

### DIFF
--- a/apps/desktop-wallet/src/modals/SettingsModal/WalletsSettingsSection.tsx
+++ b/apps/desktop-wallet/src/modals/SettingsModal/WalletsSettingsSection.tsx
@@ -81,6 +81,7 @@ const WalletsSettingsSection = () => {
               wallet={wallet}
               isCurrent={wallet.id === activeWallet.id}
               onWalletDelete={setWalletToRemove}
+              isPassphraseUsed={wallet.id === activeWallet.id && activeWallet.isPassphraseUsed}
             />
           ))}
         </BoxContainerStyled>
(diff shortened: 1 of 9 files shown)


nop33:
Pushed one more fix, sorry

apps/desktop-wallet/src/modals/SettingsModal/WalletsSettingsSection.tsx
```diff
@@ -108,15 +109,40 @@ const WalletsSettingsSection = () => {
...
               {t('Lock current wallet')}
             </Button>
-            <Button role="secondary" onClick={() => setIsQRCodeModalVisible(true)}>
-              {t('Export current wallet')}
-            </Button>
+
+            <ButtonTooltipWrapper
+              data-tooltip-id="default"
+              data-tooltip-content={
+                activeWallet.isPassphraseUsed ? t('To export this wallet use it without a passphrase') : ''
```
nop33, line 116: To avoid confusion, passphrase wallets should not be exportable, since we are not exporting the passphrase. The user ends up with a different wallet on their mobile (the one with the 24 words)

apps/desktop-wallet/src/modals/SettingsModal/WalletsSettingsSection.tsx
```diff
@@ -108,15 +109,40 @@ const WalletsSettingsSection = () => {
...
               {t('Show your secret recovery phrase')}
             </Button>
-            <Button variant="alert" onClick={() => setWalletToRemove(activeWallet as ActiveWallet)}>
-              {t('Remove current wallet')}
-            </Button>
+
+            <ButtonTooltipWrapper
+              data-tooltip-id="default"
+              data-tooltip-content={
+                activeWallet.isPassphraseUsed ? t('To delete this wallet use it without a passphrase') : ''
```
nop33, line 135: I think it makes sense to disable deleting the wallet when it's unlocked with a passphrase. USers might think that they are deleting the "passphrase-enabled" wallet, where in fact they are deleting the 24-words wallet. We don't store any "passphrase-enabled" wallet, so there's nothing to delete.

mvaivre: Very valid point!

apps/desktop-wallet/src/modals/SettingsModal/WalletsSettingsSection.tsx
mvaivre: 💯

mvaivre (approved):
LGTM!

PR #548, opened 2024-04-26
Store private keys and remove cached secrets

nop33:


apps/mobile-wallet/src/persistent-storage/wallet.ts
```diff
@@ -283,23 +213,42 @@ export const storeIsNewWallet = async (isNew: boolean) => {
...
 
-  if (!decryptedMnemonic) throw new Error('Could not initialize wallet: could not find stored wallet')
+  try {
+    await storeWalletMnemonic(mnemonicUint8Array)
 
-  const parsedDecryptedMnemonic = mnemonicJsonStringifiedObjectToUint8Array(JSON.parse(decryptedMnemonic))
-  keyring.initFromDecryptedMnemonic(parsedDecryptedMnemonic, '')
-}
+    // Step 2: Delete old mnemonic records
+    await SecureStore.deleteItemAsync(PIN_WALLET_STORAGE_KEY, defaultSecureStoreConfig)
```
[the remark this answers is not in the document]
nop33: If an error is thrown when deleting records, it will get forwarded (see what's inside the `catch`) and eventually displayed to the user. Same if an error is thrown when storing a new record. Besides, the store methods in this function have their own try/catch which reports the error in the analytics before forwarding it.

apps/mobile-wallet/__tests__/wallet.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    expect(wallet.name).toEqual('Test wallet')
+  })
+})
+
+describe(migrateDeprecatedMnemonic, () => {
+  it('should throw an error if there is no wallet metadata stored', async () => {
+    await expect(() => migrateDeprecatedMnemonic(testWalletMnemonic)).rejects.toThrow()
+  })
+
+  it('should migrate mneminic and delete old entries', async () => {
```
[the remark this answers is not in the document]
nop33: Fixed

apps/mobile-wallet/src/persistent-storage/wallet.ts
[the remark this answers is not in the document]
mvaivre: Alright. Thanks for your answer!

apps/mobile-wallet/src/persistent-storage/wallet.ts
```diff
@@ -311,3 +260,108 @@ export const dangerouslyExportWalletMnemonic = async (): Promise<string> => {
...
+    await generateAndStoreAddressKeypairForIndex(address.index)
+    key = await SecureStore.getItemAsync(storageKey, defaultSecureStoreConfig)
+
+    if (!key) throw new Error(`Could not generate address ${keyType} key`)
+  }
+
+  return key
+}
+
+const generateWalletMetadata = (name: string, firstAddressHash: string, isMnemonicBackedUp = false) => ({
```
mvaivre, line 283: Very nit: I'd call this `generateInitialWalletMetadata`

apps/mobile-wallet/src/store/addresses/addressesStorageUtils.ts
```diff
@@ -27,15 +28,23 @@ import { persistAddressesSettings } from '~/utils/addresses'
...
+
+      addressHashes.push(newAddress.hash)
+    }
 
-  store.dispatch(syncLatestTransactions(addressHashes))
+    store.dispatch(syncLatestTransactions(addressHashes))
+  } catch (e) {
+    throw new Error(getHumanReadableError(e, ''))
+  } finally {
+    keyring.clearAll()
```
mvaivre, line 48: Just thinking about this now, what about renaming the `clearAll` method to simply `clear`, as we essentially clear the whole keyring and we don't have other more granular `clearSomething` methods?

mvaivre (commented):
Code is really clear! Great job!

apps/mobile-wallet/src/store/addresses/addressesStorageUtils.ts
nop33: We have the `keyring.clearCachedSecrets`. Once the mobile wallet changes get merged, there won't be a need for this anymore and we can indeed remove it in favor of just `clear`. Since we are not caching secrets for mobile anymore, the code for the keyring can be reverted to the more simplified one. To keep this PR short, I did not go ahead with these changes.

apps/mobile-wallet/src/persistent-storage/wallet.ts
nop33: I think the `generate` prefix is enough :)

mvaivre: In general, you can generate something with isn't the initial item. But ok, very nit!

apps/mobile-wallet/src/store/addresses/addressesStorageUtils.ts
mvaivre: Sounds good!

mvaivre (approved):
Okaaay let's go! 🎢

PR #570, opened 2024-05-16
Fix auto lock

### LINKED ISSUE #568: Replace custom idle hook with useIdle from useHooks or similar
We noticed in Posthog an irregular number of "Locked wallet" events with origin "Auto lock". After spending quite some time debugging it we couldn't find its source. To make sure it's not a bug on our side, let's use a library that provides this feature, like https://usehooks.com/useidle to replace our code:

https://github.com/alephium/alephium-frontend/blob/next/apps/desktop-wallet/src/hooks/useIdleForTooLong.ts

nop33:
Fixes:
- #568


mvaivre (approved):
Looks good! Got a little warning linked to `useAutoLock`, but I checked and it was already present when we were calling `useIdleForTooLong.ts`.

```
[Violation] 'setTimeout' handler took 445ms
```

PR #577, opened 2024-05-21
Move bip39 words to shared and add tests

mvaivre:


apps/desktop-wallet/src/pages/NewWallet/CreateWalletPage.tsx
```diff
@@ -93,7 +93,14 @@ const CreateWalletPage = ({ isRestoring = false }: { isRestoring?: boolean }) =>
     sendAnalytics({ event: 'Creating wallet: Creating password: Clicked next' })
 
     try {
-      saveNewWallet({ walletName, encrypted: await encryptMnemonic(mnemonic, password) })
+      const encrypted = await encryptMnemonic(
+        mnemonic,
+        password,
+        (e: unknown) => sendErrorAnalytics(e, 'Could not validate Uint8Array mnemonic', true),
+        (e: unknown) => sendErrorAnalytics(e, 'Could not validate string mnemonic', true)
+      )
```
nop33: hm, I had removed these

packages/shared/src/errors.ts
```diff
@@ -16,5 +16,16 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
+import { bip39Words as bip39WordsString } from '@/utils/bip39'
+
+const bip39Words = bip39WordsString.split(' ')
```
nop33: Let's directly replace `' '` with `'|'` since we don't need an array anymore, like I had in the code you removed 👍 

```ts
const bip39Words = bip39WordsString.replaceAll(' ', '|')
```

mvaivre: Already done yes!

nop33 (approved):

PR #579, opened 2024-05-21
Track more analytics in auto-update system

nop33:
A user reported an issue with the auto update, so I took the opportunity to add more analytics cause we haven't been tracking any and I don't know how to help them


mvaivre (approved):
LGTM 👏  Good initiative.

PR #580, opened 2024-05-22
Check if explorer is online

nop33:


mvaivre (approved):

packages/shared/src/errors.ts
```diff
@@ -21,7 +21,15 @@ import { bip39Words } from '@/bip39'
...
+    ? ` - ${error.message}`
+    : typeof error?.toString === 'function'
+      ? ` - ${error.toString().replace('Error: [API Error] - ', '')}`
+      : '')
 
 export const cleanExceptionMessage = (error: unknown) =>
   getHumanReadableError(error, '').replace(new RegExp(`\\b(${bip39WordsString})\\b`, 'g'), '[...]')
+
+const errorHasMessageProp = (error: unknown): error is { message: string } =>
+  (error as { message: string })?.message !== undefined
```
mvaivre, line 35: Nit: I used the `in` notation extensively in my tanstack branch and I like it, it's easily readable. 
Here we could do: 

```ts
const errorHasMessageProp = (error: unknown): error is { message: string } =>
  'message' in (error as { message?: string })
```

mvaivre (approved):

PR #582, opened 2024-05-22
Better TX info page

mvaivre:


nop33:
Also, what about having the amounts stacked below the address instead in the same row? Did you try out to see how it looks? I feel like it would make an easier to understand UI and you'd have less responsiveness issues

nop33:
I quite like it like this 
image

mvaivre:
Thanks for your feedback! I liked to be able to vertically parse the address list and amount list easily. See what I mean? It make it way easier to find what you are looking for when data of the same kind is properly aligned.

nop33 (commented):
Hm, ok, you know best!

btw, I still see responsiveness issues:

image

mvaivre:
@nop33 Are you sure you've pulled the last changes before testing? I can't reproduce.

PR #583, opened 2024-05-22
Final pass on analytics

nop33:


nop33:
I quite like it like this 
image

PR #599, opened 2024-06-02
Adapt Fund password flow

mvaivre:


nop33 (approved):
Good ideas!

1. I've removed a console log that was forgotten and pushed the commit.
2. I've created a new PR with issues that I found with this one. Here it is: https://github.com/alephium/alephium-frontend/pull/600
3. I'm approving this because I don't have further changes to request :+1:

PR #606, opened 2024-06-03
Show fungible tokens with broken metadata in the correct tab

mvaivre:
... instead of showing them in the "unknown" tab.
Also patched an issue when decimals are not provided when they should be. 

_Note: typing should be improved in the future following my work on the tanstack branch (more fields can be optional, we should cast less types etc.)_


nop33 (approved):
Code looks good. I launched the dev server, but I don't really have a link (address/tx) to test. Do you have any to recommend?

nop33:
Approved

PR #607, opened 2024-06-03
Use un-versioned node & explorer endpoints

### LINKED ISSUE #605: Use base non-versioned public endpoints in all apps
Backend services are now considered stable enough to ensure that breaking changes won't happen on a regular basis.
Let's now use default base endpoints https://docs.alephium.org/infrastructure/public-services#node-and-explorer-apis

nop33:
Closes #605


mvaivre (approved):
👍 thank you!

PR #667, opened 2024-06-14
Show avg. block time in seconds

mvaivre:


nop33 (approved):
Apart from the NPM audit error, the code looks good.

PR #669, opened 2024-06-14
Add missing i18n in Explorer

nop33:


apps/explorer/locales/en-US/translation.json
```diff
@@ -1,7 +1,8 @@
 {
   "ALPH balance": "ALPH balance",
   "Address group": "Address group",
-  "Address": "Address",
+  "Addresses_one": "Address",
+  "Addresses_other": "Addresses",
```
mvaivre, line 5: Why using this syntax over "Address" / "Addresses"?

nop33: Check the docs for i18next regarding plurals: https://www.i18next.com/translation-function/plurals

mvaivre: Okay, thanks for the link. Let's hope translators understand this.

mvaivre (approved):

apps/explorer/locales/en-US/translation.json
nop33: They don't need to if they are going through Crowdin.I think it displays it accordingly. Let's see

Issue #688, opened 2024-06-23
Fix NFTs spacing in bottom modals


nop33:
Several issues:

- Most right NFT is a bit cut:
![image](https://github.com/alephium/alephium-frontend/assets/1579899/9dd34376-5acb-4aaa-82fc-c6b1e69452fb)

- NFT attributes don't stack properly
| not expanded | expanded |
| ------ | ------ |
| <img width="300" alt="image" src="https://github.com/alephium/alephium-frontend/assets/1579899/293caac1-66cb-4299-a071-eecc73bc31cd"> | <img width="300" alt="image" src="https://github.com/alephium/alephium-frontend/assets/1579899/e48672a6-c53b-479b-a2b2-ffef4d2d7a16"> |

Not-square NFTs are not displayed properly

<img src="https://github.com/alephium/alephium-frontend/assets/1579899/ef1973f9-26e8-4353-b9b1-f3f999551b57" width="300" />

PR #689, opened 2024-06-23
Fix spacing of NFTs in bottom modals & authentication errors

nop33:
Related to:
- #688


apps/mobile-wallet/src/hooks/useBiometrics.ts
```diff
@@ -163,30 +164,34 @@ export const useBiometricsAuthGuard = () => {
...
+      if (isBiometricsAuthRequired) {
+        onPromptDisplayed && onPromptDisplayed()
+        await triggerBiometricsPrompt({ successCallback, failureCallback })
+      } else {
+        successCallback()
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [biometricsRequiredForAppAccess, biometricsRequiredForTransactions]
+  )
```
nop33, line 194: Simply wrapping in a `useCallback` to make sure that a new function is created when the Redux state changes.

apps/mobile-wallet/src/navigation/RootStackNavigation.tsx
```diff
@@ -187,30 +185,62 @@ const AppUnlockModal = () => {
...
 
-          initializeAppWithStoredWallet()
+            initializeAppWithStoredWallet()
+          } catch {
+            sendAnalytics({
+              type: 'error',
+              message: `Could not migrate deprecated mnemonic of length ${deprecatedWallet.mnemonic.split(' ').length}`
+            })
+
+            navigation.navigate('LoginWithPinScreen')
```
nop33, line 223: Here is where I am navigating to the pin screen if the migration fails

apps/mobile-wallet/src/screens/LandingScreen.tsx
nop33, line 1: The changes in this file were done because of a strange behavior of Canvas: If we are in the login screen and we put the app to the background, when bringing it back to the foreground the Canvas is not visible. By forcing it to rerender when the app state changes, the problem is fixed.

apps/mobile-wallet/src/store/appSlice.ts
nop33, line 1: Now that we are using an authentication modal instead of a screen, we don't need to restore since the previous screen is always rendered behind the modal.

apps/mobile-wallet/src/screens/LandingScreen.tsx
mvaivre: ok, weird, but thanks for the explanations.

mvaivre (commented):
Code LGTM. Testing on my device...

mvaivre (approved):

PR #702, opened 2024-06-29
Create auto update snackbar

### LINKED ISSUE #618: Update desktop wallet without asking user to click "Update"
As discussed offline, we want to implement a similar update strategy as Slack has: When there is a new update, it gets downloaded in the background and it gets installed next time the app launches again. This will help us keep our users' versions up to date.

Ideas:
1. We could display a progress bar as a snackbar message when the app is launched and a download starts automatically
2. We could display a second snackbar saying that the download was completed that the app will be update next time it is launched again.

### LINKED ISSUE #696: Update bar is flickering after canceling an update
Looks like a minor progress bar issue.

- Opened the wallet.
- Got the “update wallet” notification, selected ok.
- Hit cancel update.
- Selected update again.
- Now the download bar flickers.

nop33:
Closes #696
Closes #618

- https://github.com/alephium/alephium-frontend/issues/696
- #618


mvaivre:
Code looks good! 
A few comments on looks:

image

What about tweaking the design & content a bit so that it looks like:

image

_continuing.._

nop33:
image
image
image
image

.github/workflows/translations-sync.yml
```diff
@@ -23,7 +23,9 @@ jobs:
           config: 'crowdin.yml'
           upload_sources: true
           upload_translations: true
+          auto_approve_imported: true
           download_translations: true
+          skip_untranslated_strings: true
```
nop33, line 28: I've added these 2 to speed up our translation workflow. It's annoying that when we change the keys the translations become unapproved. I think it's ok to consider translations added through GitHub as approved.

nop33: Also, if we agree that translations should go only through Crowdin, there's no point in downloading untranslated strings, it's just polluting our JSON files and it's more of a mess when we want to change something

mvaivre (approved):
LGTM! 🎸

PR #682, opened 2024-06-19
Show uncle blocks

mvaivre:
To be released after web3 1.0.2 is out of rc.

PR #703, opened 2024-06-30
Show uncle blocks (rebased)

nop33:
This PR supersedes https://github.com/alephium/alephium-frontend/pull/682.

`ghostUncles` is not exposed via the latest version of web3, which [uses explorer backend 1.19.3](https://github.com/alephium/alephium-web3/blob/master/packages/web3/package.json#L37C34-L37C40). It exists in [explorer backend 2.0.0-rc.1](https://github.com/alephium/explorer-backend/blob/v2.0.0-rc0/app/src/main/resources/explorer-backend-openapi.json#L309). Also, it's not in `BlockEntryLite` but `BlockEntry`. Hence, we need to:
- [x] wait until explorer backend v2.0.0 is released
- [x] release an rc version of web3 and update it in this branch
- [ ] wait until a version of web3 is released that uses explorer backend v2.0.0
- [ ] update web3 frontend deps to the new web3 version


mvaivre (approved):
LGTM, thanks a lot!

PR #705, opened 2024-06-30
Improve cold start by preloading data

### LINKED ISSUE #353: Connecting mobile wallet to a dApp on cold start (without initialized data) is problematic
> Sometimes when using walletconnect, it pops into the wallet, and starts to show the addresses.. but it shows my primary address and hasn't filled in the token/nft info yet.. a second later the walletconnect session address chooser goes away and i'm back in just the wallet.
(it's happened maybe 2-3 times in the past week)

nop33:
Closes #353

This will also be useful when we introduce multiple wallets.

- #353

### DIFF
--- a/.changeset/grumpy-weeks-repeat.md
+++ b/.changeset/grumpy-weeks-repeat.md
@@ -1,0 +1,5 @@
+---
+"@alephium/mobile-wallet": patch
+---
+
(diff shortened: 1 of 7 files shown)


mvaivre (approved):
Looks good and worked well on my device 🤙

Issue #658, opened 2024-06-13
Fix display of SVG data URL NFTs in mobile wallet


nop33:
Duplicate of https://github.com/alephium/alephium-frontend/issues/589

nop33:
So even though both desktop wallet and mobile wallet use the same code for fetching NFT `tokenUri`, displaying SVGs works on desktop wallet but not on mobile. This is the code that gets the JSON from the `tokenUri`: https://github.com/alephium/alephium-frontend/blob/next/packages/shared/src/store/assets/assetsActions.ts#L130-L136

```ts
  const promiseResults = await Promise.allSettled(
    nftsMetadata.map(({ tokenUri, id }) =>
      exponentialBackoffFetchRetry(tokenUri)
        .then((res) => res.json())
        .then((data) => ({ ...data, id }))
    )
  )
```

In the React Native environment we get `TypeError: Network request failed]` when the `tokenUrl` is something like

```
[...]

Issue #28, opened 2023-10-04
Improve loading of NFTs display


nop33:
- [ ] We could show some loading animation while the images are loading
- [x] Make sure there are no empty spaces. @h0ngcha0 reported:
<details><summary>Sometimes there are some empty spots on the NFT tab</summary>

![image](https://github.com/alephium/mobile-wallet/assets/1579899/8f43d7c6-ae09-4dc2-87d2-753e9e11cc4c)

</details>

mvaivre:
@nop33 we may close that soon?

nop33:
The first subtask is not yet completed.

nop33:
As [discussed on Slack](https://alephium.slack.com/archives/C01B3D7HXAT/p1719213640751289), as of now, in the mobile and desktop wallet as well as the explorer we pass the [`NFTTokenUriMetaData.image`](https://github.com/alephium/alephium-web3/blob/cfe659e94d10bb20497d60a72c2c6f17f9e5f474/packages/web3/src/api/types.ts#L259-L267) string directly to the image component (`<img src={nft.image} />` for desktop wallet & explorer or `<Image source={{ uri: nft.image }} />` for mobile wallet). 

The image can potentially be a very large file. This gets especially alarming when we display a list of NFTs, where many can be a huge size, all displayed in the same screen at the same time.

The approach that Deadrare is using is the following:

They fetch all images through their image resizing service, which has has a CDN (Google Cloud CDN) infront of it for fast cached image retrieval.

Their image resizing service is available at https://image.deadrare.io/. It is a simple Node.JS server of 92 LoC that uses Sharp: https://sharp.pixelplumbing.com/

This is an example of how they would load an NFT that has the following as its image property:
- https://arweave.net/pur_d_ASLCJUj8-2Gfj7DG1fjr1C-c31iE9oGlaovqQ/14.png
- https://image.deadrare.io/?source=https://arweave.net/pur_d_ASLCJUj8-2Gfj7DG1fjr1C-c31iE9oGlaovqQ/14.png&width=800&height=800&fit=cover&position=centre&upscale=true&pixelArt=false&w=3840&q=75

A recommendation is that:
> “You don’t want to be serving the default image size in a list view, they can be massive.”

From our perspective, we don't want to use or maintain such a backend service because we want the wallets to not depend on external services such as the explorer backend.

[...]

PR #707, opened 2024-07-01
Improve NFT loading

nop33:
This PR addresses these issues:
- https://github.com/alephium/alephium-frontend/issues/658
- https://github.com/alephium/alephium-frontend/issues/28
- https://github.com/alephium/alephium-frontend/issues/425

### DIFF
--- a/apps/mobile-wallet/src/components/NFTImage.tsx
+++ b/apps/mobile-wallet/src/components/NFTImage.tsx
@@ -18,9 +18,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import { NFT, selectNFTById } from '@alephium/shared'
 import { Image } from 'expo-image'
-import { Skeleton } from 'moti/skeleton'
 import { memo, useState } from 'react'
-import { View } from 'react-native'
 import styled, { useTheme } from 'styled-components/native'
 
 import NFTPlaceholder from '~/components/NFTPlaceholder'
@@ -37,7 +35,6 @@ const NFTImage = ({ nftId, size }: NFTImageProps) => {
   const nft = useAppSelector((s) => selectNFTById(s, nftId))
   const theme = useTheme()
 
-  const [isLoading, setIsLoading] = useState(false)
   const [hasError, setHasError] = useState(false)
 
   if (!nft) return null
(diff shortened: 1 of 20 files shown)


apps/mobile-wallet/src/components/NFTThumbnail.tsx
```diff
@@ -92,6 +134,37 @@ const NFTThumbnail = ({ nft, size }: NFTThumbnailProps) => {
...
+)
+
+const WebViewImage = ({ nft, size }: NFTThumbnailProps) => (
+  <WebView
+    source={{ html: `<img src="${nft.image}" />` }}
+    style={{ width: size, height: size, borderRadius: BORDER_RADIUS_SMALL, backgroundColor: 'transparent' }}
+    injectedJavaScript={
+      "const meta = document.createElement('meta'); meta.setAttribute('content', 'width=width, initial-scale=1, maximum-scale=1, user-scalable=1'); meta.setAttribute('name', 'viewport'); document.getElementsByTagName('head')[0].appendChild(meta); "
+    }
+    javaScriptEnabled={false}
```
mvaivre, line 150: So this should avoid any injection attack, correct?
And can you confirm that all that can be rendered in this webview is an  tag?

apps/mobile-wallet/src/components/NFTThumbnail.tsx
```diff
@@ -82,6 +118,12 @@ const NFTThumbnail = ({ nft, size }: NFTThumbnailProps) => {
                   </AttributesGrid>
                 )}
               </ScreenSection>
+
+              {!isDataUri && (
+                <ScreenSection>
+                  <Button title={t('View full size')} onPress={() => openBrowserAsync(nft.image)} />
```
mvaivre, line 124: Q: what will happen if the `image` filed is a link to a malicious / fishing website? I have the same problem on the explorer. We can't do much about it, can we?

mvaivre:
Two things I noticed:

1) glitches when loading the NFT list, as if some images are available in cache, but reloaded anyway.

https://github.com/alephium/alephium-frontend/assets/6715943/cf38f300-09f5-4412-8dd0-ed2ea7275f5b

2) Nit: if I keep the list open, the thumbnails re-render every minute or so.

apps/mobile-wallet/src/components/NFTThumbnail.tsx
nop33: The only user content that we are rendering is the `nft.image`, which I am sanitizing. So, I am fairly confident that we are safe. I don't see how anything could be injected in the webview

apps/mobile-wallet/src/components/NFTThumbnail.tsx
nop33: I don't think we can, but I am open to ideas.

nop33:
@mvaivre The things you mentioned already existed before I believe. Tanstack will probably be useful here.

apps/mobile-wallet/src/components/NFTThumbnail.tsx
mvaivre: I don't have any, except for preventing from opening the full size image..

apps/mobile-wallet/src/components/NFTImage.tsx
```diff
@@ -0,0 +1,68 @@
...
+  ) : (
+    <NFTImageStyled
+      style={{ width: size, height: size }}
+      transition={500}
+      source={{ uri: nft.image }}
+      allowDownscaling
+      contentFit="contain"
+      onError={() => setHasError(true)}
+      placeholder={{
+        blurhash: theme.name === 'dark' ? 'L00000fQfQfQfQfQfQfQfQfQfQfQ' : 'L1PGpx-;fQ-;_3fQfQfQfQfQfQfQ'
```
mvaivre, line 57: wut?

mvaivre: (that would be so cool to be able to use this :))

nop33: yes, I love the idea of blurhash. Many of the NFT indexers provide it. Unfortunately, we can't use one :(

mvaivre (approved):
🎨

PR #710, opened 2024-07-02
Enable languages

nop33:


apps/mobile-wallet/src/features/localization/useLocalization.ts
```diff
@@ -18,6 +18,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import 'dayjs/locale/fr'
 import 'dayjs/locale/el'
+import 'dayjs/locale/vi'
```
mvaivre, line 21: Should Greek and German also be here?

apps/mobile-wallet/src/features/localization/i18n.ts
```diff
@@ -22,13 +22,15 @@ import { initReactI18next } from 'react-i18next'
...
+import vi from '../../../locales/vi-VN/translation.json'
 
 i18next.use(initReactI18next).init({
   compatibilityJSON: 'v3',
   resources: {
     'en-US': { translation: en },
     'fr-FR': { translation: fr },
-    'el-GR': { translation: gr }
+    'el-GR': { translation: gr },
+    'vi-VN': { translation: vi }
```
mvaivre, line 33: Should German also be here?

apps/mobile-wallet/src/features/localization/languages.ts
```diff
@@ -16,10 +16,11 @@ You should have received a copy of the GNU Lesser General Public License
...
 
-export type Language = 'en-US' | 'fr-FR' | 'el-GR'
+export type Language = 'en-US' | 'fr-FR' | 'el-GR' | 'vi-VN'
 
 export const languageOptions: { label: string; value: Language }[] = [
   { label: 'English', value: 'en-US' },
   { label: 'Français', value: 'fr-FR' },
-  { label: 'Ελληνικά', value: 'el-GR' }
+  { label: 'Ελληνικά', value: 'el-GR' },
+  { label: 'Tiếng Việt', value: 'vi-VN' }
```
mvaivre, line 25: Should German also be here?

apps/mobile-wallet/src/features/localization/useLocalization.ts
nop33: Greek is there, the second row (`el`). We don't have full German translations yet, it's very incomplete, so I'd rather wait until we have full coverage before enabling it.

apps/mobile-wallet/src/features/localization/i18n.ts
nop33: Same answer as before

apps/mobile-wallet/src/features/localization/languages.ts
nop33: aaaand again :)

mvaivre (approved):
Alrighty

PR #715, opened 2024-07-04
Add bg color to NFT items

mvaivre:


mvaivre:
Changed my mind. Let's go without it.

PR #736, opened 2024-07-15
Fetch all token prices with Tanstack

nop33:


apps/desktop-wallet/src/App.tsx
```diff
@@ -88,6 +86,7 @@ const App = () => {
   const showDevIndication = useDevModeShortcut()
   const posthog = usePostHog()
   const { sendAnalytics } = useAnalytics()
+  useAlphPrice()
```
nop33, line 89: Prefetching ALPH price to re-create the previous behavior

apps/desktop-wallet/src/App.tsx
```diff
@@ -266,34 +264,6 @@ const App = () => {
...
-      })
-    )
-  }, [dispatch, settings.fiatCurrency, verifiedFungibleTokenSymbols.withPriceHistory])
-
-  useInterval(
-    refreshTokensLatestPrice,
-    PRICES_REFRESH_INTERVAL,
-    networkStatus !== 'online' || verifiedFungibleTokenSymbols.withPriceHistory.length === 0
-  )
-
```
nop33, line 296: Goodbye difficult to reason about code ✨

apps/desktop-wallet/src/App.tsx
```diff
@@ -253,7 +252,6 @@ const App = () => {
       } else if (verifiedFungibleTokenSymbols.uninitialized.length > 0) {
         const symbols = verifiedFungibleTokenSymbols.uninitialized
 
-        dispatch(syncTokenCurrentPrices({ verifiedFungibleTokenSymbols: symbols, currency: settings.fiatCurrency }))
```
nop33, line 256: We no longer need to fetch prices with Redux.

apps/desktop-wallet/src/App.tsx
```diff
@@ -253,7 +252,6 @@ const App = () => {
...
         dispatch(syncTokenPriceHistories({ verifiedFungibleTokenSymbols: symbols, currency: settings.fiatCurrency }))
```
nop33, line 255: But we still need to get historic prices for the chart. Once that is removed, this whole useEffect will be deleted.

apps/desktop-wallet/src/components/Inputs/SelectOptionAddress.tsx
```diff
@@ -39,7 +40,7 @@ const SelectOptionAddress = ({ address, isSelected, className }: SelectOptionAdd
   const selectAddressesTokens = useMemo(makeSelectAddressesTokens, [])
   const assets = useAppSelector((s) => selectAddressesTokens(s, address.hash))
 
-  const knownAssetsWithBalance = assets.filter((a) => a.balance > 0 && a.name)
+  const knownAssetsWithBalance = useSortTokensByWorth(assets.filter((a) => a.balance > 0 && a.name))
```
nop33, line 43: See explanation below why I introduced this new hook

apps/desktop-wallet/src/features/tokenPrices/hooks.tsx
```diff
@@ -0,0 +1,99 @@
...
+    [
+      (a) => (a.verified ? 0 : 1),
+      (a) => a.worth ?? -1,
+      (a) => a.verified === undefined,
+      (a) => a.name?.toLowerCase(),
+      'id'
+    ],
+    ['asc', 'desc', 'asc', 'asc', 'asc']
+  )
+}
```
nop33, line 85: Since we don't want to rely on the Redux priceSlice, we need to implement the sorting outside of the Redux selectors. This simple hook gets the list of tokens (which is still supplied by Redux) and sorts it based on the worth (prices being fetched by Tanstack)

apps/desktop-wallet/src/features/tokenPrices/hooks.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const tokens = useAppSelector((s) => selectAddressesKnownFungibleTokens(s, addressHashes))
+
+  return tokenPrices.reduce((totalWorth, { symbol, price }) => {
+    const verifiedFungibleToken = tokens.find((t) => t.symbol === symbol)
+
+    return verifiedFungibleToken
+      ? totalWorth + calculateAmountWorth(verifiedFungibleToken.balance, price, verifiedFungibleToken.decimals)
+      : totalWorth
+  }, 0)
+}
```
nop33, line 99: For the same reason as above, we need to remove the dependency on the priceSlice. This hook allows us to calculate the total worth of the address tokens without relying on priceSlice.

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressGridRow.tsx
```diff
@@ -54,13 +54,12 @@ const AddressGridRow = ({ addressHash, className }: AddressGridRowProps) => {
   const stateUninitialized = useAppSelector(selectIsStateUninitialized)
   const verifiedFungibleTokensNeedInitialization = useAppSelector(selectDoVerifiedFungibleTokensNeedInitialization)
   const fiatCurrency = useAppSelector((s) => s.settings.fiatCurrency)
-  const areTokenPricesInitialized = useAppSelector((s) => s.tokenPrices.status === 'initialized')
-  const selectAddessesTokensWorth = useMemo(makeSelectAddressesTokensWorth, [])
-  const balanceInFiat = useAppSelector((s) => selectAddessesTokensWorth(s, addressHash))
+  const { isPending: isPendingTokenPrices } = useAddressTokensPrices()
```
nop33, line 57: `isPending` is a super cool feature :D

apps/desktop-wallet/src/pages/UnlockedWallet/TransfersPage/FiltersPanel.tsx
```diff
@@ -59,7 +60,8 @@ const FiltersPanel = ({
   const { t } = useTranslation()
   const addresses = useAppSelector(selectAllAddresses)
   const selectAddressesTokens = useMemo(makeSelectAddressesTokens, [])
-  const assets = useAppSelector(selectAddressesTokens)
+  const addressTokens = useAppSelector(selectAddressesTokens)
+  const assets = useSortTokensByWorth(addressTokens)
```
nop33, line 64: Soon enough we'll rename the `assets` everywhere, but for now I keep it to minimize the LOC diff

apps/desktop-wallet/src/storage/addresses/addressesSelectors.ts
```diff
@@ -85,15 +82,34 @@ export const makeSelectAddressesAlphAsset = () =>
 
 export const makeSelectAddressesTokens = () =>
   createSelector(
-    [selectAllFungibleTokens, selectAllNFTs, makeSelectAddressesAlphAsset(), makeSelectAddresses(), selectAllPrices],
-    (fungibleTokens, nfts, alphAsset, addresses, tokenPrices): Asset[] => {
-      const tokenBalances = getAddressesTokenBalances(addresses)
-      const tokens = calculateAssetsData([alphAsset, ...tokenBalances], fungibleTokens, nfts, tokenPrices)
-
-      return sortAssets(tokens)
```
nop33, line 93: This is where I remove the sorting (which comes from the shared package, which relies on the priceSlice.

apps/desktop-wallet/src/features/tokenPrices/hooks.tsx
```diff
@@ -0,0 +1,99 @@
...
+  )
+}
+
+export const useAddressTokensWorth = (addressHashes?: AddressHash[] | AddressHash) => {
+  const { data: tokenPrices } = useAddressTokensPrices()
+  const selectAddressesKnownFungibleTokens = useMemo(makeSelectAddressesKnownFungibleTokens, [])
+  const tokens = useAppSelector((s) => selectAddressesKnownFungibleTokens(s, addressHashes))
+
+  return tokenPrices.reduce((totalWorth, { symbol, price }) => {
+    const verifiedFungibleToken = tokens.find((t) => t.symbol === symbol)
```
mvaivre, line 93: I really like the use of type guards in my branch, which really helped make things clearer and stricter.
Here for instance, you could use a type guard like `isListedToken(t)`.

Also, let's stop using "verified" and use "listed" instead, WDYT?

apps/desktop-wallet/src/features/tokenPrices/hooks.tsx
```diff
@@ -0,0 +1,99 @@
...
+      refetchInterval: ONE_MINUTE_MS
+    })),
+    combine: (results) => ({
+      data: results.flatMap(({ data }) => data).filter((price) => !!price),
+      isPending: results.some(({ isPending }) => isPending)
+    })
+  })
+
+  return { data, isPending }
+}
```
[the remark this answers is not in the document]
nop33: Hehe, I was having a hard time deciding as well. I guess since this triggers something for you, the best would be to stick with the pattern that we have created: If it handles an array, then use plural, if not, singular. I'll update.

apps/desktop-wallet/src/App.tsx
mvaivre: Mhh. I'm not sure I like it. Does it really bring value over fetching the price in the component that is first mounted and which requires the price?

apps/desktop-wallet/src/features/tokenPrices/hooks.tsx
nop33: Updated!

apps/desktop-wallet/src/features/tokenPrices/hooks.tsx
```diff
@@ -0,0 +1,97 @@
+/*
```
mvaivre, line 1: Should the file be called `tokenPricesHooks.tsx` to help with navigating through the project? 🤔

apps/desktop-wallet/src/storage/addresses/addressesSelectors.ts
```diff
@@ -85,15 +82,34 @@ export const makeSelectAddressesAlphAsset = () =>
...
+    acc.push({
+      id: token.id,
+      balance,
+      lockedBalance: BigInt(token.lockedBalance.toString()),
+      name: fungibleToken?.name ?? nftInfo?.name,
+      symbol: fungibleToken?.symbol,
+      description: fungibleToken?.description ?? nftInfo?.description,
+      logoURI: fungibleToken?.logoURI ?? nftInfo?.image,
+      decimals,
+      verified: fungibleToken?.verified
```
mvaivre, line 107: Linked to our discussion: the type guards allow us to get rid of this additional field. That's cool, because we can use the unaltered data types we get from the backend.

apps/desktop-wallet/src/features/tokenPrices/hooks.tsx
nop33: Sure!

apps/desktop-wallet/src/storage/addresses/addressesSelectors.ts
nop33: awesome

mvaivre (approved):
LGTM! GJ!

PR #737, opened 2024-07-16
Fetch NFT collection data and ALPH historical prices with Tanstack

nop33:
Introduce Axios for data fetching.


apps/desktop-wallet/src/modals/NFTDetailsModal.tsx
```diff
@@ -73,17 +72,53 @@ const NFTDetailsModal = ({ nftId, onClose }: NFTDetailsModalProps) => {
...
+          if (matchesNFTCollectionUriMetaDataSchema(data)) {
+            return data as NFTCollectionUriMetaData
+          } else {
+            throw new Error(
+              `Response does not match the NFT collection metadata schema. NFT collection URI: ${collectionUri}`
+            )
+          }
+        },
+    staleTime: Infinity
+  })
```
nop33, line 110: You will probably say that you'd prefer to see these queries in a separate file but hear me out! This is the only place that they are needed currently. Eventually, they will probably be moved to the `@alephium/shared` package. But for now, IMO, it makes sense to keep them here, where they are needed. WDYT?

apps/desktop-wallet/src/modals/NFTDetailsModal.tsx
```diff
@@ -96,3 +131,8 @@ const NFTMetadataContainer = styled.div`
   flex-direction: column;
   gap: var(--spacing-3);
 `
+
+const matchesNFTCollectionUriMetaDataSchema = (nftCollection: NFTCollectionUriMetaData) =>
+  typeof nftCollection.name === 'string' &&
+  typeof nftCollection.image === 'string' &&
+  typeof nftCollection.description === 'string'
```
nop33, line 138: Same as above. It will be moved to shared when we extend the Tanstack integration.

apps/desktop-wallet/src/modals/NFTDetailsModal.tsx
mvaivre: Haha :) Here's why I like the idea of a "query factory" after having using it for a while:

- If we make sure to always add a new query there, it's basically a list of endpoints we can use to build new things. 
- If we only use queries from this factory, we're sure the the Tanstack cache is used in an optimal way. No risk to change a key name by mistake.
- It make the components way lighter and more readable.

Those are just the first arguments from the top of my head :)

mvaivre: Also, I think it's a great exercise to build such a factory.
However, I'm fine with you starting without it in your own Tanstack journey! Let's see if you'll eventually feel the need.

nop33: hehe, even though I hate the word "factory" in the context of programming, I agree with you that we should eventually extract this code elsewhere. Do you think it needs to be done now? I would prefer to do it once mobile wallet also integrates Tanstack and/or when we agree upon a fetching architecture among our apps (currently explorer uses batshit so it's a bit hard to share code.

But overall I agree with all 3 points ✅

mvaivre: As said above, let's see how this goes :)

PS: I also hate the word factory. I used "queryCollection" in my previous work hehe :)

apps/desktop-wallet/src/features/tokenPrices/tokenPricesHooks.tsx
```diff
@@ -59,6 +69,41 @@ export const useAlphPrice = () => {
   return tokenPrices.find((token) => token.symbol === ALPH.symbol)?.price
 }
 
+export const useAlphHistoricalPrices = () => {
+  const currency = useAppSelector((s) => s.settings.fiatCurrency).toLowerCase()
+
+  const { data: alphPriceHistory } = useQuery({
+    queryKey: ['tokensPrice', 'historicalPrice', ALPH.symbol, currency],
```
mvaivre, line 76: I notice that you use the `tokensPrice` key here, and you used `tokenPrices` in the query above. Is this intentional? 

Using the same base key could eventually be useful to leverage `useIsFetching` ([docs](https://tanstack.com/query/latest/docs/framework/react/reference/useIsFetching))

nop33: It was not intentional. Maybe this is a good reason to have a way to not make this mistake. Maybe something like key generator? :/

Also, pretty cool feature the one you shared

nop33: pushed an update, lemme know what you think

nop33: very good catch btw

mvaivre: That's the kind of mistake I was talking about when talking about query collections. The more you duplicate / extend a query, the more this can happen. 
We could take another look at TkDodo's recommended approach to "query factories" https://tkdodo.eu/blog/the-query-options-api#query-factories

I personally find it a bit confusing to mix key-only values with functions, so I decided to use function only. But yeah, let's find the best solution :)

mvaivre: About the "key generator" idea, I'm aligned with TkDodo:

image

nop33: I noticed that in your query collections you don't reuse query keys. The idea that queries of the same collection should share the same keys, going from the most general to the most specific (iiuc).

I agree that we should not separate the query key from the query function and I also don't like TkDodo's (old) approach on key factories: https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories

I extracted the main query key. WDYT?

mvaivre: My guess is that combining a simple map of query keys (similar to what you did) close to the query collection (what I did) could be a good solution.
But let's see further down the road!

mvaivre (approved):
LGTM! Great work.

PR #740, opened 2024-07-16
Fetch all historic chart related data with Tanstack

nop33:


apps/desktop-wallet/src/App.tsx
```diff
@@ -207,12 +207,6 @@ const App = () => {
           } catch {
             sendAnalytics({ type: 'error', message: 'Could not sync address data automatically' })
           }
-
-          try {
-            dispatch(syncAddressesAlphHistoricBalances())
-          } catch {
-            sendAnalytics({ type: 'error', message: 'Could not sync alph historic balances automatically' })
-          }
```
nop33, line 215: hell yeah

apps/desktop-wallet/src/features/historicChart/HistoricWorthChart.tsx
```diff
@@ -146,110 +142,6 @@ const HistoricWorthChart = memo(function HistoricWorthChart({
 
 export default HistoricWorthChart
 
-const getFilteredChartData = (chartData: DataPoint[], startingDate: string) => {
-  const startingPoint = chartData.findIndex((point) => point.date === startingDate)
-  return startingPoint > 0 ? chartData.slice(startingPoint) : chartData
-}
-
-const getChartOptions = (
```
nop33, line 154: Moving these 2 functions to their own file

apps/desktop-wallet/src/features/historicChart/useHistoricData.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+type Timestamp = string
+type Amount = string
+
+const useHistoricData = () => {
+  const allAddressHashes = useAppSelector((s) => s.addresses.ids as AddressHash[])
+  const currency = useAppSelector((s) => s.settings.fiatCurrency).toLowerCase()
+
+  const { data: alphPriceHistory, isPending: isPendingAlphPriceHistory } = useQuery({
+    queryKey: [HISTORY_QUERY_KEY, 'price', ALPH.symbol, { currency }],
```
nop33, line 38: This key (and the next one) look like this:

image

I am not sure yet what should go in an object  and what should go plain into the keys array. In the docs I see a lot `{ filter }` being used as an object. I guess it's not very important but I am curious to know pros/cons of the 2 approaches. Just food for thought.

apps/desktop-wallet/src/features/historicChart/useHistoricData.tsx
```diff
@@ -0,0 +1,127 @@
...
+              history.push({
+                date: itemDate,
+                value: price
+              })
+            }
+          }
+        }
+
+        return history
+      })
```
nop33, line 63: This was mainly copied over from the current implementation, no new code

apps/desktop-wallet/src/features/historicChart/useHistoricData.tsx
```diff
@@ -0,0 +1,127 @@
...
+                )
+          }
+
+          return historyPerAddress
+        },
+        {} as Record<AddressHash, Record<Timestamp, Amount> | undefined>
+      ),
+      isPending: results.some(({ isPending }) => isPending),
+      hasHistoricBalances: results.some(({ data }) => data?.amountHistory?.length)
+    })
```
nop33, line 114: Instead of combining everything to arrays, I am combining them to objects. This way, we can access the data faster when we need them in the chart component (see later comment) instead of doing `.find`, which is more expensive. Reduce and map have the same complexity, so here it doesn't cost us anything to return objects instead of arrays.

apps/desktop-wallet/src/storage/addresses/addressesActions.ts
```diff
@@ -156,58 +152,6 @@ export const syncAllAddressesTransactionsNextPage = createAsyncThunk(
   }
 )
 
-export const syncAddressesAlphHistoricBalances = createAsyncThunk(
```
nop33, line 159: good riddance

apps/desktop-wallet/src/storage/addresses/addressesStorageUtils.ts
```diff
@@ -50,7 +49,6 @@ export const saveNewAddresses = (addresses: AddressBase[]) => {
 
   store.dispatch(newAddressesSaved(addresses))
   store.dispatch(syncAddressesData(addressHashes))
-  store.dispatch(syncAddressesAlphHistoricBalances(addressHashes))
```
nop33, line 53: I'm so happy that we can simply remove this line and the component will take care of making a new request when it needs to, imperatively :D

apps/desktop-wallet/src/types/addresses.ts
```diff
@@ -51,8 +45,6 @@ export type Address = AddressBase &
     allTransactionPagesLoaded: boolean
     tokens: AddressTokenBalance[]
     lastUsed: TimeInMs
-    alphBalanceHistory: EntityState<BalanceHistory>
-    alphBalanceHistoryInitialized: boolean
```
nop33, line 55: byebye Redux state

apps/desktop-wallet/src/features/tokenPrices/tokenPricesHooks.tsx
```diff
@@ -39,21 +29,32 @@ import {
...
-        })),
+        (await client.explorer.market.postMarketPrices({ currency }, symbols)).map(
+          (price, i) =>
+            ({
+              price,
+              symbol: symbols[i]
+            }) as TokenPrice
+        ),
+      initialData: () =>
+        queryClient.getQueryData([TOKEN_PRICES_KEY, 'currentPrice', [ALPH.symbol], { currency }]) as TokenPrice[],
```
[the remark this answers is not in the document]
mvaivre: Very nice!

apps/desktop-wallet/src/storage/addresses/addressesStorageUtils.ts
mvaivre: Great feeling eh!

apps/desktop-wallet/src/features/historicChart/useHistoricData.tsx
mvaivre: I think that the main point here is [this](https://tanstack.com/query/latest/docs/framework/react/guides/query-keys#query-keys-are-hashed-deterministically).

mvaivre (approved):
👏

PR #745, opened 2024-07-18
Fetch token balances with Tanstack

nop33:
Since our philosophy is to "start from the component and give it what it needs", I created hooks that give the components exactly that. I grouped the hooks based on the data they return. Hooks that return balances (alph balance, token balance, locked balance) are all grouped together. Same for hooks that return tokens info. Or hooks that return price info. I think this makes more sense for the for the devX point of view when writing components. I propose we both experiment coding with this convention and see how it goes. See [these](https://github.com/alephium/alephium-frontend/pull/745#discussion_r1685761390) [comments](https://github.com/alephium/alephium-frontend/pull/745#discussion_r1685761992) for more info.


apps/desktop-wallet/src/api/addressesFungibleTokensInfoDataHooks.tsx
```diff
@@ -0,0 +1,67 @@
...
+    })
+  })
+
+  return {
+    data,
+    isLoading
+  }
+}
+
+export const useAddressesListedFungibleTokensWithPrice = (addressHash?: AddressHash) => {
```
mvaivre, line 63: General comment: I'm not sure to understand how the hooks are split between those files. 
I'd rather have less specific files with more hooks in them, to get a better view of the API structure and of the data I can get. 

To take this one as exemple, it's not clear to me if this should be in here or in `addressesFungibleTokensPricesDataHooks.tsx`

Idea: we could maybe group hooks according to the type of data, similar to how the api is designed?..
`api/tokens/ft`
`api/tokens/nft`
`api/addresses/balances`
etc...

apps/desktop-wallet/src/features/balancesOverview/TotalAlphBalance.tsx
```diff
@@ -44,19 +44,19 @@ const TotalAlphBalance = ({ className, type }: TotalAlphBalanceProps) => {
...
-  const { totalBalance, totalLockedBalance } = useTotalAlphBalances()
+  const { data } = useAddressesAlphBalances()
 
-  return <AlphBalance balance={totalBalance - totalLockedBalance} />
+  return <AlphBalance balance={data.balance - data.lockedBalance} />
 }
 
 const LockedAlphAmount = () => {
-  const { totalLockedBalance } = useTotalAlphBalances()
+  const { data } = useAddressesAlphBalances()
```
mvaivre, line 53: Nit: I prefer to name the output variable here, so that it's easy to understand what is feeding the component.

apps/desktop-wallet/src/api/addressesFungibleTokensInfoDataHooks.tsx
nop33: still experimenting with structures. I wanted to write a message to you to present the current structure but I didn't have the time I needed last week and I put it in draft. I wanted to say something in line of "I wanna introduce this file called [...]DataHooks.tsx for hooks that mainly return `data` objects from Tanstack queries (see https://tanstack.com/query/latest/docs/framework/react/guides/render-optimizations#structural-sharing).

> To take this one as exemple, it's not clear to me if this should be in here or in addressesFungibleTokensPricesDataHooks.tsx

In this case, `addressesFungibleTokensPricesDataHooks.tsx` includes hooks that return the same thing. Prices. Hence the `[...]PricesDataHooks.tsx` suffix.

The `addressesFungibleTokensInfoDataHooks.tsx` includes hooks that return TokenInfo objects. Hence the `[...]TokensInfoDataHooks.tsx` suffix.

We can talk about it more tomorrow, I'll rework the PR and remove the draft status.

apps/desktop-wallet/src/features/balancesOverview/TotalAlphBalance.tsx
nop33: I thought about that too and that's how I had originally named them. But I changed my mind after spending some time on inspecting the rerenders. Leaving it as `data` makes it clearer IMO that this object is structurally shared and that it comes directly from Tanstack. Let's talk about it more tomorrow.

apps/desktop-wallet/src/api/addressQueries.ts
```diff
@@ -28,9 +29,32 @@ interface AddressBalanceQueryProps {
...
   queryOptions({
-    queryKey: ['address', 'balance', { addressHash, latestTxHash, networkId }],
+    queryKey: ['address', 'balance', 'ALPH', { addressHash, latestTxHash, networkId }],
     queryFn: () => client.explorer.addresses.getAddressesAddressBalance(addressHash),
     staleTime: Infinity
   })
+
+export const addressTokensBalanceQuery = ({ addressHash, networkId, latestTxHash }: AddressBalanceQueryProps) =>
+  queryOptions({
+    queryKey: ['address', 'balance', 'tokens', { addressHash, latestTxHash, networkId }],
```
nop33, line 41: A few comments about using latestTxHash in the query key:
1. It is a very clean and straightforward way to refetch the query when the last tx hash changes ✅ When the last tx hash changes, a NEW query is being created and the previous won't be used anymore.
2. Setting `staleTime: Infinity` is necessary so that we don't fetch unnecessarily if there is no new tx hash. This minimizes the calls to explorer backend ✅. However, when there will be a new latest tx hash, the previous query will never get garbage collected 🤔.
3. When the tx hash changes, the new query will not prefill its catch with the data from the latest query unless we initialize it with `initialData`. We could keep track of the previous from last TX hash so that we can do that.

I will explore some alternative ways of achieving our goals, as we discussed a few weeks ago.

apps/desktop-wallet/src/api/addressQueries.ts
```diff
@@ -16,21 +16,70 @@ You should have received a copy of the GNU Lesser General Public License
...
   queryOptions({
-    queryKey: ['address', 'balance', { addressHash, latestTxHash, networkId }],
+    queryKey: [...ADDRESS_BALANCE_QUERY_KEYS, 'ALPH', { addressHash, latestTxHash, networkId }],
     queryFn: () => client.explorer.addresses.getAddressesAddressBalance(addressHash),
+    placeholderData: queryClient.getQueryData([
+      ...ADDRESS_BALANCE_QUERY_KEYS,
+      'ALPH',
+      { addressHash, latestTxHash: previousTxHash, networkId }
+    ]) as AddressBalance,
+    staleTime: Infinity
```
nop33, line 51: Adding `placeholderData` ensures that the new query (with the new lataestTxHash in its queryKey) will be prefilled with the data of the previous one in order to avoid component re-render with `undefined` or `0` value.

If we don't do that, we'd need to remove the latest tx hash from the query keys and we'd need to find a way to invalidate these queries. I quite like the simplicity of the approach of including the latest tx hash in the query keys. And the method of adding placeholder data works well!

apps/desktop-wallet/src/api/fungibleTokenListDataHooks.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const { data, isLoading } = useQuery({
+    queryKey: [TOKEN_LIST_QUERY_KEY, network],
+    queryFn: !network
+      ? skipToken
+      : () =>
+          axios
+            .get(`https://raw.githubusercontent.com/alephium/token-list/master/tokens/${network}.json`)
+            .then(({ data }) => (data as TokenList)?.tokens),
+    initialData: network === 'mainnet' ? mainnet.tokens : network === 'testnet' ? testnet.tokens : [],
```
nop33, line 40: Might need to be replaced with `placeholderData`. To be checked.

apps/desktop-wallet/src/api/addressQueries.ts
mvaivre: I like It too!

apps/desktop-wallet/src/features/balancesOverview/TotalAlphBalance.tsx
mvaivre: What about components using multiple query hooks, hence having multiple `data` objects?
But I get you point. We could use a "Data" suffix in the variable names? Like `alphBalancesData`.

nop33: This is as close as to what the `useQuery`/`useQueries` function returns. That's why I like this pattern.

mvaivre: Ah but that's exactly what I was saying 😅 
I also like it when I see `{ data: ... }` and I was suggesting that we use 

```ts
const { data: alphBalances } = useAddressesAlphBalances()
```

nop33: awesome, yes! 🔥

apps/desktop-wallet/src/api/addressesFungibleTokensInfoDataHooks.tsx
```diff
@@ -0,0 +1,67 @@
...
+          data?.map(({ tokenId }) => {
+            const listedFungibleToken = fungibleTokenList?.find((token) => token.id === tokenId)
+            const alreadyAddedToArray = acc.some((token) => token.id === listedFungibleToken?.id)
+
+            if (listedFungibleToken && !alreadyAddedToArray) acc.push(listedFungibleToken)
+          })
+          return acc
+        },
+        // Include ALPH in the results
+        [ALPH] as TokenInfo[]
```
mvaivre, line 51: That's an elegant way to do it! 💡

apps/desktop-wallet/src/api/addressesFungibleTokensPricesDataHooks.tsx
```diff
@@ -98,14 +102,32 @@ export const useSortTokensByWorth = (tokens: Asset[]) => {
   )
 }
 
-export const useAddressesTokensWorth = (addressHashes?: AddressHash[] | AddressHash) => {
-  const { data: tokenPrices } = useAddressesTokensPrices()
-  const selectAddressesKnownFungibleTokens = useMemo(makeSelectAddressesKnownFungibleTokens, [])
-  const tokens = useAppSelector((s) => selectAddressesKnownFungibleTokens(s, addressHashes))
+export const useAddressesTokensWorth = (addressHash?: AddressHash) => {
```
mvaivre, line 105: Ok so, if the `addressHash` is omitted, this will return the list of token worth for all addresses, correct? 
I don't know if I like that or not 🤔 
I probably need to think more about this and we could discuss this, but intuitively I don't find this super explicit.

apps/desktop-wallet/src/api/transactionQueries.ts
```diff
@@ -0,0 +1,58 @@
...
+  queryOptions({
+    queryKey: [...ADDRESS_TRANSACTIONS_QUERY_KEYS, 'latest', { addressHash, networkId }],
+    queryFn: async ({ queryKey }) => {
+      const transactions = await client.explorer.addresses.getAddressesAddressTransactions(addressHash, {
+        page: 1,
+        limit: 1
+      })
+
+      const latestTxHash = transactions.length > 0 ? transactions[0].hash : undefined
+      const cachedData = queryClient.getQueryData(queryKey) as AddressLatestTransactionHashQueryData | undefined
```
mvaivre, line 47: Pretty cool.

apps/desktop-wallet/src/api/addressesFungibleTokensPricesDataHooks.tsx
nop33: We have already been using this pattern a lot in the current codebase. You can look for it in components that both work in the overview page as well as in address-specific pages/modals. It's a simple way to achieve something that otherwise is complicated. Passing an array of `[addressHash]` always creates a new reference to an object. Having a `string | undefined` instead is simpler and more efficient. Happy to discuss alternative naming.

mvaivre: Yeah, the alternative would be to create multiple functions. `useAddressTokensWorth(string)`, `useAddressesTokensWorth([string])`, `useAllAddressesTokensWorth`. I went that way in my Tanstack branch: creating a base hook which is reused in other smol more specific hooks. 

But again, I'm not yet sure which option is the best.

apps/desktop-wallet/src/features/historicChart/useHistoricData.tsx
```diff
@@ -109,17 +110,17 @@ const useHistoricData = () => {
...
     })
   })
 
   return {
     alphBalanceHistoryPerAddress,
-    isPendingHistoricalAlphBalances,
-    isPendingAlphPriceHistory,
+    isLoadingHistoricalAlphBalances,
+    isLoadingAlphPriceHistory,
     alphPriceHistory,
```
mvaivre, line 122: Should we systematically structure the output data in an object like `data: {....}`?
Also, is it useful to keep 2 different `isLoading` states?

Right now I think that it'd make things clearer if data hooks systematically return an object structured like `{data: {}, isLoading: boolean}`. WDYT?

Issue #741, opened 2024-07-17
[DW] Losing metadata of address labels


nit27:
As title, the issue is found at `Overview`, `Transfers`, and `Addresses`

<img width="750" alt="Screenshot 2024-07-17 at 14 26 46" src="https://github.com/user-attachments/assets/a5621b16-11a4-417e-98b5-0e7da705dbeb">

As you can see although they're losing their names of label, they also show the colour and they are in the rectangles  

<img width="350" alt="Screenshot 2024-07-17 at 14 26 21" src="https://github.com/user-attachments/assets/6d2779bc-4ec2-4712-83b9-f3db9735bc19">

<img width="350" alt="Screenshot 2024-07-17 at 14 26 07" src="https://github.com/user-attachments/assets/abc2c173-8c4e-40ac-97a1-02f3c5650e03">

Environment:
- MacOS 13.6.7
- DW 2.3.6

nop33:
Is this a passphrase wallet?

nop33:
We are planning on adding further explanations in the UI:

- https://github.com/alephium/alephium-frontend/issues/42

nit27:
No, it isn't. The wallet only created by a seed phrase.

Hmm, you made me notice that there is no the `Default address` on top right as usual. Strange.

nit27:
- I try to label them again, but cannot. The wallet doesn't response for `Address options` to label an address when I hit to the button `Settings`.
<img width="450" alt="Screenshot 2024-07-17 at 15 28 34" src="https://github.com/user-attachments/assets/150f1fbb-0c53-45ee-a000-d2caf33c2676">  

-----
- Also try to create a new wallet with new seed, the issue is gone (and the `Default address` on the top right comes back 😅

nop33:
Hm, looks like the default address was "forgotten" somehow...

mvaivre:
I encountered this once, but I thought it was because of a corrupted state in dev mode. It was indeed because the default address was not defined.

nit27:
I've just re-installed my MacOS (made sure my computer was not compromised) and re-imported the wallets. The issue happened again with **the 1st seed**, but the **2nd seed** 🤔. Also seen that the `Send` button of the 1st wallet didn't response. Therefore, I removed the wallet with the 1st seed. Re-imported the 1st seed and now all wallets works temporarily fine.

*p.s: Multiple wallets on Linux doesn't cause this issue.*

nop33:
[...]

PR #746, opened 2024-07-23
Return first address if default not found

nop33:
Related to #741

It seems that more people are facing this bug. As a quick solution, I propose that if a default address is not found we return the first one.


mvaivre (approved):
Thanks

nop33:
@mvaivre I added another commit to make sure that if no default address if found in the metadata we update the stored metadata.

mvaivre (approved):

mvaivre (approved):
👏

PR #748, opened 2024-07-23
Provide data to AssetList with Tanstack

nop33:
This PR focuses on providing all AssetList data through Tanstack instead of Redux. This includes:
- listed fungible tokens
- unlisted tokens which include:
   - unlisted fungible tokens
   - NFTs
   - non-standard tokens (I think this is a more accurate name than "Unknown" tokens, since "unknown" is anything that is not listed - the tokens type is "unknown" until we fetch it, and then it's one of "`fungible"`, `"non-fungible"`, `"non-standard"`, `string`, or `undefined`.)
 
image


apps/desktop-wallet/src/api/addressesNftsDataHooks.ts
```diff
@@ -0,0 +1,95 @@
...
+    queries: chunk(nftIds, TOKENS_QUERY_LIMIT).map((ids) => ({
+      queryKey: ['tokens', 'non-fungible', 'metadata', ids],
+      queryFn: () => client.explorer.tokens.postTokensNftMetadata(ids),
+      staleTime: Infinity
+    })),
+    combine: (results) => ({
+      data: results.flatMap(({ data }) => data).filter(isDefined),
+      isLoading: results.some(({ isLoading }) => isLoading)
+    })
+  })
```
nop33, line 53: First, getting the metadata (including tokenUri)

apps/desktop-wallet/src/api/addressesNftsDataHooks.ts
```diff
@@ -0,0 +1,95 @@
...
+              })
+        } as NFT
+      },
+      staleTime: ONE_DAY_MS
+    })),
+    combine: (results) => ({
+      data: results.map(({ data }) => data).filter(isDefined),
+      isLoading: results.some(({ isLoading }) => isLoading)
+    })
+  })
```
nop33, line 77: Then getting the data, from the tokenUri

apps/desktop-wallet/src/api/addressesTokensPricesDataHooks.ts
```diff
@@ -52,7 +55,7 @@ export const useAddressesTokensPrices = () => {
               symbol: symbols[i]
             }) as TokenPrice
         ),
-      initialData: () =>
+      placeholderData: () =>
```
nop33, line 58: One big difference between `initialData` and `placeholderData` is that the former is added to the cache and the query is considered `fresh` until it becomes `stale` based on the `staleTime`, whereas the latter is not added to the cache and the query runs immediately. That's the behavior we want here.

apps/desktop-wallet/src/api/addressesTokensPricesDataHooks.ts
```diff
@@ -102,21 +105,71 @@ export const useSortTokensByWorth = (tokens: Asset[]) => {
   )
 }
 
+type TokenId = string
+
 export const useAddressesTokensWorth = (addressHash?: AddressHash) => {
```
nop33, line 110: I changed the implementation of this hook to return a `Record` that maps a tokenID to the total worth. Later on I introduced the `useAddressesTokensTotalWorth` that implements the previous behavior of `useAddressesTokensWorth`.

apps/desktop-wallet/src/api/utilHooks.ts
nop33, line 1: The hooks of this file are utility hooks for other data hooks, not meant to (at least currently) be used by components.

apps/desktop-wallet/src/components/NFTCard.tsx
```diff
@@ -16,22 +16,24 @@ You should have received a copy of the GNU Lesser General Public License
...
 interface NFTCardProps {
   nftId: NFT['id']
   onClick?: () => void
 }
 
 const NFTCard = ({ nftId, onClick }: NFTCardProps) => {
-  const nft = useAppSelector((s) => selectNFTById(s, nftId))
+  const { data: nfts } = useAddressesNfts()
+
+  const nft = nfts.find((nft) => nft.id === nftId)
```
nop33, line 36: This doesn't look like a very smart way of getting the NFT data, but with the current setup it works. Maybe we can return a `Record<NftId, NFT>` instead of `NFT[]` from the `useAddressesNfts`. To be looked into during optimization phase.

apps/desktop-wallet/src/features/assetsLists/AssetsTabs.tsx
nop33, line 1: The initial `AssetsList.tsx` file that was part of the Overview page (but the AddressDetailsModal was reusing it) grew too long. I decided to split it in smaller components. Thats' why we see a big diff.

In any case, the whole collection of components does not use anymore API data from Redux, but from Tanstack only. The performance is impacted, but we can look into that in when we optimize our hooks.

apps/desktop-wallet/src/api/addressesUnlistedTokensHooks.ts
```diff
@@ -0,0 +1,70 @@
...
+import { FungibleTokenMetadata } from '@alephium/web3/dist/src/api/api-explorer'
+import { useQueries } from '@tanstack/react-query'
+import { chunk } from 'lodash'
+
+import { useAddressesUnlistedTokenTypes } from '@/api/utilHooks'
+import { isDefined } from '@/utils/misc'
+
+export const useAddressesUnlistedNonStandardTokenIds = (addressHash?: AddressHash) => {
+  const {
+    data: { 'non-standard': unlinstedNonStandardTokenIds },
```
mvaivre, line 29: typo in variable name

apps/desktop-wallet/src/api/utilHooks.ts
```diff
@@ -0,0 +1,106 @@
+/*
```
mvaivre, line 1: Seems like this file contains hooks to get info about token types. Should it be named accordingly?

apps/desktop-wallet/src/api/addressesNftsDataHooks.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export const useAddressesNfts = (addressHash?: AddressHash) => {
+  const { data: nftIds, isLoading: isLoadingNftIds } = useAddressesNftsIds(addressHash)
+
+  const { data: nftsMetadata, isLoading: isLoadingNftsMetadata } = useQueries({
+    queries: chunk(nftIds, TOKENS_QUERY_LIMIT).map((ids) => ({
+      queryKey: ['tokens', 'non-fungible', 'metadata', ids],
+      queryFn: () => client.explorer.tokens.postTokensNftMetadata(ids),
+      staleTime: Infinity
```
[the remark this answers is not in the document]
nop33: Yes, check the last message in our Slack #dev-frontend channel.

apps/desktop-wallet/src/api/utilHooks.ts
nop33: I wrote about this file here: https://github.com/alephium/alephium-frontend/pull/748#discussion_r1690410473

mvaivre: But anyhow, don't you think that the name of the file should reflect its content, rather than how it should be used? As you said in your comment, we don't know if those hooks could be useful in components in the future (they very well may be).

By calling this a "util" file I just fear that it'll become a messy file with may random hooks inside :)

apps/desktop-wallet/src/api/addressesNftsDataHooks.ts
```diff
@@ -0,0 +1,95 @@
...
+    data: nftIds,
+    isLoading
+  }
+}
+
+export const useAddressesNfts = (addressHash?: AddressHash) => {
+  const { data: nftIds, isLoading: isLoadingNftIds } = useAddressesNftsIds(addressHash)
+
+  const { data: nftsMetadata, isLoading: isLoadingNftsMetadata } = useQueries({
+    queries: chunk(nftIds, TOKENS_QUERY_LIMIT).map((ids) => ({
```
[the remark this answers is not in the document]
nop33: I'm glad you brought this up! While implementing this PR there were a few moments that I started thinking that batshit would help in some cases. This is to say that I also started feeling the need to introduce it.

<details><summary>However, it's still worth clarifying some things regarding your comments, but I'll hide it cause it's not that important: TL;DR: I don't think the problem is if we want to use `useAddressesNfts` for 1 address only. It's when a new tx comes that includes a new NFT.</summary>

> Here Tanstack will store the result according to the chunks of ids provided. So if we have 3 chunks, it'll store 3 cached entries. If later on we were to use useAddressesNfts for 1 address only [...]

When the app unlocks, we always need to fetch all NFTs of **ALL** addresses, otherwise we can't know how to render the AssetList of the Overview screen (we need to know what is a fungible, what's an non-fungible, and what's a non-standard token to render the token list properly. Since we have now fetched NFTs from all addresses, there is **no longer a need to fetch for 1 address**. So, there's nothing suboptimal with this approach in this case. It's actually quite optimal since we are reusing the cache that the Overview screen created in every other screen that needs NFTs. There is, however, something suboptimal in another case. That's the case when a new transaction comes in. 

</details>

The main point you are trying to make is that there are some cases that it makes sense to let bathshit manage the "chunking" on our `POST` queries instead of us. I agree here. By managing the `POST` requests with Batshit instead of ourselves we would not be reducing the total number of requests, but we could potentially reduce the number of data that the explorer backend has to return.

This is my explanation:

Our app launches. There are 2 addresses. Each address has 50 NFTs. That's 100 NFTs.

The Overview page needs to know what to render in the token list so using the `useAddressesNfts` hook (which internally it will use the `useAddressesNftsIds` and `useAddressesUnlistedTokenTypes` hooks) it will:
1. fetch the token balances for the 2 addresses
2. combine the results to generate an array of token IDs that are not in the token list on GitHub which will be an array of 100 items that are the NFT IDs
3. chunk them in 2 chunks of 80 and 20 IDs each and get their type from explorer backend in 2 requests
4. combine the results in a `data` object that looks like this:
   ```
   {
     'fungible': [],
     'non-fungible': ['token_id_1', ..., 'token_id_100'],
     'non-standard': []
   }
   ```
5. chunk the `non-fungible` array in 2 chunks of 80 and 20 IDs each and get the NFT metadata with 2 requests
6. get the NFT data with 100 requests

Everything good here. Everything was done with the minimal number of requests to the explorer backend ✅ 

Now, a new transaction comes in in address 1, which includes a new NFT. This is what will happen:
1. Since the last tx has of the address 1 has changes, 1 request will be triggered to get the new address token balances.
2. together with the cached token balances of address 2, the results of the above request will be combined in an array of 101 items that are the NFT IDs
3. they will be chunked in 2 chunks of 80 and 21 ids each to get their type. Here's the important part:
   1. The first chunk is cached.
   2. The second chunk is new: the array of 21 IDs is different than the 20 IDs we had before. Thus, a new request will be made for all 21, and not just for the 1 new one.
4. the results will be combined in a `data` object that looks like this:
   ```
   {
     'fungible': [],
     'non-fungible': ['token_id_1', ..., 'token_id_101'],
     'non-standard': []
   }
   ```
5. we'll chunk the `non-fungible` array in 2 chunks of 80 and 21 IDs each and get the NFT metadata with 2 requests:
   1. Here, again, the first chunk is cached.
   2. The second chunk will result in a new query that will ask for metadata for all 21 NFT IDs.
6. We'll get the NFT data of the new NFT with 1 request.


Summary: When a new tx arrived that included a new NFT, 4 new requests were made:
1. x1 `GET` request to get the new NFT ID
2. x1 `POST` request with the 21 NFT IDs to get their type
3. x1 `POST` request with the 21 NFT IDs to get their metadata
4. x1 `GET` request to get the NFT data

By managing the `POST` requests with Batshit instead of ourselves we would not be reducing the total number of requests, but we could potentially reduce the number of data that the explorer backend has to return.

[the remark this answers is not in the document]
mvaivre: Quick answer before you come back:

> By managing the POST requests with Batshit instead of ourselves we would not be reducing the total number of requests, but we could potentially reduce the number of data that the explorer backend has to return.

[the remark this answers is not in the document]
mvaivre: Quick answer before you come back. First of all, great write up. That's indeed what would be going on here. 

> By managing the POST requests with Batshit instead of ourselves we would not be reducing the total number of requests, but we could potentially reduce the number of data that the explorer backend has to return.

In that case we would indeed reduce the load on the backend. But the benefits of Batshit goes further than that IMO:
- If we systematically store single items in Tanstack (instead of arrays of items) and let Batshit batch the calls, we will be 100% sure that if a set of items is requested, only the individual items not present in the cache will be fetched, which is optimal. And this granularity may help actually reduce the number of calls needed in other use cases.
- Applying the same fetch strategy systematically reduces our mental load and simplifies the code. The endpoint requires an array of ids? => use batshit.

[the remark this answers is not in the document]
nop33: I wanted to first explore the limitations of NOT using batshit due to the recent issues that we found with explorer and types. That's why I wanted to manage the `POST` requests ourselves. Now that it becomes apparent that this approach has its limits, I would like to explore integrating batshit and figure out ways to avoid its limits.

apps/desktop-wallet/src/api/addressesNftsDataHooks.ts
```diff
@@ -0,0 +1,95 @@
...
+      queryKey: ['tokens', 'non-fungible', 'metadata', ids],
+      queryFn: () => client.explorer.tokens.postTokensNftMetadata(ids),
+      staleTime: Infinity
+    })),
+    combine: (results) => ({
```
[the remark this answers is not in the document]
nop33: We don't always use the same combine function, but yes, eventually we could dry them up.

apps/desktop-wallet/src/api/utilHooks.ts
mvaivre: Ok! Let's start like this.

mvaivre (approved):
Great work!

PR #751, opened 2024-07-30
Banxa

mvaivre:
Ready for a first pass of testing & review. Testing is especially needed on Android.

### DIFF
--- a/apps/mobile-wallet/locales/en-US/translation.json
+++ b/apps/mobile-wallet/locales/en-US/translation.json
@@ -367,7 +367,5 @@
   "View full size": "View full size",
   "Disclaimer": "Disclaimer",
   "Buy": "Buy",
-  "Could not find a receiving address": "Could not find a receiving address",
-  "Cannot proceed because no address has been found to receive the funds.": "Cannot proceed because no address has been found to receive the funds.",
   "banxaDisclaimer": "You are about to access 3rd party services provided by <1>Banxa.com</1> through an in-app browser. Alephium does not control Banxa’s services. Banxa’s terms and conditions will apply, so please read and understand them before proceeding."
 }
--- a/apps/mobile-wallet/locales/en-US/translation.json
+++ b/apps/mobile-wallet/locales/en-US/translation.json
@@ -369,5 +369,5 @@
   "Buy": "Buy",
   "Could not find a receiving address": "Could not find a receiving address",
   "Cannot proceed because no address has been found to receive the funds.": "Cannot proceed because no address has been found to receive the funds.",
-  "banxaDisclaimer": "You are about to access 3rd party services provided by <1/> through an in-app browser. Alephium does not control Banxa’s services. Banxa’s terms and conditions will apply, so please read and understand them before proceeding."
+  "banxaDisclaimer": "You are about to access 3rd party services provided by <1>Banxa.com</1> through an in-app browser. Alephium does not control Banxa’s services. Banxa’s terms and conditions will apply, so please read and understand them before proceeding."
 }
--- a/apps/mobile-wallet/locales/en-US/translation.json
+++ b/apps/mobile-wallet/locales/en-US/translation.json
@@ -367,5 +367,7 @@
   "View full size": "View full size",
   "Disclaimer": "Disclaimer",
   "You are about to access 3rd party services provided by Banxa.com through an in-app browser. Alephium does not control Banxa’s services. Banxa’s terms and conditions will apply, so please read and understand them before proceeding.": "You are about to access 3rd party services provided by Banxa.com through an in-app browser. Alephium does not control Banxa’s services. Banxa’s terms and conditions will apply, so please read and understand them before proceeding.",
-  "Buy": "Buy"
+  "Buy": "Buy",
+  "Could not find a receiving address": "Could not find a receiving address",
+  "Cannot proceed because no address has been found to receive the funds.": "Cannot proceed because no address has been found to receive the funds."
 }
--- a/apps/mobile-wallet/locales/en-US/translation.json
+++ b/apps/mobile-wallet/locales/en-US/translation.json
@@ -366,8 +366,8 @@
   "A number is expected": "A number is expected",
   "View full size": "View full size",
   "Disclaimer": "Disclaimer",
-  "You are about to access 3rd party services provided by Banxa.com through an in-app browser. Alephium does not control Banxa’s services. Banxa’s terms and conditions will apply, so please read and understand them before proceeding.": "You are about to access 3rd party services provided by Banxa.com through an in-app browser. Alephium does not control Banxa’s services. Banxa’s terms and conditions will apply, so please read and understand them before proceeding.",
   "Buy": "Buy",
   "Could not find a receiving address": "Could not find a receiving address",
-  "Cannot proceed because no address has been found to receive the funds.": "Cannot proceed because no address has been found to receive the funds."
+  "Cannot proceed because no address has been found to receive the funds.": "Cannot proceed because no address has been found to receive the funds.",
+  "banxaDisclaimer": "You are about to access 3rd party services provided by <1/> through an in-app browser. Alephium does not control Banxa’s services. Banxa’s terms and conditions will apply, so please read and understand them before proceeding."
 }
--- a/apps/mobile-wallet/locales/en-US/translation.json
+++ b/apps/mobile-wallet/locales/en-US/translation.json
@@ -364,5 +364,8 @@
   "Error while copying address": "Error while copying address",
   "This address is not valid": "This address is not valid",
   "A number is expected": "A number is expected",
-  "View full size": "View full size"
+  "View full size": "View full size",
+  "Disclaimer": "Disclaimer",
+  "You are about to use an external service provider, banxa.com, which allows you to buy ALPH right from your wallet. Alephium has no control over Banxa activities. Please reach Banxa for more information.": "You are about to use an external service provider, banxa.com, which allows you to buy ALPH right from your wallet. Alephium has no control over Banxa activities. Please reach Banxa for more information.",
+  "Buy": "Buy"
 }
--- a/apps/mobile-wallet/locales/en-US/translation.json
+++ b/apps/mobile-wallet/locales/en-US/translation.json
@@ -366,6 +366,6 @@
   "A number is expected": "A number is expected",
   "View full size": "View full size",
   "Disclaimer": "Disclaimer",
-  "You are about to use an external service provider, banxa.com, which allows you to buy ALPH right from your wallet. Alephium has no control over Banxa activities. Please reach Banxa for more information.": "You are about to use an external service provider, banxa.com, which allows you to buy ALPH right from your wallet. Alephium has no control over Banxa activities. Please reach Banxa for more information.",
+  "You are about to access 3rd party services provided by Banxa.com through an in-app browser. Alephium does not control Banxa’s services. Banxa’s terms and conditions will apply, so please read and understand them before proceeding.": "You are about to access 3rd party services provided by Banxa.com through an in-app browser. Alephium does not control Banxa’s services. Banxa’s terms and conditions will apply, so please read and understand them before proceeding.",
   "Buy": "Buy"
 }
--- a/apps/mobile-wallet/src/components/layout/BottomModal.tsx
+++ b/apps/mobile-wallet/src/components/layout/BottomModal.tsx
@@ -40,7 +40,7 @@ import { DEFAULT_MARGIN, VERTICAL_GAP } from '~/style/globalStyle'
 
 type ModalPositions = 'minimised' | 'maximised' | 'closing'
 
-const NAV_HEIGHT = 48
+const NAV_HEIGHT = 40
 const DRAG_BUFFER = 40
 
 const springConfig: WithSpringConfig = {
@@ -154,7 +154,7 @@ const BottomModal = ({
   const handleMaximize = useCallback(() => {
     'worklet'
 
-    navHeight.value = withSpring(NAV_HEIGHT + 10, springConfig)
+    navHeight.value = withSpring(NAV_HEIGHT + 6, springConfig)
     modalHeight.value = withSpring(-maxHeight, springConfig)
     position.value = 'maximised'
   }, [maxHeight, modalHeight, navHeight, position])
@@ -226,7 +226,7 @@ const BottomModal = ({
             </HandleContainer>
             <Navigation style={modalNavigationAnimatedStyle}>
               {title && <Title>{title}</Title>}
-              <Button onPress={handleClose} iconProps={{ name: 'close-outline' }} round />
+              <Button onPress={handleClose} iconProps={{ name: 'close-outline' }} round compact />
             </Navigation>
             <ContentContainer noPadding={noPadding}>
               <Content onClose={handleClose} onContentSizeChange={handleContentSizeChange} />
--- a/apps/mobile-wallet/src/components/layout/BottomModal.tsx
+++ b/apps/mobile-wallet/src/components/layout/BottomModal.tsx
@@ -226,7 +226,7 @@ const BottomModal = ({
             </HandleContainer>
             <Navigation style={modalNavigationAnimatedStyle}>
               {title && <Title semiBold>{title}</Title>}
-              <Button onPress={handleClose} iconProps={{ name: 'close-outline' }} round compact />
+              <Button onPress={handleClose} iconProps={{ name: 'x' }} round compact />
             </Navigation>
             <ContentContainer noPadding={noPadding}>
               <Content onClose={handleClose} onContentSizeChange={handleContentSizeChange} />
--- a/apps/mobile-wallet/src/components/layout/BottomModal.tsx
+++ b/apps/mobile-wallet/src/components/layout/BottomModal.tsx
@@ -33,6 +33,7 @@ import Animated, {
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
 import styled from 'styled-components/native'
 
+import AppText from '~/components/AppText'
 import Button from '~/components/buttons/Button'
 import { ModalContentProps } from '~/components/layout/ModalContent'
 import { DEFAULT_MARGIN, VERTICAL_GAP } from '~/style/globalStyle'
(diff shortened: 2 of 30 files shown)


apps/mobile-wallet/src/features/buy/BuyModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+      BackHandler.addEventListener('hardwareBackPress', onAndroidBackPress)
+      return () => {
+        BackHandler.removeEventListener('hardwareBackPress', onAndroidBackPress)
+      }
+    }
+  }, [])
+
+  const banxaURL =
+    'https://alephium.banxa-sandbox.com/' +
+    `?walletAddress=${defaultAddress?.hash}` +
```
nop33, line 68: Does it make sense to build the URL with the wallet address being `undefined`? Even though we shouldn't have the case that the defaultAddress is `undefined`, maybe we should not render this modal if that's the case, or at least initialize it with the first available address.

apps/mobile-wallet/src/features/buy/BuyModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+    `&theme=${theme.name}` +
+    `&backgroundColor=${theme.bg.primary.slice(1)}` +
+    `&textColor=${theme.font.primary.slice(1)}` +
+    `&primaryColor=${theme.global.accent}` +
+    `&secondaryColor=${theme.global.complementary}`
```
nop33, line 73: For `backgroundColor` and `textColor` we should skip the `#` but not for `primaryColor` and `secondaryColor`?

apps/mobile-wallet/src/screens/Dashboard/DashboardScreen.tsx
```diff
@@ -122,19 +121,11 @@ const DashboardScreen = ({ navigation, ...props }: ScreenProps) => {
...
-        headerLeft: () => <WalletSwitchButton />,
-        headerTitle: walletName
+        headerLeft: () => <WalletSwitchButton />
       }}
-      TitleSideComponent={
-        <NetworkBadgeContainer>
-          <Pressable onPress={() => setIsSwitchNetworkModalOpen(true)}>
-            <ActiveNetworkBadge />
-          </Pressable>
-        </NetworkBadgeContainer>
```
nop33, line 136: Is this component now gone, or transfered somewhere else?

nop33 (commented):
Very quick first code pass. 

There's a conflict with the pnpm lock file

apps/mobile-wallet/src/screens/Dashboard/DashboardScreen.tsx
mvaivre: It's gone

apps/mobile-wallet/src/features/buy/BuyModal.tsx
mvaivre: Good catch

apps/mobile-wallet/src/features/buy/BuyModal.tsx
nop33, line 1: I cannot scroll to click the button

https://github.com/user-attachments/assets/439b46e0-0ccd-426c-8b85-5b1328b21ed1

apps/mobile-wallet/src/features/buy/BuyModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+      <BottomModal
+        {...props}
+        Content={(props) => (
+          <ModalContent {...props} contentContainerStyle={{ flex: 1, paddingTop: 0 }}>
+            {!isDisclaimerAccepted && (
+              <DisclaimerContent>
+                <ScreenTitle title={t('Disclaimer')} />
+                <AppText style={{ flex: 1 }}>
+                  {t(
+                    'You are about to access 3rd party services provided by Banxa.com through an in-app browser. Alephium does not control Banxa’s services. Banxa’s terms and conditions will apply, so please read and understand them before proceeding.'
```
nop33, line 108: Wouldn't it be fair for the user to link them to the terms and conditions of Banxa here?

image

apps/mobile-wallet/src/components/layout/BottomModal.tsx
```diff
@@ -267,11 +279,21 @@ const Handle = styled(Animated.View)`
   margin-top: -15px;
 `
 
+const Title = styled(AppText)`
+  position: absolute;
+  right: 0;
+  left: 0;
+  text-align: center;
+`
```
nop33, line 287: With absolute positioning, if the title of a bottom modal is long (can be caused by translations of different screen resolutions) it will overflow on top of the "X" button

image

nop33: The way I would address this would be to not use absolute positioning and add an invisible flexbox item on the left that has the same size as the "X" button on the right. Then, the title can take all available space and still be in the center.

apps/mobile-wallet/src/features/buy/BuyModal.tsx
```diff
@@ -0,0 +1,138 @@
...
+    'https://alephium.banxa-sandbox.com/' +
+    `?walletAddress=${defaultAddress?.hash}` +
+    `&theme=${theme.name}` +
+    `&backgroundColor=${theme.bg.primary.slice(1)}` +
+    `&textColor=${theme.font.primary.slice(1)}` +
+    `&primaryColor=${theme.global.accent}` +
+    `&secondaryColor=${theme.global.complementary}`
+
+  const handleModalCloseWhenFinished = (e: WebViewNavigation) => {
+    // Close modal if url equals default return URL confirgured in Banxa dashboard
```
nop33, line 90: typo in comment :)

apps/mobile-wallet/src/features/buy/BuyModal.tsx
mvaivre: The link to their T&C can change, so I'd rather keep it simple if possible

nop33: Then maybe at list link to Banxa.com?

mvaivre:
All comments have been addressed @nop33 🙏

apps/mobile-wallet/src/features/buy/BuyModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const receivingAddress = defaultAddress?.hash || allAddresses[0]?.hash
+
+  if (!receivingAddress) {
+    showToast({
+      text1: t('Could not find a receiving address'),
+      text2: t('Cannot proceed because no address has been found to receive the funds.'),
+      type: 'error'
+    })
+    return null
+  }
```
nop33, line 79: Hm, I am not sure about this implementation. It will run on every re-render. Things like this should be inside a useEffect. Besides I don't see how valuable this is for the user. I would simply `return null` if there is no default address. There should always be a default address. And if for whatever reason there is non, there is at least 1 address. And if there isn't a single address then we have bigger problems.

apps/mobile-wallet/src/components/layout/BottomModal.tsx
```diff
@@ -213,10 +224,14 @@ const BottomModal = ({ Content, isOpen, onClose, maximisedContent, customMinHeig
...
             </HandleContainer>
-            <ContentContainer>
-              <Navigation style={modalNavigationAnimatedStyle}>
-                <Button onPress={handleClose} iconProps={{ name: 'close-outline' }} round />
-              </Navigation>
+            <Navigation style={modalNavigationAnimatedStyle}>
+              <NavigationButtonContainer align="left" />
+              {title && <Title semiBold>{title}</Title>}
+              <NavigationButtonContainer align="right">
+                <Button onPress={handleClose} iconProps={{ name: 'x' }} round compact />
```
nop33, line 231: Since we are at it, wouldn't it better to reuse the `CloseButton` which we are already using in all modals that implement `ModalWithBackdrop` as well as `AddressScreen` and `DestinationScreen` for a more consistent UI?

We also have to keep in mind Google's accessibility recommendation when it comes to the clickable area of buttons which (IIRC) needs to be at least 48dp. That doesn't mean that the visual button needs to be that big, but the clickable area (example: padding around it). We have a lot of buttons that need to be improved.

At least, if we reuse the same buttons, we can fix them all at once.

image

apps/mobile-wallet/src/components/layout/BottomModal.tsx
```diff
@@ -267,11 +282,25 @@ const Handle = styled(Animated.View)`
...
 const Navigation = styled(Animated.View)`
-  align-items: flex-end;
+  flex-direction: row;
+  align-items: center;
+  justify-content: flex-end;
+  padding: 0 ${DEFAULT_MARGIN - 1}px;
+`
+
+const NavigationButtonContainer = styled.View<{ align: 'right' | 'left' }>`
+  width: 10%;
```
nop33, line 298: That's one way to do it! But when the width of the close button is larger than 10% of the container, things would be misaligned.. This can happen on small screens.

image


For perfect alignment I can think of 2 options:
1. render an invisible button to ensure both left and right flexbox items have exactly the same width 👍 
2. measure the width of the last flexbox item and programmatically set the width of the first item 👎

apps/mobile-wallet/src/features/buy/BuyModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  const banxaURL =
+    'https://alephium.banxa-sandbox.com/' +
+    `?walletAddress=${receivingAddress}` +
```
nop33, line 83: Shouldn't the user be allowed to configure the receiving address?

apps/mobile-wallet/src/features/buy/BuyModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+                    t={t}
+                    i18nKey="banxaDisclaimer"
+                    components={{
+                      1: <LinkToWeb url="https://www.banxa.com" text="Banxa.com" />
+                    }}
+                  >
+                    {
+                      'You are about to access 3rd party services provided by <1/> through an in-app browser. Alephium does not control Banxa’s services. Banxa’s terms and conditions will apply, so please read and understand them before proceeding.'
+                    }
+                  </Trans>
```
nop33, line 118: This might cause confusion to our translators because they have no context as to what the `[...] by <1/> through [...]` is. Is it a link? A name of a company? A physical person? Each of these might have different articles and prepositions based on the language. Normally, in these cases, we would have `[...] by <1>Banxa.com</1> through [...]`. We can enhance the translators experience and avoid wrong articles and prepositions by making a small adjustment to the `LinkToWeb` component to accept `children: string` instead of `text: string` and updating to

```suggestion
                  <Trans
                    t={t}
                    i18nKey="banxaDisclaimer"
                    components={{
                      1: <LinkToWeb url="https://www.banxa.com" />
                    }}
                  >
                    {
                      'You are about to access 3rd party services provided by <1>Banxa.com></1> through an in-app browser. Alephium does not control Banxa’s services. Banxa’s terms and conditions will apply, so please read and understand them before proceeding.'
                    }
                  </Trans>

**Review by nop33 — COMMENTED** [review]
On Android I see a bottom padding, is that intended?

image


Also, I didn't get an email, did you get passed this step?

image

File: apps/mobile-wallet/src/features/buy/BuyModal.tsx (hunk 10 above)
Comment (mvaivre) [reply apps/mobile-wallet/src/features/buy/BuyModal.tsx:83]: I don't know. I'd say no, this would avoid mistakes and increase security.

File: apps/mobile-wallet/src/components/layout/BottomModal.tsx (hunk 9 above)
Comment (mvaivre) [reply apps/mobile-wallet/src/components/layout/BottomModal.tsx:298]: I'm not sure I'm getting that! What I want is the left and right margin of the header to stay consistent. What would be misaligned? We can have a tiny call to clarify this :)

File: apps/mobile-wallet/src/components/layout/BottomModal.tsx (hunk 8 above)
Comment (mvaivre) [reply apps/mobile-wallet/src/components/layout/BottomModal.tsx:231]: That's a good point. Thanks.

File: apps/mobile-wallet/src/features/buy/BuyModal.tsx (hunk 10 above)
Comment (nop33) [reply apps/mobile-wallet/src/features/buy/BuyModal.tsx:83]: No, I mean, shouldn't the user select which of their addresses they want to use, like we do when sending/receiving funds?

File: apps/mobile-wallet/src/components/layout/BottomModal.tsx (hunk 9 above)
Comment (nop33) [reply apps/mobile-wallet/src/components/layout/BottomModal.tsx:298]: I gave it another thought. It's fine with 10% both sides. Resolving.

File: apps/mobile-wallet/src/features/buy/BuyModal.tsx (hunk 10 above)
Comment (mvaivre) [reply apps/mobile-wallet/src/features/buy/BuyModal.tsx:83]: For our first implementation I'm tempted to say no.
- Let's reduce the pain and time needed to buy (there's the KYC etc.) by removing a step if we can
- Let's see what's the conversion rate
- Let's collect feedback

And then we'll reassess.

**Comment by mvaivre** [comment]
@nop33 about testing, you don't receive an email. You can use a fake email. Just use the PIN code (7203) mentioned in the docs: https://docs.banxa.com/docs/testing-information

File: apps/mobile-wallet/src/features/buy/BuyModal.tsx (hunk 7 above)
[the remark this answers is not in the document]
Comment (mvaivre) [reply apps/mobile-wallet/src/features/buy/BuyModal.tsx:79]: Returning an error to the user is useful for reporting in case of edge cases like the ones we had recently. 
Other than that I agree with you!

File: apps/mobile-wallet/src/features/buy/BuyModal.tsx (hunk 11 above)
Comment (mvaivre) [reply apps/mobile-wallet/src/features/buy/BuyModal.tsx:106]: Yep, I hesitated. I'll do it.

Comment (mvaivre) [reply apps/mobile-wallet/src/features/buy/BuyModal.tsx:106]: It actually doesn't work 🤔 Child doesn't seem to be passed to the translated component.

image

Comment (mvaivre) [reply apps/mobile-wallet/src/features/buy/BuyModal.tsx]: Done

File: apps/mobile-wallet/src/components/layout/BottomModal.tsx (hunk 8 above)
Comment (mvaivre) [reply apps/mobile-wallet/src/components/layout/BottomModal.tsx]: Done

File: apps/mobile-wallet/src/features/buy/BuyModal.tsx (hunk 7 above)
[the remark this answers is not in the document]
Comment (mvaivre) [reply apps/mobile-wallet/src/features/buy/BuyModal.tsx]: Fixed

File: apps/mobile-wallet/src/components/text/LinkToWeb.tsx [hunk 12]
```diff
@@ -16,18 +16,19 @@ You should have received a copy of the GNU Lesser General Public License
...
+import { ReactNode } from 'react'
 import { Linking, StyleProp, TextStyle } from 'react-native'
 import styled from 'styled-components/native'
 
 import AppText from '~/components/AppText'
 
 interface LinkToWebProps {
-  text: string
   url: string
+  children?: ReactNode
```
nop33, line 27: Shouldn't the children still be only `string`? Do we want to allow other components inside a `LinkToWeb` instance?

apps/mobile-wallet/locales/en-US/translation.json
```diff
@@ -360,5 +360,10 @@
   "Error while copying address": "Error while copying address",
   "This address is not valid": "This address is not valid",
   "A number is expected": "A number is expected",
-  "View full size": "View full size"
+  "View full size": "View full size",
+  "Disclaimer": "Disclaimer",
+  "Buy": "Buy",
+  "Could not find a receiving address": "Could not find a receiving address",
+  "Cannot proceed because no address has been found to receive the funds.": "Cannot proceed because no address has been found to receive the funds.",
```
nop33, line 367: We don't need these 2 strings anymore, it's better to remove them to not add extra workload for the translators.

apps/mobile-wallet/src/features/buy/BuyModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const defaultAddress = useAppSelector(selectDefaultAddress)
+  const [isDisclaimerAccepted, setIsDisclaimerAccepted] = useState(false)
+
+  const onAndroidBackPress = () => {
+    if (webViewRef.current) {
+      webViewRef.current.goBack()
+      return true // prevent default behavior (exit app)
+    }
+    return false
+  }
```
nop33, line 56: nit: I would move the declaration of this function under line 59 (inside the condition of the `useEffect` so that a new function is not declared on every single rerender, but only once, when the modal is opened, and if the platform is Android.

nop33 (commented):
Good job on the fixes 🪇 

A couple more comments. I'll test with the instructions you provided in a bit.

apps/mobile-wallet/src/components/text/LinkToWeb.tsx
mvaivre: I thought that it could indeed be handy, to wrap tags, buttons etc. with LinkToWeb.

apps/mobile-wallet/locales/en-US/translation.json
mvaivre: Removed

mvaivre:
>The only feedback I have is that when clicking "Return to Alephium" our website is loaded for 1-2 seconds.

apps/mobile-wallet/src/features/buy/BuyModal.tsx
mvaivre: Done

mvaivre:
Done @nop33  Waiting for last feedback about the redirect page name + LinkToWeb children prop :)

apps/mobile-wallet/src/components/text/LinkToWeb.tsx
nop33: I think it's best to allow `ReactNode` when we actually need them for a more type-strict code-base. I change this discussion to `nit`. I'd do it if it was up to me, but it's up to you 😉

apps/mobile-wallet/src/features/buy/BuyModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+              containerStyle={{ padding: 0 }}
+              allowsBackForwardNavigationGestures
+              onNavigationStateChange={handleModalCloseWhenFinished}
+              nestedScrollEnabled
+            />
+          </ModalContent>
+        )}
+        title={t('Buy')}
+        maximisedContent
+        noPadding
```
nop33, line 133: super tiny nit: I'd move these props to the top because they are kinda lost down there.

nop33 (approved):
Left 2 more nit comments, so I am approving ✅

Issue #764, opened 2024-08-07
Keyboard prevents input for tokens located at the bottom of the screen


MaudSimon:
**Describe the bug**
Keyboard prevent input for tokens amount located at the bottom of the screen. 
"There is no way to remove the keyboard to select the number of coins of the asset located at the bottom of the screen." Video was provided to @mvaivre, not linked here for privacy. 

**To Reproduce**
Make a transaction on an address with multiple assets, once opened, the keyboard blocks access to input fields in the lower screen and there is no way to remove the keyboard.

**Expected behavior**
Possibility to "collapse/remove" the keyboard to access input field in the background screen. 

**Environment**
This was reported on IOS 16.7.8 but most likely affects all IOS devices ? Not sure on Android.

PR #765, opened 2024-08-07
Fix keyboard content overlap

nop33:
- #764


mvaivre (approved):
Tested, seems to work well! GJ

PR #767, opened 2024-08-07
Allow user to enable device passcode for app access/tx

### LINKED ISSUE #766: Allow user to enable device passcode/pattern for app access/transactions
A user reported this case to me: They are asking me: Why do we not use the iOS device pin when there is no Face ID enrolled, like Binance does?

I thought we had device passcode fallback. And we do. But only if the biometrics settings in our app is FIRST enabled. But to allow the user to enable them, we require that biometrics are enrolled. We obviously do not consider device passcode to be biometrics (even though it is used as fallback), so we disable the biometrics settings.

To make this work, we just need to let the user enable the biometrics settings even when they have no biometrics data enrolled.

Since this is a bit controversial, I think the best would be to just change the UI to not say "biometrics" but "device passcode".

nop33:
Closes #766


mvaivre (approved):
👍

PR #774, opened 2024-08-10
Bump mobile wallet to 1.1.3

nop33:
PR created so that I can build an internal release with the current version number which will later on be promoted to a production release if everything goes well.

### DIFF
--- a/.changeset/cold-carpets-kneel.md
+++ b/.changeset/cold-carpets-kneel.md
@@ -1,5 +1,0 @@
----
-"@alephium/mobile-wallet": patch
----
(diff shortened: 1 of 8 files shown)


nop33:
I am currently getting an error when I try to build an internal release for iOS which at the same time registering my new iPhone in the devices profile. Will try again later.

```
✖ Failed to handle Apple profiles
Failed to set up credentials.
Apple provided the following error info:
Service not available because of maintenance activities, please try again after some time.
```

mvaivre (approved):

Issue #750, opened 2024-07-26
Black screen on wallet creation on iOS


mvaivre:
User created a wallet successfully. On second app launch, screen is black with no error.
User reinstalled the app, imported the wallet they created before, same black screen.

- iOS 15.5.1 (21F90)
- iPhone 12

nop33:
Duplicate of https://github.com/alephium/alephium-frontend/issues/690

nop33:
Now that the 1.1.3 is live on the AppStore, I asked the user who reported this to check if it fixes the problem.

nop33:
User confirmed that the problem is resolved. Closing issue.

Issue #695, opened 2024-06-25
Authentication loop in mobile wallet


nop33:
TBD

nop33:
Problem persists on some devices. Latest report we have is Android [LineageOS](https://lineageos.org/).

Issue #690, opened 2024-06-25
Mobile wallet sometimes shows black screen on cold start


nop33:
TBD

Version: 1.1.0 android

Maybe this has something to do with this line:
```
if (!lastNavigationState) resetNavigation(navigation)
```

if for some reason `lastNavigationState` is not null, but there is no screen rendered, then after unlocking we would see a black screen.

nop33:
Should be fixed by https://github.com/alephium/alephium-frontend/commit/ee7cee5fe02bf86d330d34b6611f2dd2151c7652. Let's keep an eye on it.

nop33:
I managed to reproduce this bug with my OnePlus 6T. It only happens when I have WC enabled and I quit the app. The next time I launch it it remains on a black screen.


https://github.com/user-attachments/assets/f5249722-060e-4af7-a00c-81c3e89be19a

PR #786, opened 2024-08-19
Fix stuck in black screen

nop33:
Fixes:
- #690

Hopefully also fixes:
- #750 
- #695

### DIFF
--- a/apps/mobile-wallet/src/navigation/RootStackNavigation.tsx
+++ b/apps/mobile-wallet/src/navigation/RootStackNavigation.tsx
@@ -140,11 +140,10 @@ export default RootStackNavigation
 const AppUnlockModal = () => {
   const dispatch = useAppDispatch()
   const isWalletUnlocked = useAppSelector((s) => s.wallet.isUnlocked)
+  const biometricsRequiredForAppAccess = useAppSelector((s) => s.settings.usesBiometrics)
   const navigation = useNavigation<NavigationProp<RootStackParamList>>()
   const { triggerBiometricsAuthGuard } = useBiometricsAuthGuard()
   const { t } = useTranslation()
-
-  const [isAuthModalVisible, setIsAuthModalVisible] = useState(false)
 
   const { width, height } = Dimensions.get('window')
   const [dimensions, setDimensions] = useState({ width, height })
@@ -155,10 +154,6 @@ const AppUnlockModal = () => {
     setDimensions({ width, height })
   }
 
-  const openAuthModal = useCallback(() => {
-    setIsAuthModalVisible(true)
-  }, [])
-
   const initializeAppWithStoredWallet = useCallback(async () => {
     try {
       dispatch(walletUnlocked(await getStoredWallet()))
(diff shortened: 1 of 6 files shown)


mvaivre (approved):
So hard to catch! Great job. 👏 
I'll test more on iOS when we build a rc, but so far so good.

patches/react-native-background-actions@3.0.1.patch
```diff
@@ -10,7 +10,7 @@ index b67ef4d1837363489a1749acb852997a6c93c1b5..6aa30a826c2f17a35ccd30b64291d3d2
      <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
      <application>
 -        <service android:name=".RNBackgroundActionsTask"/>
-+        <service android:name=".RNBackgroundActionsTask" android:foregroundServiceType="dataSync" />
++        <service android:name=".RNBackgroundActionsTask" android:foregroundServiceType="dataSync" android:stopWithTask="true" />
```
nop33, line 13: Adding `android:stopWithTask="true"` will quit the background task after a few seconds when then user kills the app.

apps/mobile-wallet/src/screens/LandingScreen.tsx
```diff
@@ -57,11 +59,25 @@ const LandingScreen = ({ navigation, ...props }: LandingScreenProps) => {
...
+  // 2. the user manually kills the app before the auto-lock timer completes
+  // 3. the WalletConnect feature is activated
+  // Since there is no screen in focus and since the default screen set in the RootStackNavigation is this screen, we
+  // need to navigate back to the dashboard.
+  useFocusEffect(
+    useCallback(() => {
+      if (isWalletUnlocked) resetNavigation(navigation)
+    }, [isWalletUnlocked, navigation])
+  )
+
```
nop33, line 80: I had the idea of setting the dashboard screen as the initial screen of the RootStackNavigation. I need to think more about this because it will have quite some implications for our auth logic. Fight for another day.

apps/mobile-wallet/src/navigation/RootStackNavigation.tsx
```diff
@@ -140,12 +140,11 @@ export default RootStackNavigation
 const AppUnlockModal = () => {
   const dispatch = useAppDispatch()
   const isWalletUnlocked = useAppSelector((s) => s.wallet.isUnlocked)
+  const biometricsRequiredForAppAccess = useAppSelector((s) => s.settings.usesBiometrics)
   const navigation = useNavigation<NavigationProp<RootStackParamList>>()
   const { triggerBiometricsAuthGuard } = useBiometricsAuthGuard()
   const { t } = useTranslation()
 
-  const [isAuthModalVisible, setIsAuthModalVisible] = useState(false)
```
nop33, line 147: There is no reason to maintain this value, we can calculate it (see `visible` prop of following modal)

apps/mobile-wallet/src/screens/LandingScreen.tsx
mvaivre: Yeah we'll need to test that thoroughly.

mvaivre (approved):
LGTM

PR #796, opened 2024-08-26
Remove balance data from Redux and rely on Tanstack

nop33:


mvaivre (approved):
LGTM! Only one tiny nit, so approving.

nop33:
Where's the nit? 😆

mvaivre (approved):
LOL. My comment wasn't sent and wasn't cached... and I can't remember what it was... SO... LGTM I guess 😆

nop33:
HAHAHAHAHA

apps/desktop-wallet/src/modals/SendModals/useAreAmountsWithinAvailableBalance.ts
mvaivre, line 1: Actually I remember! My nit was that I'm not super convinced by this hook file floating in the SendModals folder :)

nop33: Ah haha! Well, that's the only place it is needed. Only in the send workflow. Optimally we should have a `src/features/send/` folder and put it there. But that might require quite some refactoring which I don't think it's priority now. WDYT?

mvaivre: Ok!

PR #797, opened 2024-08-26
Split balance hooks to separate files (part 1)

nop33:
While increasing the coverage of Tanstack throughout the app during the last few weeks I discovered that I had trouble **knowing at a glance** what available hooks I had already created. I had to open the data hook files and scroll through them. Even then, knowing what each hook returns was **hard** unless I would hover over the hook name.

Then, for optimization reasons, I wanted to extract the `combine` functions out of the hook so that it doesn't get called on every component re-render.

For the above 2 reasons, I decided to:
1. split the hooks to 1 file per hook
2. define stronger return types.

I started only by cleaning up these 2 data hook files:
- alph balances
- tokens balances

The result is this:


apps/desktop-wallet/src/api/addressQueries.ts
```diff
@@ -41,14 +47,15 @@ export const addressAlphBalanceQuery = ({
...
 
         return {
           addressHash,
-          balances: {
-            balance: BigInt(balances.balance),
-            lockedBalance: BigInt(balances.lockedBalance)
+          alphBalances: {
+            totalBalance: BigInt(balances.balance),
+            lockedBalance: BigInt(balances.lockedBalance),
+            availableBalance: BigInt(balances.balance) - BigInt(balances.lockedBalance)
```
nop33, line 58: These values will be calculated only when the query is executed and then they will be cached.

I took the liberty and opportunity to finally rename this confusing `balances` to `totalBalances`. I believe having these 3 values will simply the UI logic a lot, and it will cause less errors in the future (I remember an issue in the past where I thought `balance` was the same as `availableBalance` and a sweep endpoint was broken).

apps/desktop-wallet/src/api/addressQueries.ts
```diff
@@ -64,6 +71,11 @@ export const addressAlphBalanceQuery = ({
   })
 }
 
+export type AddressTokensBalancesQueryFnData = {
+  addressHash: AddressHash
+  tokenBalances: TokenDisplayBalances[]
+}
+
```
nop33, line 78: Defining return types of queries will improve readability of our code, especially when we come back to it after some time while working on other stuff

apps/desktop-wallet/src/api/apiDataHooks/useAddressesTokensBalancesTotal.ts
```diff
@@ -0,0 +1,81 @@
...
+const useAddressesTokensBalancesTotal = (addressHash?: AddressHash): AddressesTokensBalancesTotal => {
+  const networkId = useAppSelector((s) => s.network.settings.networkId)
+  const { data: alphBalances, isLoading: isLoadingAlphBalances } = useAddressesAlphBalancesTotal(addressHash)
+  const { data: latestTxHashes, isLoading: isLoadingLatestTxHashes } = useAddressesLastTransactionHashes(addressHash)
+
+  const { data, isLoading } = useQueries({
+    queries: latestTxHashes.map(({ addressHash, latestTxHash, previousTxHash }) =>
+      addressTokensBalanceQuery({ addressHash, latestTxHash, previousTxHash, networkId })
+    ),
+    combine
```
mvaivre, line 47: Same Q about memoization here

apps/desktop-wallet/src/api/apiDataHooks/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { UseQueryResult } from '@tanstack/react-query'
+
+import { isDefined } from '@/utils/misc'
+
+export const combineIsLoading = <R>(results: UseQueryResult<R, Error>[]) => ({
+  isLoading: results.some(({ isLoading }) => isLoading)
+})
+
+export const simpleCombine = <R>(results: UseQueryResult<R, Error>[]) => ({
```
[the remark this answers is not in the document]
nop33: I am changing it in my next branch to:
```
mapCombine
flatMapCombine
```

so that should take care of your comment!

apps/desktop-wallet/src/api/apiDataHooks/useAddressesAlphBalances.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const useAddressesAlphBalances = (addressHash?: AddressHash): AddressesAlphBalances => {
+  const { data: latestTxHashes, isLoading: isLoadingLatestTxHashes } = useAddressesLastTransactionHashes(addressHash)
+  const networkId = useAppSelector((s) => s.network.settings.networkId)
+
+  const { data, isLoading } = useQueries({
+    queries: latestTxHashes.map(({ addressHash, latestTxHash, previousTxHash }) =>
+      addressAlphBalancesQuery({ addressHash, latestTxHash, previousTxHash, networkId })
+    ),
+    combine
```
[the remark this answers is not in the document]
nop33: What do you suggest to memoize? In the `useAddressesTokensBalances.ts` we need to merge the data returned from the `combine` with some other data, and for that it makes sense to `useMemo`. Here we don't have to do anything like that, we have what we need as the result of `combine`.

The data from `combine` will not change between re-renders of the same component unless the data returned from the query change. 

In `useAddressesTokensBalances.ts` we `useMemo` because we use `reduce` which returns always a new object. We don't want a new object if the deps haven't changed.

apps/desktop-wallet/src/api/apiDataHooks/useAddressesTokensBalancesTotal.ts
nop33: Same answer as above!

apps/desktop-wallet/src/api/apiDataHooks/useAddressesAlphBalances.ts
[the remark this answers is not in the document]
mvaivre: Right! 🆗

mvaivre (approved):
LGTM!

PR #801, opened 2024-08-27
Temporarily Update Explorer Backend URL

h0ngcha0:
To test the explorer backend migration to GCP.


mvaivre:
LGTM. @nop33 WDYT about releasing this today along with the addition of GoatCounter?

nop33:
@mvaivre sounds good

mvaivre:
Reverted the change in the shared package, so that the new node backend is used in the explorer only cc @h0ngcha0 @nop33

nop33 (approved):

PR #803, opened 2024-08-28
Re-generate wallet metadata if not found

nop33:
Related to:
- #695

### DIFF
--- a/apps/mobile-wallet/src/App.tsx
+++ b/apps/mobile-wallet/src/App.tsx
@@ -39,7 +39,7 @@ import { useLocalization } from '~/features/localization/useLocalization'
 import { useAppDispatch, useAppSelector } from '~/hooks/redux'
 import useLoadStoredSettings from '~/hooks/useLoadStoredSettings'
 import RootStackNavigation from '~/navigation/RootStackNavigation'
-import { getStoredWallet, storedWalletExists } from '~/persistent-storage/wallet'
+import { getStoredWalletMetadata, storedWalletExists } from '~/persistent-storage/wallet'
 import {
   makeSelectAddressesUnknownTokens,
   selectAllAddressVerifiedFungibleTokenSymbols,
@@ -100,7 +100,7 @@ const Main = ({ children, ...props }: ViewProps) => {
 
   useEffect(() => {
     storedWalletExists().then((walletExists) => {
-      if (walletExists) getStoredWallet().then((wallet) => dispatch(appLaunchedWithLastUsedWallet(wallet)))
+      if (walletExists) getStoredWalletMetadata().then((wallet) => dispatch(appLaunchedWithLastUsedWallet(wallet)))
     })
   }, [dispatch])
 
--- a/apps/mobile-wallet/src/App.tsx
+++ b/apps/mobile-wallet/src/App.tsx
@@ -58,13 +58,13 @@ import { themes } from '~/style/themes'
 SplashScreen.preventAutoHideAsync()
 
 const App = () => {
-  const { data: isStoredWalletDataValid, isLoading: isLoadingStoredWalletDataValidation } = useAsyncData(
+  const { data: isStoredWalletDataValid, isLoading: isValidatingStoredWalletData } = useAsyncData(
     validateAndRepareStoredWalletData
   )
 
   useEffect(() => {
-    if (!isLoadingStoredWalletDataValidation) SplashScreen.hideAsync()
-  }, [isLoadingStoredWalletDataValidation])
+    if (!isValidatingStoredWalletData) SplashScreen.hideAsync()
+  }, [isValidatingStoredWalletData])
 
   const [theme, setTheme] = useState<DefaultTheme>(themes.light)
 
--- a/apps/mobile-wallet/src/App.tsx
+++ b/apps/mobile-wallet/src/App.tsx
@@ -37,9 +37,13 @@ import { DefaultTheme, ThemeProvider } from 'styled-components/native'
 import ToastAnchor from '~/components/toasts/ToastAnchor'
 import { useLocalization } from '~/features/localization/useLocalization'
 import { useAppDispatch, useAppSelector } from '~/hooks/redux'
+import { useAsyncData } from '~/hooks/useAsyncData'
 import useLoadStoredSettings from '~/hooks/useLoadStoredSettings'
 import RootStackNavigation from '~/navigation/RootStackNavigation'
-import { getStoredWalletMetadata, storedWalletExists } from '~/persistent-storage/wallet'
+import {
+  getStoredWalletMetadataWithoutThrowingError,
+  validateAndRepareStoredWalletData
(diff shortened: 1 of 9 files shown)


apps/mobile-wallet/src/analytics.tsx
```diff
@@ -69,6 +69,8 @@ export const sendAnalytics = (params: AnalyticsParams) => {
   } else {
     const { event, props, options } = params
 
+    if (props) props.$ip = ''
+
```
nop33, line 73: Inspired by: https://posthog.com/questions/override-event-ip-address

This shouldn't be necessary since we already do that a few lines above (line 35).

apps/mobile-wallet/ios/Podfile.lock
```diff
@@ -1293,7 +1293,7 @@ PODS:
     - React-Core
   - react-native-background-timer (2.4.1):
     - React-Core
-  - react-native-compat (2.13.3):
+  - react-native-compat (2.15.1):
```
nop33, line 1296: Seems like this was somehow missed after updating the WC deps to 2.15.1

apps/mobile-wallet/src/persistent-storage/wallet.ts
```diff
@@ -87,15 +85,51 @@ export const generateAndStoreWallet = async (
...
+
+    if (walletMetadata) {
+      sendAnalytics({ type: 'error', message: 'Re-created missing wallet metadata' })
+    } else {
+      sendAnalytics({ type: 'error', message: 'Could not re-created missing wallet metadata' })
+    }
+  }
+
+  return walletMetadata
+}
```
nop33, line 127: If we detect a stored mnemonic but no metadata, we regenerate with some default values (`My wallet`, not backed up)

apps/mobile-wallet/src/App.tsx
```diff
@@ -50,7 +55,17 @@ import { selectTransactionUnknownTokenIds } from '~/store/transactions/transacti
 import { appLaunchedWithLastUsedWallet } from '~/store/wallet/walletActions'
 import { themes } from '~/style/themes'
 
+SplashScreen.preventAutoHideAsync()
+
 const App = () => {
+  const { data: isStoredWalletDataValid, isLoading: isLoadingStoredWalletDataValidation } = useAsyncData(
```
mvaivre, line 61: nit: `isValidatingStoredWalletData`

apps/mobile-wallet/src/persistent-storage/wallet.ts
```diff
@@ -56,6 +57,118 @@ const MNEMONIC_V2 = 'wallet-mnemonic-v2'
...
+        // If we have metadata but no mnemonic and no deprecated mnemonic, we should delete the metadata
+        try {
+          await deleteWithReportableError(WALLET_METADATA_STORAGE_KEY)
+        } finally {
+          walletMetadata = await getWalletMetadata(false)
+        }
+
+        if (!walletMetadata) {
+          return true
+        } else {
```
mvaivre, line 156: Nit: I like how you commented the flow above. One more comment here would be nice! 
"This should never happen. Could not delete metadata for some reason and we don't have any mnemonic, we can't unlock"

apps/mobile-wallet/src/persistent-storage/wallet.ts
```diff
@@ -107,8 +240,22 @@ export const getStoredWallet = async (error?: string): Promise<WalletMetadata> =
   return walletMetadata
 }
 
+export const getStoredWalletMetadataWithoutThrowingError = async (): Promise<WalletMetadata | null> => {
+  let walletMetadata = null
+
+  try {
+    walletMetadata = await getWalletMetadata()
```
mvaivre, line 247: Mhh it will actually throw no? 
In the `getWalletMetadata` definition you set  `throwError = true`. So is the parameter is omited, like above, errors will be thrown.

mvaivre: Also, I should probably continue to review to see where this is used, but why not directly use `getWalletMetadata(false)`?

apps/mobile-wallet/src/persistent-storage/wallet.ts
```diff
@@ -87,18 +198,40 @@ export const generateAndStoreWallet = async (
...
+      sendAnalytics({ type: 'error', error, message: 'Could not parse wallet metadata' })
+      if (throwError) throw error
+    }
   }
+
+  return walletMetadata
 }
 
-export const getStoredWallet = async (error?: string): Promise<WalletMetadata> => {
+export const getStoredWalletMetadata = async (error?: string): Promise<WalletMetadata> => {
```
mvaivre, line 234: We need to remove some of these function. It's confusing to me to have both `getStoredWalletMetadata` and `getWalletMetadata`, especially since the former rely directly on the latter.

mvaivre (commented):
Great job 👏  Especially with a melted brain :)

apps/mobile-wallet/src/App.tsx
nop33: Done

apps/mobile-wallet/src/persistent-storage/wallet.ts
nop33: done

apps/mobile-wallet/src/persistent-storage/wallet.ts
nop33: Added a TODO for now

apps/mobile-wallet/src/persistent-storage/wallet.ts
nop33: Fixed

mvaivre (approved):

nop33:
This PR needs to be refactored now that we patched MW last week. I'll work on it. I'll ask for a review again. There area some things we need to discuss about it

apps/mobile-wallet/src/persistent-storage/wallet.ts
```diff
@@ -107,8 +271,12 @@ export const getStoredWallet = async (error?: string): Promise<WalletMetadata> =
   return walletMetadata
 }
 
+export const getStoredWalletMetadataWithoutThrowingError = () => getWalletMetadata(false)
```
mvaivre, line 274: That's better, but why not directly use `getWalletMetadata(false)` at this point?

nop33: because then the code in the component where it is used gets a bit more complicated.

```ts
// now
export const getStoredWalletMetadataWithoutThrowingError = () => getWalletMetadata(false)

const { data: walletMetadata } = useAsyncData(getStoredWalletMetadataWithoutThrowingError)

// after
const { data: walletMetadata } = useAsyncData(useCallback(() => getWalletMetadata(false), []))
```

Not a strong opinion in this, I'll implement the one you prefer the most

PR #809, opened 2024-08-30
Rollback Explorer Backend URL

h0ngcha0:


h0ngcha0:
There are some issues with the explorer backend, rolling back to the AWS instance.

PR #810, opened 2024-09-01
First step on modals optimization refactoring

nop33:
This is an experiment on optimizing our modals architecture. It is directly inspired by Uniswap mobile: https://github.com/Uniswap/interface/tree/main/apps/mobile/src/features/modals

The first step refactors 2 modals: 
- AddressDetailsModal
- CSVExportModal

## Pros
By moving the open/close state of a modal into Redux we decrease the number of rerenders of unrelated components that live close to the modal

## Cons
Our app allows several modals of the same type to be open at the same time. However, Uniswap's architecture does not allow rendering a modal more than onse at the same time.

An example flow that illustrates the cons is:
   1. Overview page
   2. Address list row click (opens AddressDetailsModal)
   3. Transaction list row click (opens TransactionDetailsModal)
   4. Address badge/hash click (opens AddressDetailsModal)

The above flow will cause 2 problems:
1. the AddressDetailsModal of step 2 to rendered at step 4, losing the state of step 2
2. the TransactionDetailsModal will remain on top of the rerendered AddressDetailsModal of step 4

## Ideas for future work

We could try using an array in the Redux state to represent the modals that are currently open.

### DIFF
--- a/apps/desktop-wallet/src/components/IOList.tsx
+++ b/apps/desktop-wallet/src/components/IOList.tsx
@@ -19,16 +19,14 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 import { AddressHash, GENESIS_TIMESTAMP } from '@alephium/shared'
 import { explorer } from '@alephium/web3'
 import _ from 'lodash'
-import { useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import styled from 'styled-components'
 
 import ActionLink from '@/components/ActionLink'
 import AddressBadge from '@/components/AddressBadge'
 import Badge from '@/components/Badge'
-import { useAppSelector } from '@/hooks/redux'
-import AddressDetailsModal from '@/modals/AddressDetailsModal'
-import ModalPortal from '@/modals/ModalPortal'
+import { openModal } from '@/features/modals/modalActions'
+import { useAppDispatch, useAppSelector } from '@/hooks/redux'
 import { selectAddressIds } from '@/storage/addresses/addressesSelectors'
 import { openInWebBrowser } from '@/utils/misc'
 
@@ -56,14 +54,13 @@ const IOList = ({
   const { t } = useTranslation()
   const internalAddressHashes = useAppSelector(selectAddressIds) as AddressHash[]
   const explorerUrl = useAppSelector((state) => state.network.settings.explorerUrl)
-
-  const [selectedAddressHash, setSelectedAddressHash] = useState<AddressHash>()
+  const dispatch = useAppDispatch()
 
   const io = (isOut ? outputs : inputs) as Array<explorer.Output | explorer.Input> | undefined
 
   const handleShowAddress = (addressHash: AddressHash) =>
     internalAddressHashes.includes(addressHash)
-      ? setSelectedAddressHash(addressHash)
+      ? dispatch(openModal({ name: 'AddressDetailsModal', props: { addressHash } }))
       : openInWebBrowser(`${explorerUrl}/addresses/${addressHash}`)
 
   if (io && io.length > 0) {
@@ -102,11 +99,6 @@ const IOList = ({
             addressComponent
           )
         })}
-        <ModalPortal>
-          {selectedAddressHash && (
-            <AddressDetailsModal addressHash={selectedAddressHash} onClose={() => setSelectedAddressHash(undefined)} />
-          )}
-        </ModalPortal>
       </Addresses>
     )
   } else if (timestamp === GENESIS_TIMESTAMP) {
--- a/apps/desktop-wallet/src/modals/AddressDetailsModal.tsx
+++ b/apps/desktop-wallet/src/modals/AddressDetailsModal.tsx
@@ -32,33 +32,45 @@ import HashEllipsed from '@/components/HashEllipsed'
 import QRCode from '@/components/QRCode'
 import TransactionList from '@/components/TransactionList'
 import AssetsTabs from '@/features/assetsLists/AssetsTabs'
-import { useAppSelector } from '@/hooks/redux'
-import CSVExportModal from '@/modals/CSVExportModal'
-import ModalPortal from '@/modals/ModalPortal'
+import { closeModal, openModal } from '@/features/modals/modalActions'
+import ModalNames from '@/features/modals/modalNames'
+import { selectModalState } from '@/features/modals/modalSelectors'
+import { useAppDispatch, useAppSelector } from '@/hooks/redux'
 import SideModal from '@/modals/SideModal'
 import AmountsOverviewPanel from '@/pages/UnlockedWallet/OverviewPage/AmountsOverviewPanel'
 import { selectAddressByHash } from '@/storage/addresses/addressesSelectors'
 import { openInWebBrowser } from '@/utils/misc'
 
-interface AddressDetailsModalProps {
+export interface AddressDetailsModalProps {
   addressHash: AddressHash
-  onClose: () => void
 }
 
-const AddressDetailsModal = ({ addressHash, onClose }: AddressDetailsModalProps) => {
+const AddressDetailsModalWrapper = () => {
+  const { props } = useAppSelector((s) => selectModalState(s, ModalNames.AddressDetailsModal))
+
+  return props ? <AddressDetailsModal addressHash={props.addressHash} /> : null
+}
+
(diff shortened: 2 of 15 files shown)


apps/desktop-wallet/src/modals/AddressDetailsModal.tsx
```diff
@@ -32,30 +32,33 @@ import HashEllipsed from '@/components/HashEllipsed'
...
 import { openInWebBrowser } from '@/utils/misc'
 
-interface AddressDetailsModalProps {
+export interface AddressDetailsModalProps {
   addressHash: AddressHash
-  onClose: () => void
 }
 
-const AddressDetailsModal = ({ addressHash, onClose }: AddressDetailsModalProps) => {
+const AddressDetailsModal = memo(({ id, addressHash }: ModalBaseProp & AddressDetailsModalProps) => {
```
nop33, line 47: > `memo` lets you skip re-rendering a component when its props are unchanged.
https://react.dev/reference/react/memo

We need to make sure that **any** modal can render on top of **any other** modal. We also need to make sure that the same modal can appear **multiple times** in the stack of modals. To achieve this, as seen later, in the `AppModals.tsx` component, we are loop over an array of opened modals to render them.

`memo` is used to wrap every (refactored) modal component. The purpose of it is to avoid re-rendering a modal that is already open **when a new open opens** (since the open arrays modal changes when a new modal opens, each modal is a new instance - we want to avoid re-rendering modal instance components when their props haven't changed).

apps/desktop-wallet/src/modals/AddressDetailsModal.tsx
```diff
@@ -32,30 +32,33 @@ import HashEllipsed from '@/components/HashEllipsed'
...
   const address = useAppSelector((s) => selectAddressByHash(s, addressHash))
   const explorerUrl = useAppSelector((s) => s.network.settings.explorerUrl)
+  const dispatch = useAppDispatch()
 
-  const [isCSVExportModalOpen, setIsCSVExportModalOpen] = useState(false)
   const [showChart, setShowChart] = useState(false)
 
   if (!address) return null
 
+  const onClose = () => dispatch(closeModal({ id }))
```
nop33, line 58: Once all modals have been refactored, there will be no need to declare this function in every modal. It can be moved to the `ModalContainer`. But until then, to keep the code functional, we need to provide an `onClose` function.

apps/desktop-wallet/src/components/IOList.tsx
```diff
@@ -56,14 +54,13 @@ const IOList = ({
...
-
-  const [selectedAddressHash, setSelectedAddressHash] = useState<AddressHash>()
+  const dispatch = useAppDispatch()
 
   const io = (isOut ? outputs : inputs) as Array<explorer.Output | explorer.Input> | undefined
 
   const handleShowAddress = (addressHash: AddressHash) =>
     internalAddressHashes.includes(addressHash)
-      ? setSelectedAddressHash(addressHash)
+      ? dispatch(openModal({ name: 'AddressDetailsModal', props: { addressHash } }))
```
mvaivre, line 63: I'm testing this and I see that it's properly typed! Really cool.

apps/desktop-wallet/src/components/TransactionList.tsx
```diff
@@ -144,85 +146,75 @@ const TransactionList = ({
...
-    </>
+        </TableRow>
+      ))}
+      {displayedConfirmedTxs.map((tx) => (
+        <TableRow
+          key={`${tx.hash}-${tx.address.hash}`}
+          role="row"
+          tabIndex={0}
+          onClick={() => openTransactionDetailsModal(tx.hash, tx.address.hash)}
+          onKeyDown={() => openTransactionDetailsModal(tx.hash, tx.address.hash)}
```
mvaivre, line 189: Should we filter the key so that it's only opening the modal when pushing enter?

apps/desktop-wallet/src/modals/AddressDetailsModal.tsx
mvaivre: cf. my idea to create a `useModal` hook which could return a `closeModal` function.

apps/desktop-wallet/src/modals/AppModals.tsx
```diff
@@ -0,0 +1,49 @@
...
+          case 'AddressDetailsModal':
+            return <AddressDetailsModal id={modal.id} key={modal.id} {...modal.params.props} />
+          case 'CSVExportModal':
+            return <CSVExportModal id={modal.id} key={modal.id} {...modal.params.props} />
+          case 'NFTDetailsModal':
+            return <NFTDetailsModal id={modal.id} key={modal.id} {...modal.params.props} />
+          case 'TransactionDetailsModal':
+            return <TransactionDetailsModal id={modal.id} key={modal.id} {...modal.params.props} />
+        }
+      })}
```
mvaivre, line 44: It feels like we could DRY this up with a loop, also to make the code more maintainable, but I'm yet to find a nice way while keeping solid typing (didn't spend more than 5min on it though :))

nop33: I spent more than 5 minutes and this is the best I came up with! I am of course open to improvement!

apps/desktop-wallet/src/components/IOList.tsx
```diff
@@ -56,14 +54,13 @@ const IOList = ({
   const { t } = useTranslation()
   const internalAddressHashes = useAppSelector(selectAddressIds) as AddressHash[]
   const explorerUrl = useAppSelector((state) => state.network.settings.explorerUrl)
-
-  const [selectedAddressHash, setSelectedAddressHash] = useState<AddressHash>()
+  const dispatch = useAppDispatch()
```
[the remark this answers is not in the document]
nop33: It's not possible to separate the modal name from the modal props because then we'll be able to pass a modal name that does not correspond to the right modal props. They need to be kept in the same object. In your example they are separated into 2 vars in the arguments of `openModal`. I'll look into it.

apps/desktop-wallet/src/modals/AddressDetailsModal.tsx
nop33: I summed up my findings in the other comment: https://github.com/alephium/alephium-frontend/pull/810#discussion_r1740566632

apps/desktop-wallet/src/components/TransactionList.tsx
nop33: Fixed

apps/desktop-wallet/src/components/IOList.tsx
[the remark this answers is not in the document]
mvaivre: Honestly I like `useModal` quite a bit because you can clearly see right at the top of the component that it can trigger the opening of a modal. A bit like when you see `useTranslation`, `useTheme`, it's easy to spot the side effects and dependencies this way.

apps/desktop-wallet/src/modals/AppModals.tsx
mvaivre: Haha! LGTM then.

apps/desktop-wallet/src/components/IOList.tsx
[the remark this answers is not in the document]
nop33: I hear you, and I see the merit of your proposal, but I'd like to stick to my proposal and arguments, I am not convinced that it's worth creating `useModal`. I don't see enough benefits. Uniswap also doesn't have it. I'd like to keep it simple with just a `dispatch`.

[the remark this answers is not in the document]
nop33: I experimented a bit more with your proposal, specifically I tried creating `useModalOpen`. I discovered another reason why I don't like it. By using the `dispatch` method we can define our `open[...]Modal` functions **after** the `return null` line in many components, saving us some more memory from declaring unnecessary functions. The `useModal` being a hook needs to always be declared above conditions.

Worth mentioning that all the above are nitpicking IMO :) We are talking about tiny optimizations and tiny improvements.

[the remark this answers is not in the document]
mvaivre: Works for me. I've kept the `dispatch` in MW too ;)

mvaivre (approved):

PR #815, opened 2024-09-02
Modals optimization refactoring (part 2)

nop33:
Continuing the modal refactoring with the following modals:

```
TransactionDetailsModal
AddressOptionsModal
SettingsModal
ReceiveModal
WalletConnectModal
SecretPhraseModal
WalletQRCodeExportModal
EditWalletNameModal
NotificationsModal
AdvancedOperationsSideModal
NewAddressModal
ContactFormModal
```

### DIFF
--- a/apps/desktop-wallet/src/components/Buttons/ShortcutButtons.tsx
+++ b/apps/desktop-wallet/src/components/Buttons/ShortcutButtons.tsx
@@ -24,63 +24,66 @@ import styled, { useTheme } from 'styled-components'
 
 import Button from '@/components/Button'
 import useAnalytics from '@/features/analytics/useAnalytics'
-import { useAppSelector } from '@/hooks/redux'
+import { openModal } from '@/features/modals/modalActions'
+import { useAppDispatch, useAppSelector } from '@/hooks/redux'
 import useWalletLock from '@/hooks/useWalletLock'
-import AddressOptionsModal from '@/modals/AddressOptionsModal'
 import ModalPortal from '@/modals/ModalPortal'
 import ReceiveModal from '@/modals/ReceiveModal'
 import SendModalTransfer from '@/modals/SendModals/Transfer'
 import SettingsModal from '@/modals/SettingsModal'
 import { selectAddressByHash, selectDefaultAddress } from '@/storage/addresses/addressesSelectors'
 
-interface ShortcutButtonsProps {
+interface ShortcutButtonBaseProps {
   analyticsOrigin: string
-  send?: boolean
-  receive?: boolean
-  lock?: boolean
-  walletSettings?: boolean
-  addressSettings?: boolean
-  addressHash?: AddressHash
   highlight?: boolean
   solidBackground?: boolean
 }
 
-const ShortcutButtons = ({
-  analyticsOrigin,
-  send,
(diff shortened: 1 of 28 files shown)


apps/desktop-wallet/src/modals/AddressOptionsModal.tsx
```diff
@@ -37,19 +39,19 @@ import { saveAddressSettings } from '@/storage/addresses/addressesStorageUtils'
...
 import { getRandomLabelColor } from '@/utils/colors'
 
-interface AddressOptionsModalProps {
+export interface AddressOptionsModalProps {
   addressHash: AddressHash
-  onClose: () => void
 }
 
-const AddressOptionsModal = ({ addressHash, onClose }: AddressOptionsModalProps) => {
+const AddressOptionsModal = memo(({ id, addressHash }: ModalBaseProp & AddressOptionsModalProps) => {
```
mvaivre, line 46: About the use of memoization: I've experimented with a runtime check giving a warning if one modal isn't memoized, and it seems to work :) 
I wonder what you think about it. Yes is uses the `Children` api, but that's only to output a warning, so no risk there. 

Check out [this commit](https://github.com/alephium/alephium-frontend/commit/0b08a802661a15e4bdb72e0c9a967187de1b7270) 👀

mvaivre: It could also be added to `ModalContainer` actually, avoiding the use of `Children`.

apps/desktop-wallet/src/components/Buttons/ShortcutButtons.tsx
```diff
@@ -24,161 +24,173 @@ import styled, { useTheme } from 'styled-components'
...
+      highlight={highlight}
+    >
+      <ButtonText>{t('Receive')}</ButtonText>
+    </ShortcutButton>
+  )
+}
 
-  const handleWalletSettingsClick = () => {
-    setIsSettingsModalOpen(true)
+const SendButton = ({ addressHash, analyticsOrigin, solidBackground, highlight }: ShortcutButtonsGroupAddressProps) => {
```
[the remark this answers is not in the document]
nop33: It's intended that it still relies on the old system, but it's a bug that it doesn't open. I'll look into it, thanks!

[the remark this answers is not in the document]
nop33: I tested all 4 places where we open the send modal and it works fine:
- Overview page
- Transfers page
- Address details modal
- Addresses page -> contacts tab

apps/desktop-wallet/src/modals/AddressOptionsModal.tsx
nop33: I [implemented your proposal](https://github.com/alephium/alephium-frontend/pull/815/commits/16c343820d1d6dba2c39a925ccbd68d07998bb18) in this commit and added you as a co-author! Good job! 🤝 

I tweaked it a little bit to:
- Only run when needed
- Share it through out `@alephium/shared-react` package

mvaivre (approved):
LGTM! Great work!

PR #818, opened 2024-09-03
Navigate to active address scan screen on metadata restore

nop33:
Given the troubles that out internal testing team had today with scanning for active addresses after walelt metadata restore I decided to change the flow and reuse part of the wallet generation flow on metadata restore with some nicer, friendlier messages.

## Before

<details><summary>Click to expand screenshot</summary>

</details>

## After

| Step 1 | Step 2 | Step 3 | Step 4
| image | image | image | image |

### DIFF
--- a/apps/mobile-wallet/src/App.tsx
+++ b/apps/mobile-wallet/src/App.tsx
@@ -90,8 +90,8 @@ const App = () => {
 const useShowAppContentAfterValidatingStoredWalletData = () => {
   const [state, setState] = useState({ showAppContent: false, wasMetadataRestored: false })
 
-  const onUserConfirm = useCallback(() => {
-    setState({ showAppContent: true, wasMetadataRestored: true })
+  const onUserConfirm = useCallback((userChoseYes: boolean) => {
+    setState({ showAppContent: true, wasMetadataRestored: userChoseYes })
     store.dispatch(metadataRestored())
   }, [])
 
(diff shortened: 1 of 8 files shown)


apps/mobile-wallet/src/App.tsx
```diff
@@ -53,12 +53,13 @@ import {
...
 import { selectTransactionUnknownTokenIds } from '~/store/transactions/transactionSelectors'
 import { appLaunchedWithLastUsedWallet } from '~/store/wallet/walletActions'
+import { metadataRestored } from '~/store/wallet/walletSlice'
 import { themes } from '~/style/themes'
 
 SplashScreen.preventAutoHideAsync()
 
 const App = () => {
-  const showAppContent = useShowAppContentAfterValidatingStoredWalletData()
+  const { showAppContent, wasMetadataRestored } = useShowAppContentAfterValidatingStoredWalletData()
```
mvaivre, line 62: It's actually the mnemonic - or the "wallet" - that is restored, no? Metadata are gone, can't be restored.

nop33: haha interesting, it depends how you see it! I see it more on the perspective that metadata was not there, so we restored them/recreated them. The mnemonic was always there so no need to restore anything. Since it's strictly a programming variable and not visible to the user, I'd prefer to leave the word "metadata" in it, to help out future selves understand what's going on here with a glace. WDYT?

mvaivre (approved):
Okay, let's create a new internal testing version with this! 👏

nop33:
> Okay, let's create a new internal testing version with this! 👏

that was already done yesterday, as per my message on Slack!

PR #823, opened 2024-09-05
Refactor all send modals (part 4)

nop33:
@mvaivre  Phew, that was a tough one... I'm opening a PR just for the send modals because it's already big enough.


apps/desktop-wallet/src/modals/SendModals/CallContract/BuildTxModalContent.tsx
```diff
@@ -32,13 +32,14 @@ import useGasSettings from '@/hooks/useGasSettings'
 import useStateObject from '@/hooks/useStateObject'
 import AssetAmountsInput from '@/modals/SendModals/AssetAmountsInput'
 import GasSettings from '@/modals/SendModals/GasSettings'
+import { CallContractTxModalData } from '@/modals/SendModals/sendTypes'
```
mvaivre, line 35: Mhh, what about having this and the modals in `features/send`?

apps/desktop-wallet/src/modals/SettingsModal/DevToolsSettingsSection.tsx
```diff
@@ -108,6 +105,12 @@ const DevToolsSettingsSection = () => {
     sendAnalytics({ event: 'Requested testnet tokens' })
   }
 
+  const openCallContractModal = () =>
+    dispatch(openModal({ name: 'SendModalCallContract', props: { initialTxData: { fromAddress: defaultAddress } } }))
```
mvaivre, line 109: Should we keep "Modal" as a suffix to help with understanding the file structure better in a glance? Like `SendCallContractModal` 🤔

apps/desktop-wallet/src/storage/store.ts
```diff
@@ -45,7 +45,14 @@ export const store = configureStore({
     [modalSlice.name]: modalSlice.reducer
   },
   middleware: (getDefaultMiddleware) =>
-    getDefaultMiddleware().concat(settingsListenerMiddleware.middleware).concat(networkListenerMiddleware.middleware)
+    getDefaultMiddleware({
+      serializableCheck: {
```
mvaivre, line 49: We could maybe add a little comment here, to explain that this is because we need to use functions as props.

mvaivre (commented):
Wow, this was a lot of work. First pass on the code looks good, I'll focus on testing with WalletConnect.

apps/desktop-wallet/src/modals/SendModals/CallContract/BuildTxModalContent.tsx
nop33: YES! I just didn't want to do it yes so that you can inspect the diff! Now that you did a first pass, I'll make the move!

apps/desktop-wallet/src/modals/SettingsModal/DevToolsSettingsSection.tsx
nop33: You're absolutely right

apps/desktop-wallet/src/storage/store.ts
nop33: ok!

nop33: Done!

apps/desktop-wallet/src/modals/SettingsModal/DevToolsSettingsSection.tsx
nop33: Done!

apps/desktop-wallet/src/modals/SendModals/CallContract/BuildTxModalContent.tsx
nop33: Done!

mvaivre (approved):
🚀

PR #827, opened 2024-09-09
Refactor all remaining modals (step 6)

nop33:


apps/desktop-wallet/src/api/apiDataHooks/useAddressesAlphBalancesTotal.ts
```diff
@@ -54,7 +54,7 @@ const combine = (results: UseQueryResult<AddressAlphBalancesQueryFnData>[]): Add
     (totalBalances, { data }) => {
       totalBalances.totalBalance += data ? data.alphBalances.totalBalance : BigInt(0)
       totalBalances.lockedBalance += data ? data.alphBalances.lockedBalance : BigInt(0)
-      totalBalances.availableBalance = data ? data.alphBalances.availableBalance : BigInt(0)
+      totalBalances.availableBalance += data ? data.alphBalances.availableBalance : BigInt(0)
```
mvaivre, line 57: :)

apps/desktop-wallet/src/components/PasswordConfirmation.tsx
```diff
@@ -28,24 +28,25 @@ import { useAppDispatch, useAppSelector } from '@/hooks/redux'
...
+const PasswordConfirmation = ({
   text,
   buttonText,
   onCorrectPasswordEntered,
   walletId,
   isSubmitDisabled = false,
   highlightButton = false,
   children
-}) => {
+}: PasswordConfirmationProps) => {
```
mvaivre, line 49: 👍 😃

apps/desktop-wallet/src/modals/AddressSweepModal.tsx
```diff
@@ -45,178 +43,178 @@ import {
...
-        const message = 'Error while building transaction'
+      buildTransactions()
+    }, [dispatch, sendAnalytics, sweepAddresses.from, sweepAddresses.to, t])
 
-        dispatch(transactionBuildFailed(getHumanReadableError(e, t(message))))
-        sendAnalytics({ type: 'error', message })
-      }
-      setIsLoading(false)
-    }
+    const onAddressChange = useCallback(
```
mvaivre, line 96: Very Nit: in most places we prefix our callbacks with `handle`, like `handleAddressChange`, so that in the component we have `onAddressChange={handleAddressChange}`. I found it pretty elegant but we're not really consistent as we can see in this file. Do you think we should enforce this or we don't care?

apps/desktop-wallet/src/modals/WalletRemovalModal.tsx
```diff
@@ -17,27 +17,47 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+const WalletRemovalModal = memo(({ id, walletId, walletName }: ModalBaseProp & WalletRemovalModalProps) => {
   const { t } = useTranslation()
   const theme = useTheme()
+  const dispatch = useAppDispatch()
+  const { sendAnalytics } = useAnalytics()
+  const activeWalletId = useAppSelector((s) => s.activeWallet.id)
+
+  const removeWallet = () => {
+    walletStorage.delete(walletId)
+    addressMetadataStorage.delete(walletId)
```
mvaivre, line 51: I find it weird to not have a utility function dedicated to this.

mvaivre: Also, will this throw if it fails?

mvaivre (commented):
👏

apps/desktop-wallet/src/modals/AddressSweepModal.tsx
nop33: Thanks for catching this, I don't think it's too important to enforce it, but I'll do a cleanup in this PR and be more careful in the future. OTOH however, I like it when the callback has the same name as the prop because then it doesn't need to be written twice for example in a context value. I'll mention your name next time I find something "discussable" ;)

nop33: What I don't like with `handle...` is to call it programmatically. I like it when I see it as a callback to buttons like
```tsx
<Button title="Generate" onClick="handleGenerateClick" />
```

but I don't like it when we need to call it as:

```tsx
const foo = () => {
  // doing some stuff here, blah blah blah
  // ...
  handleGenerateClick()
  // ...
}

<Button title="Generate" onClick="handleGenerateClick" />
}
```

In the latter case I prefer to see:

```tsx
const foo = () => {
  // doing some stuff here, blah blah blah
  // ...
  onGenerateClick()
  // ...
}

<Button title="Generate" onClick="onGenerateClick" />
}
```

WDYT?

Same goes for `onClose`

nop33: (Rename was done in 212f3b421)

apps/desktop-wallet/src/modals/WalletRemovalModal.tsx
nop33: `localStorage.removeItem` doesn't throw, so no, it won't thrown. MAnaging storage is so much simpler on a browser env than on a mobile device

nop33: Also, about the utility, I can create it with these 2 lines of code, but we are not using this code anywhere else.

apps/desktop-wallet/src/modals/AddressSweepModal.tsx
mvaivre: Yep I agree with you fully.

apps/desktop-wallet/src/modals/WalletRemovalModal.tsx
nop33: @mvaivre Do you want me to create this utility? I find it unnecessary. Lemme know

mvaivre: Don't you think it could be great, just in case we want to test it at some point?
If you don't feel the need for it I'm ok with that as well.

BTW; I'm gonna approve this PR already ;)

mvaivre (approved):
Code looks good! 👏  I haven't thoroughly tested yet, I plan to do it once all the tanstack stuff is reviewed and merged.

apps/desktop-wallet/src/modals/WalletRemovalModal.tsx
nop33: Added to the next release: https://github.com/alephium/alephium-frontend/issues/831

Issue #234, opened 2024-01-04
Refactor the `Amount` component


mvaivre:
The Amount component is due for a cleanup. We need to make its props more strict.

We can probably do better with clever types. Just an idea :

- Expect a mandatory amount type prop to be either fiat or token
- If the type is token, then the decimals field is mandatory.

cc @nop33 to be discussed.

nop33:
I 100% agree.

PR #830, opened 2024-09-10
WIP: Tanstack

nop33:


apps/desktop-wallet/src/api/apiDataHooks/useToken.ts
nop33, line 1: aaaaand we are now full circle to something that is the closest to what you originally proposed. I now start to understand why batshit will be necessary. I tried for a long time to do a top->down approach with Tanstack (first fetching all address data and deriving everything from there) but I struggled a lot with re-renders and dependencies so now I am experimenting with a bottom->up approach. This, however, will require batshit because the number of requests increases by A LOT. TBD

apps/desktop-wallet/src/api/utils.ts
nop33, line 1: The changes in this file also happen to close https://github.com/alephium/alephium-frontend/issues/234

apps/desktop-wallet/src/components/Amount.tsx
nop33, line 1: The changes in this file also happen to close https://github.com/alephium/alephium-frontend/issues/234

apps/desktop-wallet/src/api/apiDataHooks/useToken.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+export default useToken
+
+export const isListedFT = (token: Token['data']): token is ListedFT => (token as ListedFT).logoURI !== undefined
+
+export const isFT = (token: Token['data']): token is ListedFT | UnlistedFT =>
+  (token as ListedFT | UnlistedFT).symbol !== undefined
+
+export const isNFT = (token: Token['data']): token is NFT => (token as NFT).nftIndex !== undefined
```
nop33, line 92: I am playing around with type guards similar to the ones you created. I will probably and eventually clean up my other implementation of the `type` field, but I am not sure yet. I want to experiment in the next PR.

apps/desktop-wallet/src/api/apiDataHooks/useToken.ts
```diff
@@ -0,0 +1,92 @@
...
+import useFTList from '@/api/apiDataHooks/useFTList'
+import { fungibleTokenMetadataQuery, nftDataQuery, nftMetadataQuery, tokenTypeQuery } from '@/api/queries/tokenQueries'
+import { ListedFT, NonStandardToken, TokenId, UnlistedFT } from '@/types/tokens'
+
+type Token = {
+  data: ListedFT | UnlistedFT | NFT | NonStandardToken
+  isLoading: boolean
+}
+
+const useToken = (id: TokenId): Token => {
```
nop33, line 33: So, I think you will be aligned with what's happening in this hook.
1. We get the fungible token list. If our token is there, all following queries will be skipped and we'll return a structurally shared object
2. If the token wasn't there (and only then, thanks to `skip`) we query for its type. This will determine which queries will be executed in the next component re-render.
3. In the next re-render, since the token is not a listed one, we have its type, so we query for its metadata.
4. Finally we fetch for the nft data if the type is NFT and we have received the metadata.

Some notes:
1. Any component that uses this hook will render at least 2 times on app launch
   1. 1st time the `isLoadingFtList` will be `true` and the hook will not run any other queries
   2. 2nd time the token list will be available so if the token is part of it, no more re-renders.
2. If the id is of an NFT, the component that uses this hook will be re-render at least 5 times.
```
   1. fTList: undefined,         isLoadingFtList: true
   2. fTList: [...],             isLoadingFtList: false,     tokenType: undefined,   isLoadingTokenType: true
   3. tokenType: 'non-fungible', isLoadingTokenType: false,  nftMetadata: undefined, isLoadingNftMetadata: true
   4. nftMetadata: {...},        isLoadingNftMetadata: false, nftData: undefined,    isLoadingNftData: true
   5. nftData: {...},            isLoadingNftData: false
```

apps/desktop-wallet/src/components/Amount.tsx
```diff
@@ -16,111 +16,138 @@ You should have received a copy of the GNU Lesser General Public License
...
+  const { className, color, value, highlight, tabIndex, showPlusMinus } = props
 
-    if (fadeDecimals && ['K', 'M', 'B', 'T'].some((char) => amount.endsWith(char))) {
-      quantitySymbol = amount.slice(-1)
-      amount = amount.slice(0, -1)
-    }
-  }
-
-  const [integralPart, fractionalPart] = amount.split('.')
+  const toggleDispatchMode = () => discreetMode && dispatch(discreetModeToggled())
```
mvaivre, line 59: `toggleDispatchMode`? 🤔 😆

apps/desktop-wallet/src/components/AssetLogo.tsx
```diff
@@ -16,40 +16,49 @@ You should have received a copy of the GNU Lesser General Public License
...
-  </div>
-)
+const AssetLogo = memo(({ tokenId, size, className }: AssetLogoProps) => {
+  const { data: token } = useToken(tokenId)
+  const image = isListedFT(token) ? token.logoURI : isNFT(token) ? token.image : undefined
+  const name = isFT(token) || isNFT(token) ? token.name : undefined
+
+  return (
+    <AssetLogoStyled className={className} size={size} isSquare={isNFT(token)}>
+      {image?.endsWith('.mp4') ? (
```
mvaivre, line 40: This reminds me that we should port the more generic and efficient solution from the explorer, the one creating thumbnails and compatible with more video formats.

https://github.com/alephium/alephium-frontend/blob/next/apps/explorer/src/components/VideoThumbnail.tsx

apps/desktop-wallet/src/components/AssetLogo.tsx
```diff
@@ -16,40 +16,49 @@ You should have received a copy of the GNU Lesser General Public License
...
-    ) : assetName ? (
-      <Initials size={size}>{assetName.slice(0, 2)}</Initials>
-    ) : (
-      <HelpCircle size={size} />
-    )}
-  </div>
-)
+const AssetLogo = memo(({ tokenId, size, className }: AssetLogoProps) => {
+  const { data: token } = useToken(tokenId)
+  const image = isListedFT(token) ? token.logoURI : isNFT(token) ? token.image : undefined
```
mvaivre, line 35: I like that.

apps/desktop-wallet/src/components/Inputs/SelectOptionToken.tsx
```diff
@@ -31,25 +31,22 @@ type SelectOptionTokenProps = TokenDisplay & {
...
 }
 
-const SelectOptionToken = ({ isSelected, className, ...token }: SelectOptionTokenProps) => {
-  const { image, name, symbol, amount, decimals } = getTokenDisplayData(token)
+const SelectOptionToken = ({ isSelected, className, id, ...tokenProps }: SelectOptionTokenProps) => {
+  const { data: token } = useToken(id)
+
+  const name = isFT(token) || isNFT(token) ? token.name : undefined
+  const symbol = isFT(token) ? token.symbol : undefined
+  const amount = tokenProps.type !== 'NFT' ? tokenProps.availableBalance : undefined
```
mvaivre, line 39: It's a funny mix between using type guards and using the `.type` property :)

mvaivre (commented):
Very clean, I love it! ✨

apps/desktop-wallet/src/components/Inputs/SelectOptionToken.tsx
nop33: Yeah, it's until I fully migrate to one of the 2, it's temp

apps/desktop-wallet/src/api/apiDataHooks/useToken.ts
```diff
@@ -0,0 +1,92 @@
...
+            ...nftMetadata,
+            ...nftData
+          } as NFT)
+        : undefined,
+    [nftData, nftMetadata]
+  )
+
+  return {
+    data: listedFT
+      ? (listedFT as ListedFT)
```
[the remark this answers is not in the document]
nop33: `ListedFT` is the same as `TokenInfo` (from `@alephium/token-list`). So the casting is simply a rename here.

apps/desktop-wallet/src/api/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+      )))
+
+export const convertDecimalsToNumber = (token: FungibleTokenMetadata) => {
+  const parsedDecimals = parseInt(token.decimals)
+
+  return {
+    ...token,
+    decimals: Number.isInteger(parsedDecimals) ? parsedDecimals : 0
+  }
+}
```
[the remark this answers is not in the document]
nop33: For sure, many many things could be shared and should be properly tested.

apps/desktop-wallet/src/components/AssetLogo.tsx
nop33: Wanna create an issue to not forget about it?

apps/desktop-wallet/src/components/Amount.tsx
nop33: Fixed in 639256568

mvaivre (approved):
LGTM! onto the next one 🚀

PR #832, opened 2024-09-11
Refactor transaction rows and modals

nop33:
Around 500 LOC are just the licence headers 😅 

This PR massively improves the performance of places where transactions are displayed. It also moves code into features and splits huge components into smaller ones for more efficient re-rendering. I'm sorry for the huge diff, it's about 1500 LOC.

Important learning: move state as close as possible to the “leaf” components (the ones at the end of the component tree). That way, when the state changes, only them will update and re-render.

Another learning is isolate “leaf” components so that state updates of one will not affect another. For example: if we have a component that renders both FT and NFT, it’s better to split it in 2 sub-components so that the FT subcomponent one won’t re-render when the NFT subcomponent updates once the NFT images load (which can take a while)

Here's a preview of the file structure:

image


apps/desktop-wallet/src/components/TransactionRow.tsx
nop33, line 1: I would prefer to split this file in multiple files, but to avoid making the PR look more scare I left it like this

apps/desktop-wallet/src/components/DataList.tsx
```diff
@@ -19,23 +19,19 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
-
 interface DataListProps {
   children: ReactNode
   title?: string
 }
 
-const DataList = ({ children, title }: DataListProps) => {
-  validateChildrenType({ children, childType: DataListRow, parentName: 'DataList' })
+const DataList = ({ children, title }: DataListProps) => (
+  // validateChildrenType({ children, childType: DataListRow, parentName: 'DataList' })
```
nop33, line 28: I could not find a way to create a component in a separate file that is a `DataList.Row` without having this validation screening in the console. Any help is appreciated.

apps/desktop-wallet/src/components/TransactionList.tsx
```diff
@@ -55,7 +56,7 @@ interface TransactionListProps {
   limit?: number
   compact?: boolean
   hideHeader?: boolean
-  hideFromColumn?: boolean
+  isInAddressDetailsModal?: boolean
```
nop33, line 59: This rename made code readability so much easier...

apps/desktop-wallet/src/features/transactionsDisplay/transactionRow/OtherAmounts.tsx
nop33, line 1: This is a new component that I created because I found the way we displayed delta amounts lacking. See before and after

## Before
image



## After

image

apps/desktop-wallet/src/utils/transactions.ts
nop33, line 1: Here I am braking down the `getTransactionalInfo` to smaller functions and hooks so that it can be used by the smaller components that I created. It's better for testing as well.

apps/desktop-wallet/src/features/transactionsDisplay/transactionDetailsModal/TransactionDetailsModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+import { ModalBaseProp } from '@/features/modals/modalTypes'
+import AddressesDataRows from '@/features/transactionsDisplay/transactionDetailsModal/AddressesDataRows'
+import DirectionalInfo from '@/features/transactionsDisplay/transactionDetailsModal/DirectionalInfo'
+import FTAmounts from '@/features/transactionsDisplay/transactionDetailsModal/FTAmounts'
+import FTsDataListRow from '@/features/transactionsDisplay/transactionDetailsModal/FTsDataListRow'
+import GasUTXOsExpandableSection from '@/features/transactionsDisplay/transactionDetailsModal/GasUTXOsExpandableSection'
+import LockTimeDataListRow from '@/features/transactionsDisplay/transactionDetailsModal/LockTimeDataListRow'
+import NFTsDataListRow from '@/features/transactionsDisplay/transactionDetailsModal/NFTsDataListRow'
+import NSTsDataListRow from '@/features/transactionsDisplay/transactionDetailsModal/NSTsDataListRow'
+import TransactionType from '@/features/transactionsDisplay/transactionDetailsModal/TransactionType'
```
mvaivre, line 40: I wonder if we should maybe create an `index.tsx` file in those feature folders, to allow for importing things in one line instead of this 🤔 WDYT?

apps/desktop-wallet/src/features/transactionsDisplay/transactionDetailsModal/types.ts
```diff
@@ -0,0 +1,25 @@
...
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { AddressHash } from '@alephium/shared'
+import { Transaction } from '@alephium/web3/dist/src/api/api-explorer'
+
+export interface TransactionDetailsModalSectionProps {
```
mvaivre, line 22: I'm not so sure about having the props separated from the main component 🤔 
Maybe I'm wrong, but I seem to remember you saying that you liked having types close to where they are used.

apps/desktop-wallet/src/features/transactionsDisplay/transactionRow/OtherAmounts.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+import { useMemo } from 'react'
+import { useTranslation } from 'react-i18next'
+
+import Amount from '@/components/Amount'
+import { TransactionRowProps } from '@/features/transactionsDisplay/transactionRow/types'
+import useTransactionTokens from '@/features/transactionsDisplay/useTransactionTokens'
+import { useTransactionInfoType } from '@/utils/transactions'
+
+interface TransactionOtherTokenAmountsProps extends TransactionRowProps {
```
mvaivre, line 27: Linked to my previous comment : if we indeed go for a dedicated types.ts file, should this be declared there?

apps/desktop-wallet/src/utils/transactions.ts
```diff
@@ -124,73 +129,166 @@ export const directionOptions: {
...
+          return 'in'
+        }
+      }
+    }
   }
+}
 
-  const tokenAssets = [...tokens.map((token) => ({ ...token, ...fungibleTokens[token.id] }))]
-  const assets = amount !== undefined ? [{ ...ALPH, amount }, ...tokenAssets] : tokenAssets
+export const useTransactionLockTime = (tx: Transaction | PendingTransaction) =>
```
mvaivre, line 228: I feel like many of the hooks define in this file could be moved to the `transactionsDisplay` folder.. WDYT?

mvaivre (commented):
Incredible work! 👏 

I'm getting a bunch of error in the console:
image
image

apps/desktop-wallet/src/api/apiDataHooks/useNFT.ts
```diff
@@ -0,0 +1,48 @@
...
+  )
+
+  return {
+    data: useMemo(
+      () =>
+        !!nftMetadata && !!nftData
+          ? ({
+              ...nftMetadata,
+              ...nftData
+            } as NFT)
```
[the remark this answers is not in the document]
nop33: It's to improve the dev experience!

apps/desktop-wallet/src/api/apiDataHooks/useToken.ts
```diff
@@ -78,7 +57,7 @@ const useToken = (id: TokenId): Token => {
       : unlistedFT
         ? (unlistedFT as UnlistedFT)
         : nft || ({ id } as NonStandardToken),
```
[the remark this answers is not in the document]
nop33: Same answer as above!

apps/desktop-wallet/src/api/apiDataHooks/utils.ts
```diff
@@ -33,3 +33,11 @@ export const flatMapCombine = <R>(results: UseQueryResult<R | R[], Error>[]) =>
   data: results.flatMap(({ data }) => data).filter(isDefined),
   ...combineIsLoading(results)
 })
+
+export const mapCombineDefined = <R>(results: UseQueryResult<R, Error>[]) => ({
```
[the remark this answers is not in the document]
nop33: The difference is that this actually returns only the defined elements. The other one seems that it doesn't.

apps/desktop-wallet/src/api/queries/tokenQueries.ts
```diff
@@ -63,7 +63,7 @@ export const fungibleTokenMetadataQuery = ({ id, skip }: TokenQueryProps) =>
       ? async () => {
           const tokenMetadata = await fungibleTokenMetadataBatchFetcher.fetch(id)
 
-          return tokenMetadata ? convertDecimalsToNumber(tokenMetadata) : undefined
+          return tokenMetadata ? (convertDecimalsToNumber(tokenMetadata) as UnlistedFT) : undefined
```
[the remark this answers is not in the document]
nop33: Again, dev experience, I find it nicer to see `UnlistedFT` than an object with all its properties, especially when it is used in combined types like `ListedFT | UnlistedFT | NFT | ...`

apps/desktop-wallet/src/features/transactionsDisplay/transactionDetailsModal/TransactionDetailsModal.tsx
nop33: Not sure exactly what you mean, can you demonstrate? Will code navigation work fine if we do what you propose? (Cmd + Click on component name should navigate to the component file)

apps/desktop-wallet/src/features/transactionsDisplay/transactionDetailsModal/types.ts
nop33: Indeed, but what if we want to share them amongst many components that are in different fileS?

apps/desktop-wallet/src/features/transactionsDisplay/transactionRow/OtherAmounts.tsx
nop33: I believe I only put in types.ts the types that are shared. I do prefer to keep the types close to the component (as long as we do not duplicate them)

apps/desktop-wallet/src/utils/transactions.ts
nop33: You're probably right, I'll have a look!

nop33:
> I'm getting a bunch of error in the console:

@mvaivre  In my last PR these errors do not appear anymore. Do you mind checking there instead? 🙏  https://github.com/alephium/alephium-frontend/pull/835

apps/desktop-wallet/src/api/apiDataHooks/utils.ts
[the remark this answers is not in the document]
nop33: Good catch, I merged these 2 in one here: c405cfdff

apps/desktop-wallet/src/utils/transactions.ts
nop33: Good one! I split most of these functions and hooks into `features/transactionsDisplay` and `features/send` in this commit: e195942bd

Only a very small file remains: https://github.com/alephium/alephium-frontend/blob/next-tanstack-address-token-rows/apps/desktop-wallet/src/utils/transactions.ts

apps/desktop-wallet/src/api/queries/tokenQueries.ts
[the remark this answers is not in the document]
mvaivre: Okay. I had issue in the past with casting. There were some mistakes (fields not defined) that were not catch by the compiler because of this. That's why I'm kinda hesitant now.

apps/desktop-wallet/src/features/transactionsDisplay/transactionDetailsModal/TransactionDetailsModal.tsx
mvaivre: Same as what we do in our libs, like in our shared lib: `src/store/index.ts`

```ts
export * from '@/store/store'
export * from '@/store/addresses'
export * from '@/store/prices'
export * from '@/store/assets'
export * from '@/store/network'
export * from '@/store/global'
export * from '@/store/settings'
export * from '@/store/clients'
```

nop33: Ah, I see. Well, we could do that. But the LOC won't change because then we'd end up with

```ts
import {
 AddressesDataRows,
 DirectionalInfo,
 FTAmounts,
 FTsDataListRow,
 GasUTXOsExpandableSection,
 LockTimeDataListRow,
 NFTsDataListRow,
 NSTsDataListRow,
 TransactionType
} from '@/features/transactionsDisplay/transactionDetailsModal'
```

Then we need to always remember to maintain it. It's not rly a pattern we have in our apps, unlike our exportable packages.

Another thing, we tend to have `index.ts` fles to represent the actual component (like `AddressesPage/index.tsx`, but tbh I never liked that cause it doesn't help with file navigation so here I went for `transactionDetailsModal/TransactionDetailsModal.tsx`.

Anyway, I personally find it extra work for not much gain. WDYT?

apps/desktop-wallet/src/features/transactionsDisplay/transactionDetailsModal/types.ts
nop33: Not sure what you mean. The `TransactionDetailsModalSectionProps` is used by 11 components in 11 different files. Where should I put it if not in a dedicated file?

apps/desktop-wallet/src/api/queries/tokenQueries.ts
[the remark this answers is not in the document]
nop33: I think you've got a very good point here. We should probably prefer to define return types instead of casting. I've updated the code in this commit to reflect that. WDYT of this change?  d67ef36cf

apps/desktop-wallet/src/api/apiDataHooks/useNFT.ts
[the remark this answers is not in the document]
nop33: Linked to my [earlier comment](https://github.com/alephium/alephium-frontend/pull/832#discussion_r1762675555), if I change this to `UnlistedFT`, the compiler properly complains, as expected:

image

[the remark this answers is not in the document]
nop33:
```ts
{
  ...nftMetadata,
  ...nftData
} as NFT)
```

To remove the above type casting, we'd need to do this, which I find cumbersome... 

```ts
const useGenerateNFT = (
  nftMetadata: NFTMetadata | undefined,
  nftData: NFTTokenUriMetaData | undefined
): NFT | undefined =>
  useMemo(
    () =>
      !!nftMetadata && !!nftData
        ? {
            ...nftMetadata,
            ...nftData
          }
        : undefined,
    [nftData, nftMetadata]
  )
```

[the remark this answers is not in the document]
nop33: Both of the above snippets achieve the desired result which is this:

image

Without them, the result looks like this:

image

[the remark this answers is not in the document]
nop33: Seems like people in this thread agree that the `as` type assertion (aka casting in our discussion) is not something we should resort to. I'll try removing them and see how that affects the code and the dev experience.

https://www.reddit.com/r/typescript/comments/wd3f7j/should_i_avoid_casting_types/

apps/desktop-wallet/src/api/apiDataHooks/useToken.ts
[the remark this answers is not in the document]
nop33: So here, if I remove all type assertions, the return value of `useToken` looks like this. It's impossible to tell what it returns (notice the `...` at the end)

image

Not sure what to do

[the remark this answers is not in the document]
nop33: What we could do  (which we are already doing btw here) is this:

```ts
type Token = {
  data: ListedFT | UnlistedFT | NFT | NonStandardToken
  isLoading: boolean
}

const useToken = (id: TokenId): Token => {
```

I like that in the places where `useToken` is used I see the full list of available token options:

image


even if in the hook itself I only see `Token` when I hover over the hook name, but then again the definition is right above anyway so it doesn't matter

image

I'd go for this approach instead of type assertion.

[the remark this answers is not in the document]
nop33: Implemented in 1c490834c

apps/desktop-wallet/src/api/apiDataHooks/useNFT.ts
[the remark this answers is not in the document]
nop33: Implemented in 1c490834c

apps/desktop-wallet/src/features/transactionsDisplay/transactionDetailsModal/types.ts
mvaivre: NVM, I thought there was a `TransactionDetailsModalSection` component :)

PR #833, opened 2024-09-11
Add `outputs` field for Explorer turbo config

h0ngcha0:
CI pipeline for deploying Explorer is consistently failing, the reason seems to be that the `build` directory is not generated after running `pnpm ci:build`. 

I am not sure when/where this has become an issue, adding the `outputs` fields in turbo config fixes this.


nop33 (approved):
LGTM 👍

PR #835, opened 2024-09-12
Rework address token rows (Tanstack)

nop33:


apps/desktop-wallet/src/App.tsx
```diff
@@ -86,7 +86,7 @@ const App = () => {
   const _window = window as unknown as AlephiumWindow
   const electron = _window.electron
 
-  useInitializeClient()
+  useInitializeThrottledClient()
```
nop33, line 89: Until we decide if we port this to all our projects I created a new client to avoid breaking the other apps.

apps/desktop-wallet/src/api/addresses.ts
```diff
@@ -17,7 +17,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
 import { NonSensitiveAddressData } from '@alephium/keyring'
-import { AddressHash, client } from '@alephium/shared'
+import { AddressHash, throttledClient } from '@alephium/shared'
```
nop33, line 20: I replace `client` with `throttledClient` throughout the app.

apps/desktop-wallet/src/api/addressesUnlistedTokensHooks.ts
```diff
@@ -46,11 +45,11 @@ export const useAddressesUnlistedFTs = (addressHash?: AddressHash) => {
   const { data, isLoading } = useQueries({
     queries: chunk(unlistedFungibleTokenIds, TOKENS_QUERY_LIMIT).map((ids) => ({
       queryKey: ['tokens', 'fungible', 'unlisted', ids],
-      queryFn: () => client.explorer.tokens.postTokensFungibleMetadata(ids),
+      queryFn: () => throttledClient.explorer.tokens.postTokensFungibleMetadata(ids),
       staleTime: Infinity
     })),
     combine: (results) => ({
-      data: results.flatMap(({ data }) => data && data.map(convertDecimalsToNumber)).filter(isDefined) as UnlistedFT[],
+      data: results.flatMap(({ data }) => data && data.map(convertTokenDecimalsToNumber)).filter(isDefined),
```
nop33, line 52: Casting removed, from the previous PR review

apps/desktop-wallet/src/api/apiDataHooks/address/useAddressAlphBalances.ts
nop33, line 1: As discussed on Slack, I introduced the `/address` and `/wallet` hooks.

apps/desktop-wallet/src/api/apiDataHooks/address/useAddressFTs.ts
```diff
@@ -0,0 +1,80 @@
...
+  })
+
+  const { data: tokenPrices } = useTokenPrices({ skip: !sort })
+
+  return {
+    unlistedFTs: useMemo(
+      () => (sort ? orderBy(unlistedFTs, ['name', 'id'], ['asc', 'asc']) : unlistedFTs),
+      [sort, unlistedFTs]
+    ),
+    listedFTs: useMemo(
```
nop33, line 55: As you proposed last time, the sorted tokens are separated into different keys

apps/desktop-wallet/src/api/apiDataHooks/address/useAddressTokensBalances.ts
nop33, line 1: This is a failed attempt from GitHub to create a good diff... It thought I renamed the file. Oh well.

apps/desktop-wallet/src/api/apiDataHooks/useSeparateTokens.ts
nop33, line 1: This is an example of a utility hook to DRY the actual `/address` and `/wallet` hooks.

apps/desktop-wallet/src/api/apiDataHooks/useToken.ts
```diff
@@ -52,20 +52,18 @@ const useToken = (id: TokenId): Token => {
   )
 
   return {
-    data: listedFT
-      ? (listedFT as ListedFT)
-      : unlistedFT
-        ? (unlistedFT as UnlistedFT)
-        : nft || ({ id } as NonStandardToken),
+    data: listedFT ?? unlistedFT ?? nft ?? { id },
```
nop33, line 55: removed the type assertions, everything seem fine.

apps/desktop-wallet/src/api/apiDataHooks/wallet/combineBalances.ts
nop33, line 1: Shared between 2 different hooks

apps/desktop-wallet/src/api/apiDataHooks/wallet/useWalletAlphBalancesTotal.ts
```diff
@@ -0,0 +1,45 @@
...
+import { useQueries } from '@tanstack/react-query'
+
+import { DataHook, SkipProp } from '@/api/apiDataHooks/types'
+import combineBalances from '@/api/apiDataHooks/wallet/combineBalances'
+import useWalletLastTransactionHashes from '@/api/apiDataHooks/wallet/useWalletLastTransactionHashes'
+import { addressAlphBalancesQuery } from '@/api/queries/addressQueries'
+import { useAppSelector } from '@/hooks/redux'
+import { DisplayBalances } from '@/types/tokens'
+
+const useWalletAlphBalancesTotal = (props?: SkipProp): DataHook<DisplayBalances | undefined> => {
```
nop33, line 28: I prefer to see 

```ts
useWalletAlphBalancesTotal({ skip: isLoadingWhatever })
```

instead of 

```ts
useWalletAlphBalancesTotal(isLoadingWhatever
```

but I also want to be able to to 

```ts
useWalletAlphBalancesTotal()
```

so there we go :)

apps/desktop-wallet/src/api/apiDataHooks/address/useAddressTokensBalances.ts
nop33, line 1: Not accurant file rename detection from GH...

apps/desktop-wallet/src/api/apiDataHooks/useAlphBalancesByAddress.ts
```diff
@@ -32,8 +32,8 @@ export interface AddressesAlphBalances {
   isLoading: boolean
 }
 
-const useAddressesAlphBalances = (addressHash?: AddressHash): AddressesAlphBalances => {
-  const { data: latestTxHashes, isLoading: isLoadingLatestTxHashes } = useAddressesLastTransactionHashes(addressHash)
+const useAlphBalancesByAddress = (): AddressesAlphBalances => {
```
nop33, line 35: I find this name gives a better idea of what is returned (an object where keys are address hashes and values are balances)

apps/desktop-wallet/src/api/apiDataHooks/useSeparateTokens.ts
nop33, line 1: An example of a utility hook created to DRY other hooks.

apps/desktop-wallet/src/api/apiDataHooks/useSortFTs.ts
nop33, line 1: Another example of a utility hook.

apps/desktop-wallet/src/api/apiDataHooks/wallet/combineBalances.ts
nop33, line 1: Do you think we should keep the "simpler" combine functions in the util.ts file, or each in separate files or should we should move this one into that file? I'm fine with all.

apps/desktop-wallet/src/api/queries/tokenQueries.ts
```diff
@@ -16,48 +16,45 @@ You should have received a copy of the GNU Lesser General Public License
...
     queryKey: ['token', 'type', id],
     queryFn: !skip
       ? async () => {
-          const tokenInfo = await tokenTypeBatchFetcher.fetch(id)
+          const tokenInfo = await batchers.tokenTypeBatcher.fetch(id)
 
           return tokenInfo?.stdInterfaceId
             ? { ...tokenInfo, stdInterfaceId: tokenInfo.stdInterfaceId as TokenStdInterfaceId }
-            : undefined
+            : null
```
nop33, line 51: Tanstack queries should never return `undefined` because that means the query has failed.

apps/desktop-wallet/src/api/queries/tokenQueries.ts
```diff
@@ -86,9 +83,9 @@ export const fungibleTokenMetadataQuery = ({ id, skip }: TokenQueryProps) =>
     queryKey: ['token', 'fungible', 'metadata', id],
     queryFn: !skip
       ? async () => {
-          const tokenMetadata = await fungibleTokenMetadataBatchFetcher.fetch(id)
+          const tokenMetadata = await batchers.ftMetadataBatcher.fetch(id)
```
nop33, line 86: please see below for the reason `batchers` is needed.

apps/desktop-wallet/src/features/assetsLists/FTsBalancesList.tsx
nop33, line 1: This file is a good demonstration of the concept of "address" and "wallet" components that visually they render the same things (and thus use the same shared styled components) but get their data from different hooks.

apps/desktop-wallet/src/features/send/sendUtils.ts
nop33, line 1: Code moved from `utils/transactions.ts`

apps/desktop-wallet/src/features/transactionsDisplay/transactionDisplayUtils.ts
nop33, line 1: code moved from `utils/transactions.ts`

apps/desktop-wallet/src/features/walletConnect/walletConnectContext.tsx
```diff
@@ -410,7 +410,11 @@ export const WalletConnectContextProvider: FC = ({ children }) => {
             const p = request.params as ApiRequestArguments
 
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            const call = (client.explorer as any)[`${p.path}`][`${p.method}`] as (...arg0: any[]) => Promise<any>
+            const call = (throttledClient.explorer as any)[`${p.path}`][`${p.method}`] as (
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              ...arg0: any[]
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            ) => Promise<any>
```
nop33, line 417: We should look into whether this can eventually be simplified, i rly don't like it, but I also don't wanna touch it.

apps/mobile-wallet/package.json
```diff
@@ -134,7 +134,7 @@
   "jest": {
     "preset": "jest-expo",
     "transformIgnorePatterns": [
-      "node_modules/(?!(?:.pnpm/)?((jest-)?react-native|@react-native(-community)?|expo(nent)?|@expo(nent)?/.*|@expo-google-fonts/.*|react-navigation|@react-navigation/.*|@unimodules/.*|unimodules|sentry-expo|native-base|react-native-svg|nanoid))"
+      "node_modules/(?!(?:.pnpm/)?((jest-)?react-native|@react-native(-community)?|expo(nent)?|@expo(nent)?/.*|@expo-google-fonts/.*|react-navigation|@react-navigation/.*|@unimodules/.*|unimodules|sentry-expo|native-base|react-native-svg|nanoid|p-throttle))"
```
nop33, line 137: It was so freakign hard to make tests pass with this stupid p-throttle that only exposes an ESM version... The diff in the next few files is related to this.

packages/shared-react/src/useInitializeThrottledClient.ts
```diff
@@ -0,0 +1,55 @@
...
+export const useInitializeThrottledClient = () => {
+  const network = useSharedSelector((s) => s.network)
+  const dispatch = useSharedDispatch()
+
+  const shouldStopTryingToInitializeClient = network.status !== 'offline'
+
+  const initializeClient = useCallback(async () => {
+    try {
+      throttledClient.init(network.settings.nodeHost, network.settings.explorerApiHost)
+      batchers.init()
```
nop33, line 36: This line is very important. The bug that I couldn't figure out why is happening is that UNLESS we re-create the batchers when the network changes, the `fetcher` function of each batcher will remain the same as the "old" client, with the old explorer and node endpoints. So, similar to how we update the Client class when the network changes, we also need to update the batchers.

packages/shared/src/api/throttledClient.ts
```diff
@@ -0,0 +1,59 @@
...
+import { ExplorerProvider, NodeProvider } from '@alephium/web3'
+import pThrottle from 'p-throttle'
+
+import { defaultNetworkSettings } from '@/network'
+import { NetworkSettings } from '@/types/network'
+
+const throttle = pThrottle({
+  limit: 10,
+  interval: 1000
+})
```
nop33, line 28: This is the same API rate limit as the one confirmed by Honchao.

packages/shared/src/api/throttledClient.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const throttledFetch = throttle((url, options = {}) => fetch(url, options))
```
nop33, line 30: You would think that 

```ts
const throttledFetch = throttle(fetch)
```

is enough, but you'd be surprised to know how weird behavior this gives. I still don't fully understand why, but this works so there you go

apps/desktop-wallet/src/api/apiDataHooks/utils.ts
```diff
@@ -24,17 +24,12 @@ export const combineIsLoading = <R>(results: UseQueryResult<R, Error>[]) => ({
...
-
 export const flatMapCombine = <R>(results: UseQueryResult<R | R[], Error>[]) => ({
   data: results.flatMap(({ data }) => data).filter(isDefined),
   ...combineIsLoading(results)
 })
 
-export const mapCombineDefined = <R>(results: UseQueryResult<R, Error>[]) => ({
+export const combineDefined = <R>(results: UseQueryResult<R, Error>[]) => ({
   data: results.reduce((acc, { data }) => {
     if (data) acc.push(data)
```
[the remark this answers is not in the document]
nop33: Done in https://github.com/alephium/alephium-frontend/pull/841

packages/shared/src/api/throttledClient.ts
mvaivre: I had the exact same issue in the past! Same conclusion as yours :P

apps/desktop-wallet/src/features/assetsLists/types.ts
```diff
@@ -20,18 +20,15 @@ import { AddressHash } from '@alephium/shared'
 
 export type TokensTabValue = 'fts' | 'nfts' | 'nsts'
 
-export interface AssetsTabsProps {
+export type TokensTabsBaseProps = {
```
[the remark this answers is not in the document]
nop33: > Let's use interfaces only?

We can't use `interfaces` only, sometimes we have to use types. But in this example we can indeed use `interface`. A reason to use `type` instead of `interface` is when you want to merge 2 interfaces/types together.:

```ts
type A = {}
type B = {}

// with type:
type ABType = A & B

// with interface:
interface ABInterface extends A, B {}
```

In the above case, I prefer `type` over `interface` cause we don't need that ugly `{}` at the end.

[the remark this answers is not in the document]
nop33: Fixed in this PR!
- https://github.com/alephium/alephium-frontend/pull/842

apps/desktop-wallet/src/features/balancesOverview/TotalAlphBalance.tsx
```diff
@@ -45,25 +45,29 @@ const TotalAlphBalance = ({ className, type }: TotalAlphBalanceProps) => {
...
-  const { isLoading } = useAddressesAlphBalancesTotal()
+  if (isLoading) return <AmountLoader />
 
-  return isLoading ? <SkeletonLoader height="30px" /> : <AmountStyled tokenId={ALPH.id} tabIndex={0} value={balance} />
+  return data?.lockedBalance !== undefined && <AmountStyled tokenId={ALPH.id} tabIndex={0} value={data.lockedBalance} />
 }
 
 export default TotalAlphBalance
 
+const AmountLoader = () => <SkeletonLoader height="30px" />
```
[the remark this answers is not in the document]
nop33: It could! I'll work on it in my new PR:
- https://github.com/alephium/alephium-frontend/pull/842

I'll let u know when it's done.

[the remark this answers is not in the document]
nop33: Implemented in e58a90a8d:
- https://github.com/alephium/alephium-frontend/pull/842

apps/desktop-wallet/src/features/assetsLists/types.ts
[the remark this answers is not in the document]
mvaivre: Thanks! Yeah I know the differences between the 2, I was really talking about this specific case :) Because we usually use interfaces for Props.

mvaivre (approved):
Awesome! 👏 
Getting there..! 🚀

PR #846, opened 2024-09-21
Show toast when failing to authenticate

nop33:
People still seem to have problems with infinite auth loops: https://github.com/alephium/alephium-frontend/issues/695

This PR shows a tooltip when the auth fails with a message, hoping that it will help us get better feedback.


mvaivre (approved):
Thanks, good idea.

PR #850, opened 2024-09-24
Update WC client status if no WC client found

### LINKED ISSUE #849: WC client might become uninitialized after being initialized while app still in memory

nop33:
Closes #849

### DIFF
--- a/apps/mobile-wallet/src/contexts/walletConnect/WalletConnectContext.tsx
+++ b/apps/mobile-wallet/src/contexts/walletConnect/WalletConnectContext.tsx
@@ -211,6 +211,12 @@ export const WalletConnectContextProvider = ({ children }: { children: ReactNode
     }
   }, [dispatch, walletConnectClientInitializationAttempts])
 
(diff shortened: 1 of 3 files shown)


apps/mobile-wallet/src/contexts/walletConnect/WalletConnectContext.tsx
```diff
@@ -211,6 +211,12 @@ export const WalletConnectContextProvider = ({ children }: { children: ReactNode
     }
   }, [dispatch, walletConnectClientInitializationAttempts])
 
+  useEffect(() => {
+    if (walletConnectClientStatus === 'initialized' && !walletConnectClient) {
+      dispatch(walletConnectClientInitializeFailed(t('Lost connection to WalletConnect')))
```
nop33, line 216: This will set the status to `initialization-failed` which will update the UI accordingly and the user won't be stuck at a "success" indication.

mvaivre (approved):
Again, good catch!

Issue #712, opened 2024-07-04
Use Tanstack


nop33:
# Goals

- Reduce load on explorer backend (less requests)
- Improve devX by having to manage less async operations through side-effects
- Improve app performance

# Steps

- [x] 1. Replace Redux data fetching and slices with [Tanstack Query](https://tanstack.com/query/latest) for a (working) part of the codebase of the desktop wallet.
- [x] 2. Persist latest tx hash of each wallet address. When polling, query explorer backend for latest tx hash of each address. If returned hashes are different than the persisted ones, invalidate related Tanstack caches.
    - Related: https://github.com/alephium/alephium-frontend/issues/726
- [x] 3. Persist Tanstack state between app restarts. On app start, apply same methodology like polling.
    - Related: https://github.com/alephium/alephium-frontend/issues/126
- [ ] 4. Apply above steps to mobile wallet (see [2]) and extract common logic to shared package

# TODO

- [x] Install `@tanstack/eslint-plugin-query` for better linting [1]
- [ ] Install `react-native-react-query-devtools` for React Native [1]
    - Or maybe this Expo plugin `@dev-plugins/react-query`? [2]
    - Or even Reactotron plugin `reactotron-react-query`? [3]

[1]: https://tanstack.com/query/latest/docs/framework/react/devtools
[2]: https://tanstack.com/query/latest/docs/framework/react/react-native
[3]: https://github.com/hsndmr/reactotron-react-query

Issue #126, opened 2023-11-24
Implement offline caching strategy


mvaivre:
In order to reduce our dependency on external API like CoinGecko and offer a smoother UX when the network is spotty, let's implement an offline caching strategy.

- Latest known amount and prices should be available at wallet start.
- If an internet connection is available, we sync that data.
- We add an indicator in the header to inform the user about the status of the data sync (in sync vs. not in sync vs. offline)

For history data, let's implement a granular fetching, which only try and get the data for the dates which were not loaded yet.

This should help mitigate #115

nop33:
Just wanted to say that I am very happy we now have a mono-repo, opening this issue in the previous set up wouldn't have been so simple :D 

> For history data, let's implement a granular fetching, which only try and get the data for the dates which were not loaded yet.

Related to https://github.com/alephium/alephium-frontend/issues/60

nop33:
We can totally persist the Redux store so that on app launch the user seems the latest data that they saw when they last closed the app. I am not sure this will address the problems that led to opening this issue:
1. Storing the latest known amounts of each address (basically persisting the addresses slice) does reduce our dependency on Coingecko API, and neither does storing the latest known price, since we will anyways need to fetch the price to "sync". The only benefit is (arguably) UX related. The amount of API calls is the same. Am I missing something?

As for the historic data, that could be discussed in #60, but it also doesn't affect external APIs like Coingeko...

nop33:
Interesting article: https://blog.logrocket.com/persist-state-redux-persist-redux-toolkit-react/

PR #851, opened 2024-09-24
Persist Tanstack cache

nop33:
Docs: 
- https://tanstack.com/query/latest/docs/framework/react/plugins/persistQueryClient

Related to:
- #712 
- #126

### DIFF
--- a/apps/desktop-wallet/src/index.tsx
+++ b/apps/desktop-wallet/src/index.tsx
@@ -21,8 +21,8 @@ import '@/i18n'
 import '@yaireo/tagify/dist/tagify.css' // Tagify CSS: important to import after index.css file
 
 import isPropValid from '@emotion/is-prop-valid'
-import { QueryClientProvider } from '@tanstack/react-query'
 import { ReactQueryDevtools } from '@tanstack/react-query-devtools'
+import { PersistQueryClientProvider } from '@tanstack/react-query-persist-client'
 import { StrictMode, Suspense } from 'react'
 import ReactDOM from 'react-dom'
 import { Provider } from 'react-redux'
@@ -35,6 +35,7 @@ import Tooltips from '@/components/Tooltips'
 import AnalyticsProvider from '@/features/analytics/AnalyticsProvider'
 import * as serviceWorker from '@/serviceWorker'
 import { store } from '@/storage/store'
+import tanstackIndexedDBPersister from '@/storage/tanstackIndexedDBPersister'
 
 // The app still behaves as if React 17 is used. This is because
 // `react-custom-scrollbars` is not working with React 18 yet.
@@ -48,10 +49,13 @@ ReactDOM.render(
         <Router>
           <Suspense fallback="loading">
             <StyleSheetManager shouldForwardProp={shouldForwardProp}>
-              <QueryClientProvider client={queryClient}>
+              <PersistQueryClientProvider
+                client={queryClient}
+                persistOptions={{ persister: tanstackIndexedDBPersister, maxAge: Infinity }}
+              >
                 <App />
                 <ReactQueryDevtools initialIsOpen={false} />
-              </QueryClientProvider>
+              </PersistQueryClientProvider>
               <Tooltips />
             </StyleSheetManager>
           </Suspense>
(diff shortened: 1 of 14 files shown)


apps/desktop-wallet/src/index.tsx
```diff
@@ -48,10 +49,13 @@ ReactDOM.render(
         <Router>
           <Suspense fallback="loading">
             <StyleSheetManager shouldForwardProp={shouldForwardProp}>
-              <QueryClientProvider client={queryClient}>
+              <PersistQueryClientProvider
+                client={queryClient}
+                persistOptions={{ persister: tanstackIndexedDBPersister, maxAge: Infinity }}
```
nop33, line 54: My reasoning for using `Inifnity` as the `maxAge`:

Definitions:
- `maxAge`: _The max-allowed age of the persisted cache. If a persisted cache is found that is older than this time, it will be **silently** discarded (defaults to 24 hours)_
- `gctime`: _The time that `inactive` query results will be kept in memory._
- `inactive`: _Query results that have no mount React components to consume them._

Our current setup has this config:
- `maxAge: Infinity`
- `gcTime`: 5 minutes (default)

The docs suggest that the `gcTime` should be `>= maxAge`, otherwise the garbage collector will remove data from the persisted cache before `maxAge` is reached. However, I don't think this recommendation applies to us.

My argument for the above config is that we don't want the persisted cache discarded after a specific period of time. We want to keep the query data **indefinitely** when `latestTxHash` is still the latest one. We only want it to be discarded when data is garbage collected. Data will be garbage collected after being `inactive` for 5 minutes. Those data are mainly the query results that include an old `latestTxHash`. We have no use for that data.

So let's go through an example:
1. User updates their app to a version that includes Tanstack
2. User launches app and unlocks a wallet
3. ALL data is fetched and they get persisted
4. User closes the app
5. User opens the app after 3 days: The app gets hydrated with the persisted data. Almost all data is currently `inactive` since the user hasn't unlocked any wallet.
7. User unlocks a wallet: The rendered React components will now use some of the `inactive` query data which will make them now `active`. Some old query data will remain `inactive`
8. 5 minutes after the app was launched, any `inactive` data will be garbage collected and hence also removed from persisted cache.

There is only one catch:
1 - 6: Same as above
7. User stays in the "welcome" screen for more than 5 minutes: Since there hasn't been any React components that needed any of the re-hydrated `inactive` data, all data will be garbage collected and hence removed from the persisted cache.
8. User unlocks a wallet after 5 minutes: ALL data are fetched from scratch (as our current prod app behaves)

Reasons why user will remain in "welcome" screen for "too long":
1. Well, because they want to
2. Because of the auto-lock mechanism

Maybe we should set the `gcTime` higher. I would not set it to 24 days though which is the maximum because then it will never run (unless a user keeps their computer on and the app running for more than 24 days). I would also not set it to 1 day. Maybe 1 hour?

---

PS: `stateTime` is irrelevant for this discussion. For for the sake of completion:
- `staleTime`: The time after which the data fetched by the query is considered `stale`. Before that, data is considered `fresh`.

Ref:
- https://tanstack.com/query/latest/docs/framework/react/guides/important-defaults
- https://tanstack.com/query/latest/docs/framework/react/plugins/persistQueryClient

mvaivre (approved):
Really, just like that?? 😄 Nice.

LGTM, but we really need to test that thoroughly to make sure hydration works as expected.

apps/desktop-wallet/src/index.tsx
nop33: To extend the above "catch" case: If a user unlocks wallet A, then locks it, then unlocks wallet B, and keeps the app oper for longer than our `gcTime` (5 minutes), then all persisted cached data of wallet A will be deleted.

In other words: persisted data of a locked wallet get deleted after the `gcTime` (as well as any "inactive" query data of the unlocked wallet).

However, if we set `gcTime` higher, say 1 hour, then only users who keep the app running for longer than 1 hour will have their inactive data cleaned.

Optimally, what we would want is:
1. Persisted cached data of unlocked wallets to NOT be affected by `gcTime` (aka `gcTime: Infinity`)
1. Small `gcTime` data of unlocked wallet for as long as the wallet is unlocked.

mvaivre (approved):
Awesome. LGTM!

PR #855, opened 2024-09-25
NFT display improvements + sanitizing

mvaivre:


apps/explorer/src/api/assets/assetsApi.ts
```diff
@@ -157,7 +164,27 @@ export const assetsQueries = {
...
+              data = (await res.json()) as NFTTokenUriMetaData
+
+              if (!matchesNFTTokenUriMetaDataSchema(data)) {
+                data = undefined
+                return Promise.reject()
+              }
+            } catch (e) {
+              if (dataUri.startsWith('data:application/json;utf8,')) {
+                data = JSON.parse(dataUri.split('data:application/json;utf8,')[1])
+              }
```
nop33, line 179: This also needs to be checked if it matches the schema

nop33: Moreover, I believe that by using `axios`, we don't need to do this weird `startsWith('data` thing. See how it's done in the desktop wallet tanstack branch: https://github.com/alephium/alephium-frontend/blob/next-tanstack/apps/desktop-wallet/src/api/queries/tokenQueries.ts#L101-L119

package.json
```diff
@@ -53,7 +53,14 @@
...
       "fast-xml-parser@<4.4.1": ">=4.4.1",
       "elliptic@<6.5.7": ">=6.5.7",
-      "micromatch@<4.0.8": ">=4.0.8"
+      "micromatch@<4.0.8": ">=4.0.8",
+      "webpack@>=5.0.0-alpha.0 <5.94.0": ">=5.94.0",
+      "send@<0.19.0": ">=0.19.0",
+      "serve-static@<1.16.0": ">=1.16.0",
+      "path-to-regexp@>=2.0.0 <3.3.0": ">=3.3.0",
+      "vite@>=5.3.0 <5.3.6": ">=5.3.6",
+      "vite@>=5.3.0 <=5.3.5": ">=5.3.6",
```
nop33, line 62: This seems a bit unecessary to have both? Don't they do the same thing?

nop33: Also, since `vite` is in our devDependencies, why not update it there directly and instead override it?

package.json
```diff
@@ -53,7 +53,14 @@
       "fast-loops@<1.1.4": ">=1.1.4",
       "fast-xml-parser@<4.4.1": ">=4.4.1",
       "elliptic@<6.5.7": ">=6.5.7",
-      "micromatch@<4.0.8": ">=4.0.8"
+      "micromatch@<4.0.8": ">=4.0.8",
+      "webpack@>=5.0.0-alpha.0 <5.94.0": ">=5.94.0",
```
nop33, line 57: Same comment as below, it's in our devDeps, so let's update it there directly and remove it from overrides

package.json
```diff
@@ -53,7 +53,14 @@
       "fast-loops@<1.1.4": ">=1.1.4",
       "fast-xml-parser@<4.4.1": ">=4.4.1",
       "elliptic@<6.5.7": ">=6.5.7",
```
nop33, line 55: I know it's not part of this PR, but since I started researching if we are actually using these packages as deps or not, I discovered that we can update `elliptic` in the shared package instead of overriding it here.

nop33 (commented):
Can you please address the failing CI? Thanks!

package.json
nop33: Same for:
- electron-updater

package.json
mvaivre: Done!

package.json
mvaivre: Fixed

package.json
mvaivre: Done

apps/explorer/src/api/assets/assetsApi.ts
mvaivre: Done with Axios 👍

.changeset/rare-nails-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"@alephium/shared": patch
+---
+
+Export NFT schema check and saniizing functions
```
nop33, line 5: Typo.

nit: Also, I don't find this a particularly useful entry in our changelog. I'd replace it with

```suggestion
Sanitize NFT data
```

apps/desktop-wallet/package.json
```diff
@@ -141,10 +141,10 @@
     "ts-json-schema-generator": "^1.5.0",
     "type-fest": "^3.5.1",
     "typescript": "^5.5.3",
-    "vite": "^4.5.2",
+    "vite": "^5.4.8",
     "vite-plugin-svgr": "^3.2.0",
     "vite-tsconfig-paths": "^4.2.0",
-    "vitest": "^0.31.0",
+    "vitest": "^2.1.2",
```
nop33, line 147: wow, what a huge jump, no wonder you had troubles. thank you for this!

apps/explorer/src/pages/AddressInfoPage/NFTList.tsx
```diff
@@ -144,22 +144,14 @@ interface NFTItemProps {
...
-
-  return (
-    <NFTCardStyled onClick={onClick}>
-      <NFTPictureContainer>
-        <NFTThumbnail src={nft.file?.image} showPlayIconIfVideo playOnHover />
-      </NFTPictureContainer>
-      {nft.file?.name ? (
-        <NFTName>{nft.file.name}</NFTName>
-      ) : (
-        <MissingMetadataText>{t('Wrong/old format')}</MissingMetadataText>
```
nop33, line 158: What happens now when there are wrong or old format? The promise rejects, and then?

package.json
nop33: But I still see elliptic being listed here. It should be removed from this file and any other place where we override it

.changeset/rare-nails-knock.md
mvaivre: Done

apps/explorer/src/pages/AddressInfoPage/NFTList.tsx
nop33: I see now

image
image

all good. But maybe we could still extract some info even if not all of them pass the check. We could probably get the name of the NFT so that the user knows which one we refer to. WDYT?

nop33: http://localhost:5173/addresses/157iTPFJbe7FqaYGcQmomxqWY9G4n57945eERb6Zt6Ukw

nop33: nvm, the particular case above is because the ipfs seems to return an error.

In the future, once all apps are with Tanstack, we should merge all different ways of fetching and validating nft data.

Consider this discussion resolved ✅

nop33 (approved):

nop33 (changes requested):
Sorry, I approved it without checking the CI. Seems that elliptic is outdated

nop33 (approved):

PR #857, opened 2024-09-26
Delete address

nop33:
Closes:
- #632


apps/desktop-wallet/src/modals/AddressOptionsModal.tsx
```diff
@@ -137,6 +166,21 @@ const AddressOptionsModal = memo(({ id, addressHash }: ModalBaseProp & AddressOp
...
+            data-tooltip-content={t('To delete this address set another one as the default one first.')}
+            style={{ width: '100%' }}
+          >
+            <ModalFooterButton role="secondary" variant="alert" disabled={true}>
+              {t('Delete')}
+            </ModalFooterButton>
+          </div>
+        ) : (
+          <ModalFooterButton role="secondary" variant="alert" onClick={handleDeletePress} disabled={address.isDefault}>
+            {t('Delete')}
```
mvaivre, line 181: I wanted to suggest to use another word, but I remembered that you used "Forget" in your follow up PR, which I like :)

apps/desktop-wallet/src/features/modals/modalSlice.ts
```diff
@@ -40,7 +41,7 @@ const modalSlice = createSlice({
         modalAdapter.removeOne(state, id)
       })
 
-    builder.addMatcher(isAnyOf(walletSwitched, walletLocked, activeWalletDeleted), (state) => {
+    builder.addMatcher(isAnyOf(walletSwitched, walletLocked, activeWalletDeleted, addressDeleted), (state) => {
```
mvaivre, line 44: I understand why you did that here but I'm not sure I like it. The other actions (`walletSwitched`, `walletLocked` and `activeWalletDeleted`) are all at the wallet level. For those actions, it make sense to close all opened modal, because the whole context changes. 
 
It's different when you delete one address. I'm not sure that we should use this reducer every time we want to close more than 1 modal. In the future it could lead to closing modals that we don't want to close. WDYT?

One idea to fix this would be to allow `closeModal` to receive a modal name, allowing use to close specific modals from different part of the app.

nop33: I acknowledge your concerns. 

It's true that you can get to the AddressOptionsModal (where the delete button is) in multiple ways:
- from the Addresses page -> AddressDetailsModal
- from the TransactionDetailsModal -> AddressDetailsModal
- from an infinite number of TransactionDetailsModal -> AddressDetailsModal -> TransactionDetailsModal -> ... -> 

in all above cases, deleting an address would close all modals. Which I think it's acceptable for now. But I agree it could be improved.

> One idea to fix this would be to allow closeModal to receive a modal name, allowing use to close specific modals from different part of the app.

That'd mean that ALL modals with the same name would be closed, right? Even if they are multiple. That's also a bit weird. I think we should look into another solution.

Maybe we shouldn't close any modals. Maybe we should just saw a snackbar that says "address deleted successfully." WDYT?

nop33: Ah, now I remember why I wanted to close all modals. Imagine this:
1. User opens AddressDetailsModal
2. User opens AddressOptionsModal
3. User clicks "Forget"
4. User confirms in the ConfirmModal

If we only close the ConfirmModal and the AddressOptionsModal, then the AddressDetailsModal stays open with nothing to display

image

What we would optimally want is to close ANY modals related to the deleted address. That can be a tricky thing to define. That's why I went for simply closing all modals for now.

How do you think we should proceed?

nop33: If we want to identify which modals are related to an address, best I can think of is to create a reducer in the modal slice that loops over all open modals and checks if any of them includes a prop called `addressHash` that matches the deleted one...

mvaivre: I would say that closing all modals with the same name would be a little better than closing all the modals (a bit more specific), but I agree that this isn't optimal either.
However, we could fix this by augmenting `closeModal` a bit, allowing to pass not only a name, but also a param to check? We could maybe try this with a new `closeModals` action:

Usage:
```ts
closeModals({ name: 'AddressDetailsModal', params: { addressHash: deletedAddressHash }) })
```

WDYT?

nop33: Solution discussed on Slack and implemented :)

apps/desktop-wallet/src/modals/SettingsModal/CopyPrivateKeyConfirmationModal.tsx
```diff
@@ -26,57 +26,51 @@ import InfoBox from '@/components/InfoBox'
...
+import { AddressModalProps } from '@/features/modals/modalTypes'
 import { useAppDispatch, useAppSelector } from '@/hooks/redux'
 import CenteredModal from '@/modals/CenteredModal'
 import { selectAddressByHash } from '@/storage/addresses/addressesSelectors'
 import { copiedToClipboard, copyToClipboardFailed } from '@/storage/global/globalActions'
 
-export interface CopyPrivateKeyConfirmationModalProps {
-  addressHash: AddressHash
-}
+const CopyPrivateKeyConfirmationModal = memo(({ id, addressHash, ...props }: AddressModalProps) => {
```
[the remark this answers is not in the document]
nop33: Fixed!

mvaivre (approved):

PR #858, opened 2024-09-27
UI redesign part3

mvaivre:
In this PR:

- [x] Fix Dashboard bugs (amount not appearing sometimes)
- [ ] Migrate all modals to the new system. => Just lacking ONE... WalletConnectSessionRequestModal. 
- [x] Refine the new loading / landing UI


nop33:
Maybe it's time to break this PR in smaller chunks again? :/ It seems to be growing a lot. Maybe I can start reviewing and you can work on part 4?

apps/mobile-wallet/src/components/AnimatedBackground.tsx
```diff
@@ -76,6 +83,10 @@ const AnimatedCirclesBackground = ({
   // Circle animations
   const angle = useSharedValue(0)
 
+  // Using sin allows to smoothen movements (espacially when phon is upside down)
+  const sinRoll = useDerivedValue(() => Math.sin(gyroscope.sensor.value.roll))
+  const sinPitch = useDerivedValue(() => Math.sin(gyroscope.sensor.value.pitch))
+
```
mvaivre, line 89: This prevents the weird movements we've seen before

apps/mobile-wallet/src/components/ConsolidationModal.tsx
```diff
@@ -25,43 +25,55 @@ import AppText from '~/components/AppText'
...
-        </View>
-      </ScreenSection>
-      <ScreenSection centered>
-        <ButtonsRow>
-          <Button title={t('Cancel')} onPress={props.onClose} flex variant="accent" short />
-          <Button title={t('Consolidate')} onPress={onConsolidate} variant="highlight" flex short />
-        </ButtonsRow>
-      </ScreenSection>
-    </ModalContent>
+    <BottomModal modalId={id}>
```
mvaivre, line 49: Comment about the whole PR: every time I've wrapped a modal in `BottomModal`, the whole component is marked as changed, but it's not. 

What you should take a look at it how I extracted the callbacks, like `handleConsolidate` above. The main difference is that the modals are now responsible for closing themselves once the callbacks are executed. Before, it was the parents which were handling this.

apps/mobile-wallet/src/components/layout/Screen.tsx
```diff
@@ -29,23 +30,36 @@ export interface ScreenProps extends ViewProps {
   headerOptions?: BaseHeaderOptions & {
     type?: 'default' | 'stack'
   }
+  safeAreaPadding?: boolean
+  usesKeyboard?: boolean
 }
 
-const Screen = ({ children, headerOptions, ...props }: ScreenProps) => {
+const Screen = ({ children, headerOptions, safeAreaPadding, usesKeyboard, ...props }: ScreenProps) => {
```
mvaivre, line 37: Added the same functionalities as `ScrollScreen`. We used to use scroll screens when not actually needed. I've replaced some of them with simple non-scrollable screens.

apps/mobile-wallet/src/components/layout/TransactionsFlashList.tsx
```diff
@@ -0,0 +1,176 @@
...
+    )
+
+    const loadNextTransactionsPage = useCallback(async () => {
+      if (isLoading || allConfirmedTransactionsLoaded) return
+
+      dispatch(syncAllAddressesTransactionsNextPage({ minTxs: 10 }))
+    }, [allConfirmedTransactionsLoaded, dispatch, isLoading])
+
+    return (
+      <FlashList
```
mvaivre, line 94: It was using a FlatList before.

apps/mobile-wallet/src/features/modals/BottomModal.tsx
```diff
@@ -18,249 +18,104 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+  const {
+    modalHeightAnimatedStyle,
+    handleAnimatedStyle,
+    backdropAnimatedStyle,
+    handleContentSizeChange,
+    panGesture,
+    handleClose,
+    contentScrollHandlers,
+    isScrollable
+  } = useBottomModalState({
```
mvaivre, line 69: Extracted the animation and gesture logic in a hook, allowing reusability in `BottomModalFlashList`.

apps/mobile-wallet/src/features/modals/BottomModalFlashList.tsx
```diff
@@ -0,0 +1,170 @@
...
+          </ModalStyled>
+        </Container>
+      </KeyboardAvoidingViewStyled>
+    </GestureDetector>
+  )
+}
+
+export default BottomModalFlashList
+
+const KeyboardAvoidingViewStyled = styled(KeyboardAvoidingView)`
```
mvaivre, line 107: Styles are the same as in `BottomModal` for now. But it may change in the future - I want to refactor the modals header & title. The implementation will probably not be the same in both BottomModal variants.

apps/mobile-wallet/src/features/modals/ModalContent.tsx
```diff
@@ -38,6 +38,8 @@ export interface ModalFlatListContentProps<T> extends ModalContentBaseProps, Fla
 
 const scrollDefaultProps = { bounces: false, scrollEventThrottle: 16 }
 
+// TODO: DELETE THIS COMPONENT ONCE BOTTOM MODAL IS REFACTORED
```
mvaivre, line 41: We will actually keep ModalContent probably, but the file can be cleaned up a lot.

apps/mobile-wallet/src/features/modals/useBottomModalState.tsx
```diff
@@ -0,0 +1,260 @@
...
+      isContentDragged.value = true
+    },
+    onScrollEndDrag: () => {
+      isContentDragged.value = false
+
+      if (modalHeightDelta.value < -1) {
+        handleClose()
+      }
+    }
+  }
```
mvaivre, line 185: So, that was a pain in the 🍑 
It's not perfect yet, there are some glitches when the use scrolls down the content a bit, then go back up again. But for closing modals quickly it works well.

apps/mobile-wallet/src/features/send/modals/SelectAddressModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const dispatch = useAppDispatch()
+
+  const handleAddressPress = (addressHash: AddressHash) => {
+    dispatch(closeModal({ id }))
+    onAddressPress(addressHash)
+  }
+
+  return (
+    <BottomModalFlashList modalId={id}>
+      {(props) => (
```
mvaivre, line 44: Here's how I decided to implement FlashLists in modals. I prefer this rather than having a `Content` prop like before. And passing the whole `FlashList` as a children makes reusability easier than adding `FlashList` props directly to `BottomModalFlashList`.

apps/mobile-wallet/src/contexts/walletConnect/WalletConnectContext.tsx
mvaivre, line 1: This diff highlight the main limitation with the new way to open modals. Because we need to pass callbacks in `openModal`, we need to take into account where `openModal` is called. All needed callbacks must be declared before the function is used, which can be a little tricky. Also, more `useCallback` hooks are now required.

apps/mobile-wallet/src/screens/NFTs/NFTListScreen.tsx
```diff
@@ -31,13 +35,17 @@ type NFTListScreenProps = StackScreenProps<InWalletTabsParamList, 'NFTListScreen
 const NFTListScreen = ({ navigation }: NFTListScreenProps) => {
   const { t } = useTranslation()
   const { screenScrollY, screenScrollHandler } = useScreenScrollHandler()
+  const listRef = useRef<FlashList<NFT>>(null)
```
nop33, line 38: What is the reason this ref is needed?

apps/mobile-wallet/src/screens/LandingScreen.tsx
```diff
@@ -17,29 +17,27 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
 import RootStackParamList from '~/navigation/rootStackRoutes'
 import { storedWalletExists } from '~/persistent-storage/wallet'
-import AnimatedCirclesBackground from '~/screens/Dashboard/AnimatedCirclesBackground'
 import { methodSelected, WalletGenerationMethod } from '~/store/walletGenerationSlice'
 import { DEFAULT_MARGIN } from '~/style/globalStyle'
-import { themes } from '~/style/themes'
 import { resetNavigation } from '~/utils/navigation'
 
-interface LandingScreenProps extends StackScreenProps<RootStackParamList, 'LandingScreen'>, ScreenProps {}
+interface LandingScreenProps extends NativeStackScreenProps<RootStackParamList, 'LandingScreen'>, ScreenProps {}
```
nop33, line 40: This is how the screen looks the moment I launched the app.

1. The font color is black?
2. There is top and bottom padding within the rounded box.

image

apps/mobile-wallet/package.json
```diff
@@ -31,6 +31,7 @@
     "@react-navigation/elements": "^1.3.14",
     "@react-navigation/material-top-tabs": "^6.6.2",
     "@react-navigation/native": "^6.1.3",
+    "@react-navigation/native-stack": "^6.11.0",
```
nop33, line 34: The screen transactions are really not smooth anymore on Android...

## In this PR 

https://github.com/user-attachments/assets/3021b51e-0a23-47ec-8c82-7ff046821e44

## In banxa-next



https://github.com/user-attachments/assets/108d43f2-f498-4c9a-88ed-4b60aeb06e18

apps/mobile-wallet/src/features/modals/BottomModalFlashList.tsx
```diff
@@ -0,0 +1,170 @@
...
+  minHeight?: number
+  navHeight?: number
+  noPadding?: boolean
+  children: (scrollHandlers: {
+    contentContainerStyle: ContentStyle
+    onScroll: ContentScrollHandlers['onScroll']
+    onScrollBeginDrag: ContentScrollHandlers['onScrollBeginDrag']
+    onScrollEndDrag: ContentScrollHandlers['onScrollEndDrag']
+    onContentSizeChange: (w: number, h: number) => void
+  }) => ReactNode
```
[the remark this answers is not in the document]
nop33: Also this is a very weird pattern that we don't use anywhere else... We always define `children` as a `ReactNode`. Now here it's a function that returns a `ReactNode` which makes the instances of this component look very strange:

```ts
const NftGridModal = withModal<NftGridModalProps>(({ id, addressHash, nftsData }) => (
  <BottomModalFlashList modalId={id} noPadding>
    {(props) => <NFTsGrid addressHash={addressHash} nfts={nftsData} {...props} />}
  </BottomModalFlashList>
))
```

What if we just use a different prop than `children`, like `flashListRenderer`?

```ts
const NftGridModal = withModal<NftGridModalProps>(({ id, addressHash, nftsData }) => (
  <BottomModalFlashList
    modalId={id}
    noPadding
    flashListRenderer={(props) => <NFTsGrid addressHash={addressHash} nfts={nftsData} {...props} />}
  />
))
```

That's a much more familiar syntax.

apps/mobile-wallet/src/features/nftsDisplay/NftGridModal.tsx
```diff
@@ -19,23 +19,18 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
-    id={id}
-    Content={(props) => (
-      <ModalContent {...props}>
-        <NFTsGrid nfts={nftsData} scrollEnabled={false} />
-      </ModalContent>
-    )}
-  />
+const NftGridModal = withModal<NftGridModalProps>(({ id, addressHash, nftsData }) => (
+  <BottomModalFlashList modalId={id} noPadding>
+    {(props) => <NFTsGrid addressHash={addressHash} nfts={nftsData} {...props} />}
```
nop33, line 32: very very weird syntax, I would never expect the `children` prop to be a function, it took me a few minutes to figure out what's going on here.

apps/mobile-wallet/src/features/send/modals/SelectAddressModal.tsx
nop33: Please see by comments above.

apps/mobile-wallet/src/features/send/modals/SelectAddressModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+          renderItem={({ item: address }) => (
+            <AddressBox
+              key={address.hash}
+              addressHash={address.hash}
+              style={{
+                marginBottom: 20
+              }}
+              onPress={() => handleAddressPress(address.hash)}
+            />
+          )}
```
nop33, line 58: This is an example where we are using a render function, much more familiar to us than the new pattern you introduced

apps/mobile-wallet/src/features/send/modals/SelectAddressModal.tsx
nop33: and right below

apps/mobile-wallet/src/features/send/screens/DestinationScreen.tsx
```diff
@@ -168,110 +176,79 @@ const DestinationScreen = ({ navigation, route: { params }, ...props }: Destinat
...
+      contentPaddingTop
+      screenTitle={t('Destination')}
+      screenIntro={t('Send to an address, a contact, or one of your other addresses.')}
+      onScroll={screenScrollHandler}
+      {...props}
+    >
+      <ScreenSection>
+        <Controller
+          name="toAddressHash"
+          render={({ field: { onChange, onBlur, value } }) => (
```
nop33, line 191: Another example where we use render function prop

apps/mobile-wallet/src/features/walletconnect/WalletConnectPairingsModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+      await unpairFromDapp(pairingTopic)
+    }
+
+    const handlePasteWcUrlPress = () => {
+      onPasteWcUrlPress()
+    }
+
+    const handleScanQRCodePress = () => {
+      onScanQRCodePress()
+    }
```
nop33, line 62: That's a bit redundant wouldn't you say? It's literally a rename, exactly the same as

```ts
    const handlePasteWcUrlPress = onPasteWcUrlPress

    const handleScanQRCodePress = onScanQRCodePress
```

so , personally, I'd remove it alltogether

apps/mobile-wallet/src/hooks/layout/useAutoScrollOnDragEnd.tsx
```diff
@@ -16,17 +16,20 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
nop33, line 19: why is this needed?

nop33: if it's really needed, maybe it can just be moved to the line where it is needed instead of marking the whole file?

nop33: Interesting how this is not needed in `FlatList`:

```ts
export class FlatList<ItemT = any> extends FlatListComponent<
  ItemT,
  FlatListProps<ItemT>
> {}
```

apps/mobile-wallet/src/navigation/RootStackNavigation.tsx
```diff
@@ -104,7 +103,7 @@ const RootStackNavigation = ({ initialRouteName }: RootStackNavigationProps) =>
                 initialRouteName={initialRouteName || 'LandingScreen'}
                 screenOptions={{ headerShown: false }}
               >
-                <RootStack.Group screenOptions={{ cardStyleInterpolator: CardStyleInterpolators.forFadeFromCenter }}>
+                <RootStack.Group screenOptions={{ animation: 'fade' }}>
```
nop33, line 106: Maybe switching back to fade from center would fix most of the glitches?

apps/mobile-wallet/src/navigation/RootStackNavigation.tsx
```diff
@@ -72,7 +71,7 @@ import { mnemonicMigrated, walletUnlocked } from '~/store/wallet/walletActions'
 import { showExceptionToast, showToast } from '~/utils/layout'
 import { resetNavigation, rootStackNavigationRef } from '~/utils/navigation'
 
-const RootStack = createStackNavigator<RootStackParamList>()
+const RootStack = createNativeStackNavigator<RootStackParamList>()
```
nop33, line 74: I'll just leave a comment here so that we keep track of the Slack discussion

## This PR

https://github.com/user-attachments/assets/9eca8074-0fd6-4cbf-b6c0-701a01b8637f


## Prod (master)

https://github.com/user-attachments/assets/36865dee-a1fb-4cda-8110-2481d9332c86

apps/mobile-wallet/src/navigation/RootStackNavigation.tsx
```diff
@@ -252,10 +250,15 @@ const AppUnlockModal = ({ initialRouteName }: Required<RootStackNavigationProps>
   useAutoLock(unlockApp)
 
   return (
-    <Modal visible={!!lastUsedWalletId && biometricsRequiredForAppAccess && !isWalletUnlocked} animationType="none">
-      <View style={{ backgroundColor: 'black', flex: 1 }}>
-        <AnimatedCirclesBackground isAnimated />
-      </View>
+    <Modal visible={!!lastUsedWalletId && biometricsRequiredForAppAccess && !isWalletUnlocked} animationType="fade">
```
nop33, line 253: Hm, I remember that changing it to `fade` creates a weird artefact but I don't remember exactly how. I need to do a functional test.

apps/mobile-wallet/src/navigation/RootStackNavigation.tsx
nop33: I played a bit with different values for this (and also moving this prop to like 104 and it's only the `default` animation and the `simple_push` that create the glitch.

nop33 (commented):
I went through all files. The GH UI is very laggy due to the amount of them 😆 

I will do some more functional tests once I get some feedback on the review :+1:

apps/mobile-wallet/src/features/nftsDisplay/NftGridModal.tsx
mvaivre: It's a pattern that I've seen in other libraries in the past. Can't remember which ones, but I could dig.
Would you prefer going back to having a `FlashList` prop in `BottomModalFlashList`?

apps/mobile-wallet/src/features/send/modals/SelectAddressModal.tsx
mvaivre: I could go back to this one. I prefer the "new one" because I find it more readable (the DOM structure feels less deep and less cluttered to me). But yeah, I don't mind that much.

nop33: This new pattern is very confusing to me and I would prefer reusing the `render` function prop convention that we are using everywhere else

nop33: or we change it everywhere else and I learn to get used to a new pattern, I don't mind. I just don't like having 2 at the same time, I find it confusing.

apps/mobile-wallet/src/features/nftsDisplay/NftGridModal.tsx
nop33: I read the blogs, left a comment in the other thread

apps/mobile-wallet/src/features/send/modals/SelectAddressModal.tsx
nop33: The main reason I don't like it is that we are kinda "overriding" the default from React, which is 
```ts
children?: ReactNode | undefined
```

A React developer is familiar with it. But an outsider of our repo will have to learn it and get used to it

I'd personally prefer to not touch the `children` prop. I wonder what your thoughts are. I find the benefits of readability (which I do not necessarily agree on since it took me by surprise when reviewing the PR) to be negligible compared to drawbacks.

apps/mobile-wallet/src/screens/LandingScreen.tsx
mvaivre: Fixed in follow up PR!

apps/mobile-wallet/package.json
mvaivre: Fixed in follow up PR!

apps/mobile-wallet/src/features/send/modals/SelectAddressModal.tsx
mvaivre: I prefer the first example above because we can clearly see the hierarchy (children being indented). It's harder to see in the second one. 

BUT I agree with your arguments and will update this in the follow up PR :)

apps/mobile-wallet/src/navigation/RootStackNavigation.tsx
mvaivre: Fixed in the follow up PR!

apps/mobile-wallet/src/navigation/RootStackNavigation.tsx
mvaivre: Fixed in follow up PR!

apps/mobile-wallet/src/navigation/RootStackNavigation.tsx
mvaivre: I don't see any artefact in the next PR 🤞

apps/mobile-wallet/src/features/walletconnect/WalletConnectPairingsModal.tsx
mvaivre: Fixed in d3c078e8b0bfb90b9555a4a41190c727095d6161

apps/mobile-wallet/src/hooks/layout/useAutoScrollOnDragEnd.tsx
mvaivre: Moved to the line instead of marking the whole file (in the follow up PR)

apps/mobile-wallet/src/components/layout/Screen.tsx
```diff
@@ -29,23 +30,36 @@ export interface ScreenProps extends ViewProps {
...
     </ScreenStyled>
   )
+
+  return usesKeyboard ? (
+    <KeyboardAvoidingView behavior="height" style={{ flex: 1 }}>
+      {screen}
+    </KeyboardAvoidingView>
+  ) : (
+    screen
+  )
```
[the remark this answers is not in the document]
mvaivre: I need to double check. I seem to remember that this caused issue in some screens.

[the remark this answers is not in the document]
mvaivre: Tried it edc2b33801cf85d4bc3616267aec3156a139576b let's see if this causes any issue!

apps/mobile-wallet/src/contexts/SendContext.tsx
```diff
@@ -205,23 +222,6 @@ export const SendContextProvider = ({ children }: { children: ReactNode }) => {
...
-                props.onClose && props.onClose()
-              }}
-              fees={unsignedTxData.fees}
-            />
-          )}
-          isOpen={isConsolidateModalVisible}
-          onClose={() => setIsConsolidateModalVisible(false)}
-        />
-      </Portal>
-      {fundPasswordModal}
```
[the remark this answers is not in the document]
mvaivre: `triggerFundPasswordAuthGuard` is still available (and called where it used to be called). This will open the modal.

apps/mobile-wallet/src/features/modals/BottomModal.tsx
```diff
@@ -18,249 +18,104 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+    maximisedContent,
+    minHeight,
+    navHeight,
+    onClose
+  })
 
   return (
     <GestureDetector gesture={panGesture}>
-      <ExternalContainer>
+      <KeyboardAvoidingViewStyled behavior="height" enabled={!maximisedContent}>
```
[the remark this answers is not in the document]
mvaivre: Tried in the follow up PR!

apps/mobile-wallet/src/features/modals/BottomModalFlashList.tsx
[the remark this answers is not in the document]
mvaivre: Will make the switch in the follow up PR.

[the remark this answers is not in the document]
nop33: ok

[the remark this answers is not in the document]
nop33: Let me know when you make the switch so that I can approve this PR 🙏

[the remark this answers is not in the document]
mvaivre: Fixed in [e79ea23](https://github.com/alephium/alephium-frontend/pull/883/commits/e79ea2341730631776081f4a0fdc5910ec1543e9)

apps/mobile-wallet/src/contexts/walletConnect/WalletConnectContext.tsx
```diff
@@ -959,6 +820,171 @@ export const WalletConnectContextProvider = ({ children }: { children: ReactNode
...
   }
 
+  const openWalletConnectSessionProposalModal = useCallback(
+    () =>
+      sessionProposalEvent &&
+      dispatch(
+        openModal({
+          name: 'WalletConnectSessionProposalModal',
+          props: {
+            approveProposal,
```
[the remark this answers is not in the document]
mvaivre: https://github.com/alephium/alephium-frontend/issues/947

nop33 (approved):

PR #864, opened 2024-10-03
Sort addresses with Tanstack & delete multiple addresses advanced action

nop33:


apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletAddressesActivityTimestampsSorted.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { useMemo } from 'react'
+
+import { SkipProp } from '@/api/apiDataHooks/apiDataHooksTypes'
+import { useFetchWalletActivityTimestamps } from '@/api/apiDataHooks/wallet/useFetchWalletLastTransactions'
+import { useAppSelector } from '@/hooks/redux'
+import { selectDefaultAddress } from '@/storage/addresses/addressesSelectors'
+
+const useFetchWalletAddressesActivityTimestampsSorted = (props?: SkipProp) => {
+  const { hash: defaultAddressHash } = useAppSelector(selectDefaultAddress)
+  const { data, isLoading } = useFetchWalletActivityTimestamps({ skip: props?.skip })
```
mvaivre, line 29: Nit: I like when we name `data` to make it extra clear what those data are about.

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletTransactionsLimitted.tsx
```diff
@@ -0,0 +1,47 @@
...
+*/
+
+import { useQuery } from '@tanstack/react-query'
+
+import useLimitedWalletAddresses from '@/api/apiDataHooks/utils/useLimitedWalletAddresses'
+import { useFetchWalletLastTransaction } from '@/api/apiDataHooks/wallet/useFetchWalletLastTransactions'
+import { walletLatestTransactionsQuery } from '@/api/queries/transactionQueries'
+import { useAppSelector } from '@/hooks/redux'
+
+const useFetchWalletTransactionsLimitted = () => {
```
mvaivre, line 26: Typo

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletAddressesActivityTimestampsSorted.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import { orderBy } from 'lodash'
+import { useMemo } from 'react'
+
+import { SkipProp } from '@/api/apiDataHooks/apiDataHooksTypes'
+import { useFetchWalletActivityTimestamps } from '@/api/apiDataHooks/wallet/useFetchWalletLastTransactions'
+import { useAppSelector } from '@/hooks/redux'
+import { selectDefaultAddress } from '@/storage/addresses/addressesSelectors'
+
+const useFetchWalletAddressesActivityTimestampsSorted = (props?: SkipProp) => {
```
mvaivre, line 27: I'm confused. We have 2 very close hooks, this one and `useFetchWalletAddressesSortedByActivity`. 👀
Just comparing the names I can't tell the difference. I see that the latter rely on this one, and the later follow our naming convention but not this one. 
Looking at the code it seems that all `useFetchWalletAddressesSortedByActivity` do is removing the timestamps and returning just an array of AddressHashes.

There's probably something I'm missing, but as of now I'm not sure why 2 different hooks are necessary 🤔

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletBalancesAlph.ts
```diff
@@ -16,30 +16,60 @@ You should have received a copy of the GNU Lesser General Public License
...
+  useFetchWalletBalancesAlphBy({ combine: combineBalances, skip: props?.skip })
+
+export const useFetchWalletBalancesAlphByAddress = (props?: SkipProp) =>
+  useFetchWalletBalancesAlphBy({ combine: combineBalancesByAddress, skip: props?.skip })
+
+interface UseFetchWalletBalancesAlphProps<T> extends SkipProp {
+  combine: (results: UseQueryResult<AddressAlphBalancesQueryFnData>[]) => { data: T; isLoading: boolean }
+}
+
+const useFetchWalletBalancesAlphBy = <T>({ combine, skip }: UseFetchWalletBalancesAlphProps<T>) => {
```
mvaivre, line 47: Why keeping the "By" at the end?

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletTransactionsInfinite.ts
```diff
@@ -19,22 +19,22 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
 import { useFetchWalletLastTransaction } from '@/api/apiDataHooks/wallet/useFetchWalletLastTransactions'
 import { walletTransactionsInfiniteQuery } from '@/api/queries/transactionQueries'
 import { useAppSelector } from '@/hooks/redux'
-import { selectAllAddressHashes } from '@/storage/addresses/addressesSelectors'
 
-const useFetchWalletInfiniteTransactions = () => {
+const useFetchWalletTransactionsInfinite = () => {
   const networkId = useAppSelector((s) => s.network.settings.networkId)
-  const allAddressHashes = useAppSelector(selectAllAddressHashes) // TODO: Display warning if > 80 addresses
+  const { addressHashes, isLimited } = useLimitedWalletAddresses()
```
mvaivre, line 29: Would it make sense to simplify things by having only one `useFetchWalletAddressesses` hook that we would systematically use to get the wallet addresses? It could take care of limiting addresses if needed and sort them by activity by default. Would it make sense?

I'm still confused by the number of hooks that looks the same:
- useLimitedWalletAddresses (which is actually relying on a fetch, so we should probably use a "fetch" prefix here?)
- useFetchWalletAddressesSortedByActivity
- useFetchWalletAddressesActivityTimestampsSorted

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletTransactionsLimitted.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const networkId = useAppSelector((s) => s.network.settings.networkId)
+  const { addressHashes, isLimited } = useLimitedWalletAddresses()
```
mvaivre, line 28: Here again, I would avoid to have a specific `useFetchWalletTransactionsLimited` hook.
I'd rather have a single `useFetchWalletLatestTransactions` which returns a `isLimitedByAddressListSize` boolean, based on the returned value of a single `useFetchWalletAddresses` (which would return the `isLimited` boolean).

mvaivre (commented):
👏 
Part 1 of my review. There's quite a lot to discuss already so I'm stopping here for now, and maybe we can have a call to go over my questions! ☕

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletAddressesActivityTimestampsSorted.tsx
nop33: I understand your confusion. Let me explain:

> TL;DR:
> - `useFetchWalletAddressesSortedByActivity` is a 1-to-1 replacement for `useAppSelector(selectAllAddressHashes)`
> - `useFetchWalletAddressesActivityTimestampsSorted` is a more general hook that includes the timestamps

In many components we need to get the list of addresses of the wallet. Until now, we've been relying on Redux for that using:

```ts
const allAddressHashes = useAppSelector(selectAllAddressHashes)
```

The above was taking care of sorting the addresses due to the `lastUsed` property that the `Address` Redux slice entity had. The sorting was defined in the adapter. Now that all API-related data have been removed from Redux, we need to use Tanstack for sorting. Hence the creation of `useFetchWalletAddressesSortedByActivity` which returns **exactly** the same result as `useAppSelector(selectAllAddressHashes)`.

Another feature of the above hook is that until we have the last used timestamps data for ALL addresses, it will return the unsorted list, otherwise it would return a different result the moment a new timestamp is fetched, which would create a lot of jumping around in the UI:

```ts
data: !isLoading && !props?.skip ? sortedAddressHashes : allAddressHashes
```

The above `useFetchWalletAddressesSortedByActivity` hook is based on `useFetchWalletAddressesActivityTimestampsSorted`. The latter is a more basic hook that does not have the above "loading" feature that I mentioned, and it also returns the timestamp together with each address which is useful in the case of the bulk address delete where we preselect the addresses that have never been used.

Do you have another idea to achieve the desired results? Maybe with different names?

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletBalancesAlph.ts
nop33: Because there is `useFetchWalletBalancesAlph` already

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletTransactionsInfinite.ts
nop33: Let me know if the answer above covers this: https://github.com/alephium/alephium-frontend/pull/864#discussion_r1790267700

In short, we still need a way to extract only the hashes since that's what our components need (why extract them in every component and not DRY our code by extracting them in one place?), and in other cases we need additional information. In some cases we need max 80 addresses (when we fetch txs for example) and in other cases we want all the addresses (addresses page). There are also cases where we don't need to fetch anything and we simply want to check if an address is an internal address for example, so there we only use Redux.

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletTransactionsLimitted.tsx
nop33: Are you proposing of merging `useFetchWalletTransactionsLimited` and `useFetchWalletTransactionsInfinite`? Or are you proposing remaining this hook from `useFetchWalletTransactionsLimitted` to `useFetchWalletLatestTransactions`?

If it's the first case, that would result to a very complicated hook because one needs to use `useQuery` and the other `useInfiniteQuery` with much different logic for handling refreshing.

If it's the second case, I don't mind renaming, but it makes sense to me to have the prefix `Limited` and `Infinite`.

Is there a third case?

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletTransactionsLimitted.tsx
nop33: Fixed!

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletAddressesActivityTimestampsSorted.tsx
nop33: Renamed :+1:

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletTransactionsLimitted.tsx
mvaivre: Much better after the renames we talked about!

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletBalancesAlph.ts
mvaivre: `useFetchWalletBalancesAlphBase`, `useFetchWalletBalancesAlphRaw`?

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletAddressesActivityTimestampsSorted.tsx
mvaivre: Much better after the renames we talked about!

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletBalancesAlph.ts
nop33: How about we do the same we did for tokens where we have:
```ts
// Base hook, not exported
const useFetchWalletBalancesTokens

// exported hooks
export const useFetchWalletBalancesTokensArray
export const useFetchWalletBalancesTokensByToken
export const useFetchWalletBalancesTokensByAddress
```

so that'd mean

```ts
// Base hook, not exported
const useFetchWalletBalancesAlph

// exported hooks
export const useFetchWalletBalancesAlphArray
export const useFetchWalletBalancesAlphByAddress
```

apps/desktop-wallet/src/features/addressDeletion/useDeleteAddress.ts
```diff
@@ -0,0 +1,46 @@
...
+import { AddressHash } from '@alephium/shared'
+
+import { useAppDispatch, useAppSelector } from '@/hooks/redux'
+import { addressDeleted } from '@/storage/addresses/addressesActions'
+import { addressMetadataStorage } from '@/storage/addresses/addressMetadataPersistentStorage'
+
+const useDeleteAddress = () => {
+  const activeWalletId = useAppSelector((s) => s.activeWallet.id)
+  const dispatch = useAppDispatch()
+  const addressEntities = useAppSelector((s) => s.addresses.entities)
```
mvaivre, line 28: Why not selecting directly the address we want to delete?

mvaivre (commented):
👏

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletBalancesAlph.ts
nop33: Implemented in https://github.com/alephium/alephium-frontend/pull/869/commits/bd35c6ef3e8813f9d8df6e7695abc8e68dbe4883

apps/desktop-wallet/src/api/apiDataHooks/utils/useLimitedWalletAddresses.ts
```diff
@@ -0,0 +1,49 @@
...
+*/
+
+import { ADDRESSES_QUERY_LIMIT } from '@alephium/shared'
+import { useMemo } from 'react'
+
+import useFetchWalletAddressesHashesSortedByActivity from '@/api/apiDataHooks/wallet/useFetchWalletAddressesHashesSortedByActivity'
+import { useAppSelector } from '@/hooks/redux'
+import { selectAllAddressHashes } from '@/storage/addresses/addressesSelectors'
+
+const useLimitedWalletAddresses = () => {
```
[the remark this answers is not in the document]
nop33: update its name to `useCappedAddressesHashes` in the next PR: https://github.com/alephium/alephium-frontend/pull/869

apps/desktop-wallet/src/features/addressDeletion/ForgetAddressButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { useAppSelector } from '@/hooks/redux'
+import { ModalFooterButton } from '@/modals/CenteredModal'
+import { selectDefaultAddress } from '@/storage/addresses/addressesSelectors'
+
+interface ForgetAddressButtonProps {
+  addressHash: AddressHash
+  addressName: string
+}
+
+const ForgetAddress = ({ addressHash, addressName }: ForgetAddressButtonProps) => {
```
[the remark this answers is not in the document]
nop33: Fixed in 98d99bbc4

apps/desktop-wallet/src/features/addressDeletion/ForgetAddressButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const { t } = useTranslation()
+  const { hash: defaultAddressHash } = useAppSelector(selectDefaultAddress)
+  const isDefault = addressHash === defaultAddressHash
+
+  const deleteAddress = useDeleteAddress()
+  const onConfirmDelete = () => deleteAddress(addressHash)
```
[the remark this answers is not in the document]
nop33: Deleted in 1afea8f20

apps/desktop-wallet/src/features/addressDeletion/useDeleteAddress.ts
mvaivre: okay!

nop33:
@mvaivre are there any other changes required?

mvaivre (approved):
GJ 👏👏👏👏

PR #877, opened 2024-10-09
Redesign sent TX snackbar popup

mvaivre:


apps/desktop-wallet/src/features/send/sentTransactions/SentTransactionSnackbarPopup.tsx
```diff
@@ -64,20 +67,21 @@ const SentTransactionSnackbarPopup = memo(({ txHash }: SentTransactionSnackbarPo
...
+        <Progress status={sentTx.status} />
         <Rows>
-          <Message status={sentTx.status} />
-
-          <HashEllipsedStyled hash={txHash} tooltipText={t('Copy hash')} showSnackbarOnCopied={false} />
-
-          <Progress status={sentTx.status} />
+          <b>
+            <Message status={sentTx.status} />
+          </b>
```
nop33, line 74: nit: Let's use `font-weight` on the `Message`, WDYT?

apps/desktop-wallet/src/features/send/sentTransactions/SentTransactionSnackbarPopup.tsx
```diff
@@ -112,12 +119,21 @@ const Progress = ({ status }: Pick<SentTransaction, 'status'>) => {
     }
   }, [status])
 
-  useInterval(() => setProgress((prevValue) => prevValue + 0.015), 1000, status === 'confirmed' || progress > 0.9)
+  useInterval(
+    () => setProgress((prevValue) => Math.min(prevValue + 0.015, 1)),
```
nop33, line 123: I don't understand the purpose of this `Math.min`, since we stop it at 0.9?

nop33 (commented):
Nice job! A little bug I noticed

image

apps/desktop-wallet/src/features/send/sentTransactions/SentTransactionSnackbarPopup.tsx
mvaivre: The way you did it `Message` only returns a string, not an HTML component, so I didn't want to touch it.
But when I think about it again we should indeed return a component to follow our convention, or don't use uppercase on `Message`.

On it.

apps/desktop-wallet/src/features/send/sentTransactions/SentTransactionSnackbarPopup.tsx
mvaivre: You're right! Tried another approach, then forgot to remove the `min` function!

apps/desktop-wallet/src/features/send/sentTransactions/SentTransactionSnackbarPopup.tsx
mvaivre: Done

apps/desktop-wallet/src/features/send/sentTransactions/SentTransactionSnackbarPopup.tsx
mvaivre: Fixed

nop33 (approved):

PR #878, opened 2024-10-09
Add clickable NFT name and total worth in Send modal flow

### LINKED ISSUE #630: Show fiat amount before sending a transaction in desktop wallet

nop33:
In the context of closing #77 which is partially implemented in `next-tanstack` I would like to merge these small changes:

| before | after |
| ----- | ----- |
| image | image |

EDIT: I also added this feature that will make our users who asked for it happy:
- #630
- https://github.com/alephium/alephium-frontend/discussions/432

| before | after |
| ----- | ----- |
| image | image |

Closes #630

### DIFF
--- a/.changeset/late-squids-smoke.md
+++ b/.changeset/late-squids-smoke.md
@@ -1,0 +1,5 @@
+---
+"alephium-desktop-wallet": patch
+---
+
+Display total transaction worth when sending
--- a/.changeset/selfish-apricots-give.md
+++ b/.changeset/selfish-apricots-give.md
@@ -1,0 +1,5 @@
+---
+"alephium-desktop-wallet": patch
+---
+
+Display address derivation path in developer tools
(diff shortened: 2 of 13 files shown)


mvaivre:
Awesome. About the total worth, we should maybe nuance that a bit. It's the total worth of asset with a known price. 

We should probably redesign the asset amount rows to associate the fiat prices 🤔 
I have in mind something like: 
image

WDYT?

nop33:
I totally agree with you! I don't like these huge centered amounts that we currently display.

nop33:
What do you think about this version? I'm kinda digging the golden highlighted fiat value :)

| dark | light |
| ----- | ----- |
| image |  image |

mvaivre:
> What do you think about this version? I'm kinda digging the golden highlighted fiat value :)

I'm not a fan TBH 🙃 It puts to much emphasis on the fiat value (bold + highlight color!).
For me the most important data is the token amount. I think I would keep the fiat value in secondary color, without bold.

nop33:
Your wish is my command

| dark | light |
| ----- | ----- |
| image |  image |

PR #884, opened 2024-10-14
Remove app reload and update gcTime

nop33:
I have a compromise to propose:
1. We set **default** `gcTime: Infinity` so that we prevent data of locked wallets from being deleted from persisted storage
2. We set `gcTime: FIVE_MINUTES` to all queries whose `queryKey` can change
3. We call `queryClient.removeQueries` for the address-queries with `gcTime: Infinity` when a wallet or address gets deleted

What we end up with this approach is:
1. The balance queries will be preserved until wallet/address is deleted and will not be affected by gcTime when app is locked
2. Transaction queries will be cleared by gcTime when app is locked
3. Token metadata queries will be preserved indefinitely (or until user clears cache manually)

I think all the above is a good compromise.

### DIFF
--- a/apps/desktop-wallet/src/api/apiDataHooks/utils/useFetchFtList.ts
+++ b/apps/desktop-wallet/src/api/apiDataHooks/utils/useFetchFtList.ts
@@ -23,8 +23,6 @@ import axios from 'axios'
 
 import { useAppSelector } from '@/hooks/redux'
 
-const TOKEN_LIST_QUERY_KEY = 'tokenList'
-
 export interface FTList {
   data: TokenList['tokens'] | undefined
   isLoading: boolean
@@ -39,13 +37,14 @@ const useFetchFtList = (props?: FTListProps): FTList => {
   const network = networkId === 0 ? 'mainnet' : networkId === 1 ? 'testnet' : undefined
 
   const { data, isLoading } = useQuery({
-    queryKey: [TOKEN_LIST_QUERY_KEY, network],
+    queryKey: ['tokenList', network],
+    staleTime: ONE_DAY_MS,
+    gcTime: Infinity,
     queryFn:
       !network || props?.skip
         ? skipToken
         : () => axios.get(getTokensURL(network)).then(({ data }) => (data as TokenList)?.tokens),
-    placeholderData: network === 'mainnet' ? mainnet.tokens : network === 'testnet' ? testnet.tokens : [],
-    staleTime: ONE_DAY_MS
+    placeholderData: network === 'mainnet' ? mainnet.tokens : network === 'testnet' ? testnet.tokens : []
   })
 
   // TODO: Maybe return an object instead of an array for faster search?
--- a/apps/desktop-wallet/public/electron.js
+++ b/apps/desktop-wallet/public/electron.js
@@ -326,10 +326,6 @@ app.on('ready', async function () {
     deepLinkUri = null
   })
 
-  ipcMain.handle('app:reload', () => {
-    mainWindow.reload()
-  })
-
   createWindow()
 })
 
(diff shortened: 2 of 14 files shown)


nop33:
I was testing this and I noticed that overriding `gcTime` doesn't work when the default is set to `Infinity` :(

image

apps/desktop-wallet/src/api/apiDataHooks/address/useFetchAddressInfiniteTransactions.ts
```diff
@@ -31,18 +32,13 @@ interface UseFetchAddressInfiniteTransactionsProps {
 const useFetchAddressInfiniteTransactions = ({ addressHash }: UseFetchAddressInfiniteTransactionsProps) => {
   const networkId = useAppSelector((s) => s.network.settings.networkId)
 
-  const [fetchedTransactionListAt, setFetchedTransactionListAt] = useState(0)
-  const refresh = useCallback(() => setFetchedTransactionListAt(new Date().getTime()), [])
```
nop33, line 35: I simplified this a lot my removing the `timestamp` from the `queryKey`, it's much cleaner now.

apps/desktop-wallet/src/api/apiDataHooks/address/useFetchAddressInfiniteTransactions.ts
```diff
@@ -31,18 +32,13 @@ interface UseFetchAddressInfiniteTransactionsProps {
...
-      networkId,
-      skip: isLoadingLatestTx
-    })
-  )
+
+  const query = addressTransactionsInfiniteQuery({ addressHash, networkId, skip: isLoadingLatestTx })
+
+  const { data, fetchNextPage, isLoading, hasNextPage, isFetchingNextPage } = useInfiniteQuery(query)
+
+  const refresh = useCallback(() => queryClient.refetchQueries({ queryKey: query.queryKey }), [query.queryKey])
```
nop33, line 41: Instead of refreching by generating a new `timestamp`, we are now prefetching the proper way.

apps/desktop-wallet/src/api/apiDataHooks/address/useFetchAddressBalancesAlph.ts
```diff
@@ -21,9 +21,10 @@ import { useQuery } from '@tanstack/react-query'
 import { UseFetchAddressProps } from '@/api/apiDataHooks/address/addressApiDataHooksTypes'
 import { addressAlphBalancesQuery } from '@/api/queries/addressQueries'
 import { useAppSelector } from '@/hooks/redux'
+import { selectCurrentlyOnlineNetworkId } from '@/storage/settings/networkSelectors'
 
 const useFetchAddressBalancesAlph = ({ addressHash, skip }: UseFetchAddressProps) => {
-  const networkId = useAppSelector((s) => s.network.settings.networkId)
+  const networkId = useAppSelector(selectCurrentlyOnlineNetworkId)
```
nop33, line 27: this was hard to find out, but essentially, when switching networks the Redux state was being updated BEFORE the `throttledClient` was updated, leading to weird states (that were preserved in indexedDB) where `mainnet` endpoints had `testnet` data.

apps/desktop-wallet/src/api/apiDataHooks/utils/useFetchFtList.ts
```diff
@@ -35,17 +34,18 @@ type FTListProps = {
...
 const useFetchFtList = (props?: FTListProps): FTList => {
-  const networkId = useAppSelector((s) => s.network.settings.networkId)
+  const networkId = useAppSelector(selectCurrentlyOnlineNetworkId)
   const network = networkId === 0 ? 'mainnet' : networkId === 1 ? 'testnet' : undefined
 
   const { data, isLoading } = useQuery({
-    queryKey: [TOKEN_LIST_QUERY_KEY, network],
+    queryKey: ['tokenList', network],
+    staleTime: ONE_DAY_MS,
+    gcTime: Infinity,
```
nop33, line 43: I added `gcTime` in every query so that it's clear and so that we don't have to remember defaults.

PR #885, opened 2024-10-15
Custom Tanstack persister for each wallet

nop33:
Supersedes https://github.com/alephium/alephium-frontend/pull/884

1. Implements 1 persister per wallet and swaps them dynamically on wallet unlock. Only the cache of the currently unlocked wallet stays in memory.
2. No persister when the app is locked, everything is removed from memory
1. Removes app reloading
2. Fixes network switch
3. Persist only mainnet data

### DIFF
--- a/apps/desktop-wallet/src/api/apiDataHooks/address/useFetchAddressInfiniteTransactions.ts
+++ b/apps/desktop-wallet/src/api/apiDataHooks/address/useFetchAddressInfiniteTransactions.ts
@@ -18,10 +18,11 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import { AddressHash } from '@alephium/shared'
 import { useInfiniteQuery } from '@tanstack/react-query'
-import { useCallback, useMemo, useState } from 'react'
+import { useCallback, useMemo } from 'react'
 
 import useFetchAddressLatestTransaction from '@/api/apiDataHooks/address/useFetchAddressLatestTransaction'
 import { addressTransactionsInfiniteQuery } from '@/api/queries/transactionQueries'
+import queryClient from '@/api/queryClient'
 import { useAppSelector } from '@/hooks/redux'
 
 interface UseFetchAddressInfiniteTransactionsProps {
@@ -31,18 +32,13 @@ interface UseFetchAddressInfiniteTransactionsProps {
 const useFetchAddressInfiniteTransactions = ({ addressHash }: UseFetchAddressInfiniteTransactionsProps) => {
   const networkId = useAppSelector((s) => s.network.settings.networkId)
 
-  const [fetchedTransactionListAt, setFetchedTransactionListAt] = useState(0)
-  const refresh = useCallback(() => setFetchedTransactionListAt(new Date().getTime()), [])
-
   const { data: addressLatestTx, isLoading: isLoadingLatestTx } = useFetchAddressLatestTransaction({ addressHash })
-  const { data, fetchNextPage, isLoading, hasNextPage, isFetchingNextPage } = useInfiniteQuery(
-    addressTransactionsInfiniteQuery({
-      addressHash,
-      timestamp: fetchedTransactionListAt,
-      networkId,
-      skip: isLoadingLatestTx
-    })
-  )
+
+  const query = addressTransactionsInfiniteQuery({ addressHash, networkId, skip: isLoadingLatestTx })
+
+  const { data, fetchNextPage, isLoading, hasNextPage, isFetchingNextPage } = useInfiniteQuery(query)
+
+  const refresh = useCallback(() => queryClient.refetchQueries({ queryKey: query.queryKey }), [query.queryKey])
 
   const fetchedConfirmedTxs = useMemo(() => data?.pages.flat() ?? [], [data?.pages])
   const latestFetchedTxHash = fetchedConfirmedTxs[0]?.hash
--- a/apps/desktop-wallet/src/api/apiDataHooks/token/useFetchToken.ts
+++ b/apps/desktop-wallet/src/api/apiDataHooks/token/useFetchToken.ts
@@ -24,22 +24,26 @@ import { DataHook } from '@/api/apiDataHooks/apiDataHooksTypes'
 import useFetchNft from '@/api/apiDataHooks/token/useFetchNft'
 import useFetchFtList from '@/api/apiDataHooks/utils/useFetchFtList'
 import { fungibleTokenMetadataQuery, tokenTypeQuery } from '@/api/queries/tokenQueries'
+import { useAppSelector } from '@/hooks/redux'
+import { selectCurrentlyOnlineNetworkId } from '@/storage/settings/networkSelectors'
 import { ListedFT, NonStandardToken, TokenId, UnlistedFT } from '@/types/tokens'
 
 type UseFetchTokenResponse = DataHook<ListedFT | UnlistedFT | NFT | NonStandardToken>
 
 const useFetchToken = (id: TokenId): UseFetchTokenResponse => {
   const { data: fTList, isLoading: isLoadingFtList } = useFetchFtList()
+  const networkId = useAppSelector(selectCurrentlyOnlineNetworkId)
 
   const listedFT = fTList?.find((t) => t.id === id)
 
   const { data: tokenType, isLoading: isLoadingTokenType } = useQuery(
-    tokenTypeQuery({ id, skip: isLoadingFtList || !!listedFT })
+    tokenTypeQuery({ id, networkId, skip: isLoadingFtList || !!listedFT })
   )
 
   const { data: unlistedFT, isLoading: isLoadingUnlistedFT } = useQuery(
     fungibleTokenMetadataQuery({
       id,
+      networkId,
       skip: isLoadingTokenType || tokenType?.stdInterfaceId !== explorer.TokenStdInterfaceId.Fungible
     })
   )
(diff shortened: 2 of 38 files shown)


apps/desktop-wallet/src/api/apiDataHooks/address/useFetchAddressBalancesAlph.ts
```diff
@@ -21,9 +21,10 @@ import { useQuery } from '@tanstack/react-query'
 import { UseFetchAddressProps } from '@/api/apiDataHooks/address/addressApiDataHooksTypes'
 import { addressAlphBalancesQuery } from '@/api/queries/addressQueries'
 import { useAppSelector } from '@/hooks/redux'
+import { selectCurrentlyOnlineNetworkId } from '@/storage/settings/networkSelectors'
 
 const useFetchAddressBalancesAlph = ({ addressHash, skip }: UseFetchAddressProps) => {
-  const networkId = useAppSelector((s) => s.network.settings.networkId)
+  const networkId = useAppSelector(selectCurrentlyOnlineNetworkId)
```
nop33, line 27: this was hard to find out, but essentially, when switching networks the Redux state was being updated BEFORE the `throttledClient` was updated, leading to weird states (that were preserved in indexedDB) where `mainnet` endpoints had `testnet` data.

apps/desktop-wallet/src/api/apiDataHooks/address/useFetchAddressInfiniteTransactions.ts
```diff
@@ -31,18 +32,13 @@ interface UseFetchAddressInfiniteTransactionsProps {
 const useFetchAddressInfiniteTransactions = ({ addressHash }: UseFetchAddressInfiniteTransactionsProps) => {
   const networkId = useAppSelector((s) => s.network.settings.networkId)
 
-  const [fetchedTransactionListAt, setFetchedTransactionListAt] = useState(0)
-  const refresh = useCallback(() => setFetchedTransactionListAt(new Date().getTime()), [])
```
nop33, line 35: I simplified this a lot my removing the `timestamp` from the `queryKey`, it's much cleaner now.

apps/desktop-wallet/src/api/apiDataHooks/address/useFetchAddressInfiniteTransactions.ts
```diff
@@ -31,18 +32,13 @@ interface UseFetchAddressInfiniteTransactionsProps {
...
-      networkId,
-      skip: isLoadingLatestTx
-    })
-  )
+
+  const query = addressTransactionsInfiniteQuery({ addressHash, networkId, skip: isLoadingLatestTx })
+
+  const { data, fetchNextPage, isLoading, hasNextPage, isFetchingNextPage } = useInfiniteQuery(query)
+
+  const refresh = useCallback(() => queryClient.refetchQueries({ queryKey: query.queryKey }), [query.queryKey])
```
nop33, line 41: Instead of refreching by generating a new timestamp, we are now prefetching the proper way.

apps/desktop-wallet/src/api/apiDataHooks/token/useFetchToken.ts
```diff
@@ -24,22 +24,26 @@ import { DataHook } from '@/api/apiDataHooks/apiDataHooksTypes'
...
 
   const { data: tokenType, isLoading: isLoadingTokenType } = useQuery(
-    tokenTypeQuery({ id, skip: isLoadingFtList || !!listedFT })
+    tokenTypeQuery({ id, networkId, skip: isLoadingFtList || !!listedFT })
   )
 
   const { data: unlistedFT, isLoading: isLoadingUnlistedFT } = useQuery(
     fungibleTokenMetadataQuery({
       id,
+      networkId,
```
nop33, line 46: I added the network ID to each query so that we can skip persisting in storage testnet/devnet queries.

apps/desktop-wallet/src/api/apiDataHooks/utils/useFetchFtList.ts
```diff
@@ -35,17 +34,21 @@ type FTListProps = {
...
+  const networkId = useAppSelector(selectCurrentlyOnlineNetworkId)
   const network = networkId === 0 ? 'mainnet' : networkId === 1 ? 'testnet' : undefined
 
   const { data, isLoading } = useQuery({
-    queryKey: [TOKEN_LIST_QUERY_KEY, network],
+    queryKey: ['tokenList', network],
+    staleTime: ONE_DAY_MS,
+    // The token list is essential for the whole app so we don't want to ever delete it. Even if we set a lower gcTime,
+    // it will never become inactive (since it's always used by a mount component).
+    gcTime: Infinity,
```
nop33, line 45: I added comments above each use of `gcTime` so that it's clear now and in the future as to why we chose these values.

apps/desktop-wallet/src/api/persistQueryClientContext.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export const initialPersistQueryClientContext: PersistQueryClientContextType = {
+  restoreQueryCache: () => Promise.resolve(),
+  clearQueryCache: () => {}
+}
+
+export const PersistQueryClientContext = createContext<PersistQueryClientContextType>(initialPersistQueryClientContext)
+
+// Inspired by https://github.com/TanStack/query/blob/1adaf3ff86fa2bf720dbc958714c60553c4aae08/packages/react-query-persist-client/src/PersistQueryClientProvider.tsx
+export const PersistQueryClientContextProvider = ({ children }: { children: ReactNode | ReactNode[] }) => {
```
mvaivre, line 53: Isn't `children: ReactNode` enough? 🤔

apps/desktop-wallet/src/api/queries/addressQueries.ts
```diff
@@ -33,21 +33,27 @@ export type AddressAlphBalancesQueryFnData = {
 export const addressAlphBalancesQuery = ({ addressHash, networkId, skip }: AddressLatestTransactionQueryProps) =>
   queryOptions({
     queryKey: ['address', addressHash, 'balance', 'ALPH', { networkId }],
```
mvaivre, line 35: Nit: Here we pass `networkId` in an object, whereas in the PR you added it in one of the first keys in the array. We should maybe always do it the same way.

apps/desktop-wallet/src/api/queries/tokenQueries.ts
```diff
@@ -78,29 +83,37 @@ export const combineTokenTypeQueryResults = (results: UseQueryResult<TokenInfo |
...
 
-export const nftMetadataQuery = ({ id, skip }: TokenQueryProps) =>
+export const nftMetadataQuery = ({ id, networkId, skip }: TokenQueryProps) =>
   queryOptions({
     queryKey: ['token', 'non-fungible', 'metadata', id],
-    queryFn: !skip ? async () => (await batchers.nftMetadataBatcher.fetch(id)) ?? null : skipToken,
-    staleTime: Infinity
+    staleTime: Infinity,
+    gcTime: Infinity, // We don't want to delete the NFT metadata when the user navigates away from NFT components.
+    meta: { isMainnet: networkId === 0 },
```
mvaivre, line 107: I'm not a huge fan of this. It's not super DRY. And what if we forget it?

mvaivre (commented):
GJ!! And thanks again for the explanations before, it helped.

apps/desktop-wallet/src/api/apiDataHooks/utils/useFetchFtList.ts
```diff
@@ -35,17 +34,21 @@ type FTListProps = {
 }
 
 const useFetchFtList = (props?: FTListProps): FTList => {
-  const networkId = useAppSelector((s) => s.network.settings.networkId)
+  const networkId = useAppSelector(selectCurrentlyOnlineNetworkId)
   const network = networkId === 0 ? 'mainnet' : networkId === 1 ? 'testnet' : undefined
 
   const { data, isLoading } = useQuery({
-    queryKey: [TOKEN_LIST_QUERY_KEY, network],
+    queryKey: ['tokenList', network],
```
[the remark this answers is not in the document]
nop33: No reason at all, this was written as one of the first things. Do you want me to change it?

apps/desktop-wallet/src/api/persistQueryClientContext.tsx
nop33: You're right, I'll remove it.

apps/desktop-wallet/src/api/queries/addressQueries.ts
nop33: Where did I add it as one of the first keys in the array?

apps/desktop-wallet/src/api/queries/tokenQueries.ts
nop33: How do you recommend we do it?

nop33: If we forget it, the data will be persisted. That's the worst case

nop33: I wanted to use the `networkId` from the `queryKey`, but each query can have it in a different index in the array... I thought a more standardized way would be to use `meta`.

apps/desktop-wallet/src/api/apiDataHooks/utils/useFetchFtList.ts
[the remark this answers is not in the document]
nop33: Done in cdc166a6e

apps/desktop-wallet/src/api/queries/tokenQueries.ts
nop33: Would you prefer a function

```ts
export const generateNetworkQueryMeta = (networkId?: number) => ({
  isMainnet: networkId === 0
})
```

and then using it like so:

```ts
const { data, isLoading } = useQuery({
    queryKey: ['tokenList', { networkId }],
    staleTime: ONE_DAY_MS,
    gcTime: Infinity,
    meta: generateNetworkQueryMeta(networkId),
```

nop33:
Approved over Slack, merging.

PR #886, opened 2024-10-15
Improve tanstack perf

nop33:


apps/desktop-wallet/src/api/apiDataHooks/market/useFetchTokenPrices.ts
```diff
@@ -42,10 +41,15 @@ const useFetchTokenPrices = (props?: SkipProp) => {
 export default useFetchTokenPrices
 
 export const useFetchTokenPrice = (symbol: string) => {
-  const { data, isLoading } = useFetchTokenPrices()
+  const fiatCurrency = useAppSelector((s) => s.settings.fiatCurrency)
+
+  const { data, isLoading } = useQuery({
+    ...tokensPriceQuery({ symbols: pricedTokens, currency: fiatCurrency.toLowerCase() }),
+    select: (data) => data.find((tokenPrice) => tokenPrice.symbol === symbol)?.price
```
mvaivre, line 48: Hey, didn't know about this one!

mvaivre (approved):
Looks great! 👌

apps/desktop-wallet/src/api/apiDataHooks/market/useFetchTokenPrices.ts
nop33: yeah, and if we want to go more crazy about it, we can use `reselect` to extract the inline function into a selector with `createSelector`.

Issue #865, opened 2024-10-04
App crashes in fresh install on Samsung Z Fold 6


nop33:
The app closes as soon as the user clicks it. No error messages. No crash reports in Posthog.

Stock android.

Tried:
- updating
- force close
- reinstall
- rebooted
- delete app data

nop33:
This is one of the stacktraces that affects most of our users (Android)

```
Exception com.facebook.react.common.JavascriptException:
  at com.facebook.react.modules.core.ExceptionsManagerModule.reportException (ExceptionsManagerModule.java:65)
  at java.lang.reflect.Method.invoke
  at com.facebook.react.bridge.JavaMethodWrapper.invoke (JavaMethodWrapper.java:372)
  at com.facebook.react.bridge.JavaModuleWrapper.invoke (JavaModuleWrapper.java:146)
  at com.facebook.jni.NativeRunnable.run
  at android.os.Handler.handleCallback (Handler.java:959)
  at android.os.Handler.dispatchMessage (Handler.java:100)
  at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage (MessageQueueThreadHandler.java:27)
  at android.os.Looper.loopOnce (Looper.java:232)
  at android.os.Looper.loop (Looper.java:317)
  at com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run (MessageQueueThreadImpl.java:233)
  at java.lang.Thread.run (Thread.java:1012)
```

This is the chatGPT suggests:

<img width="663" alt="image" src="https://github.com/user-attachments/assets/f0279e85-d943-4263-89e6-1e66d97fdf6b">
<img width="651" alt="image" src="https://github.com/user-attachments/assets/e1b946f9-c1c0-41a8-8157-ae3e31eeb370">
<img width="636" alt="image" src="https://github.com/user-attachments/assets/10b5d5b3-7ccf-420e-8054-e585240c03e3">

nop33:
Genimi suggests

<img width="1327" alt="image" src="https://github.com/user-attachments/assets/095ef640-1164-4b68-bba7-876cdca1cd1b">

PR #888, opened 2024-10-16
Set up Sentry in mobile wallet

nop33:
Trying to debug:
- https://github.com/alephium/alephium-frontend/issues/865

Most of the diff is auto-generated after installing Sentry and running the wizard.

Also run `expo-doctor` to clean up some deps versions, as advised by our AI friends here: #865


apps/mobile-wallet/ios/Alephium/Info.plist
```diff
@@ -91,4 +91,4 @@
     <key>UIViewControllerBasedStatusBarAppearance</key>
     <false/>
   </dict>
-</plist>
+</plist>
```
mvaivre, line 94: That's weird, there are multiple new lines EOF across the PR, but I guess we don't care :)

mvaivre (approved):
LGTM! 👁️

PR #894, opened 2024-10-18
Post 2.4.0 fixes

nop33:


apps/desktop-wallet/src/features/walletConnect/WalletConnectSessionProposalModal.tsx
```diff
@@ -272,7 +272,7 @@ const WalletConnectSessionProposalModal = memo(
                 <Trans
                   t={t}
                   i18nKey="walletConnectNewAddress"
-                  values={{ group: chainInfo.addressGroup }}
+                  values={{ currentNetworkName }}
```
nop33, line 275: random small bug I noticed when testing this PR

apps/desktop-wallet/src/App.tsx
```diff
@@ -219,3 +222,20 @@ const AppContainerStyled = styled.div<{ showDevIndication: boolean }>`
       border: 5px solid ${theme.global.valid};
     `};
 `
+
+const useClearQueryPersistedCacheOnVersionUpdate = () => {
```
mvaivre, line 226: Is there a reason why this is not extracted out of `App.tsx` like the other hooks?

mvaivre: Also, nit rename : `useClearPersistedQueryCacheOnVersionUpdate()`. If we say that it's "persisted", we don't need to say that it's a "cache". WDYT?

apps/desktop-wallet/src/api/persistedCacheVersionStorage.ts
```diff
@@ -0,0 +1,37 @@
...
+The library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+export type PersistedCacheVersion = string
```
mvaivre, line 19: Nit: I would always specify that we're talking about the `PersistedQueryCache`, like you did in the related hooks and other functions.
We may persist other caches in the future.

apps/desktop-wallet/src/features/walletConnect/walletConnectContext.tsx
```diff
@@ -739,9 +519,13 @@ export const WalletConnectContextProvider: FC = ({ children }) => {
         sendUserRejectedResponse,
         sendSuccessResponse,
         sendFailureResponse,
-        refreshActiveSessions
+        refreshActiveSessions,
+        respondToWalletConnectWithError
       }}
     >
+      {sessionRequestEvent && isWalletUnlocked && (
+        <WalletConnectSessionRequestEventHandler sessionRequestEvent={sessionRequestEvent} />
```
mvaivre, line 527: I understand why you created a component instead of a hook for this (to avoid running anything if it's not needed), but this highlights when of the limitation of React that is a bit frustrating to me: the rules of hooks. Anyway, no need to comment, just a little rant from me 😆

mvaivre (approved):
Code looks good to me. Only nit comments so I approve!
Let's make an rc, test like crazy (the whole testing suite, I'll go through it thoroughly on Monday morning), and release 🚀

apps/desktop-wallet/src/features/walletConnect/walletConnectContext.tsx
nop33: Yes, exactly 💯

apps/desktop-wallet/src/App.tsx
nop33: What do you mean? All hooks used in the App.tsx are in the App.tsx file

apps/desktop-wallet/src/api/persistedCacheVersionStorage.ts
nop33: Done ✅

apps/desktop-wallet/src/App.tsx
nop33: Rename done and pushed.

PR #905, opened 2024-10-21
Enforce svgs to be imported as paths

mvaivre:


mvaivre:
Seems like this isn't needed.

PR #909, opened 2024-10-22
Fix Sentry reported bugs

nop33:


apps/mobile-wallet/setupErrorHandling.ts
```diff
@@ -29,11 +29,13 @@ if (!__DEV__) {
...
     Sentry.captureException(error, { data: { isFatal } })
 
+    const url = `mailto:developer@alephium.org?subject=Crash report&body=${error}`
+
     Alert.alert('An error occurred', 'We are working on fixing this.', [
       { text: 'Cancel', style: 'cancel' },
       {
         text: 'Send report',
-        onPress: () => Linking.openURL(`mailto:developer@alephium.org?subject=Crash report&body=${error}`)
+        onPress: () => Linking.canOpenURL(url).then((result) => (result ? Linking.openURL(url) : false))
```
mvaivre, line 38: 👍

apps/mobile-wallet/src/persistent-storage/wallet.ts
```diff
@@ -273,6 +279,15 @@ export const getStoredWalletMetadata = async (error?: string): Promise<WalletMet
   return walletMetadata
 }
 
+export const isStoredWalletMetadataMigrated = (
```
mvaivre, line 282: It's confusing to me that we mix `walletMetada` and `addressMetadata` like that. Is that correct?

This feels wrong:
- `addressMetadata as WalletMetadata`
- `addressMetadata is WalletMetadata`

mvaivre (approved):

PR #913, opened 2024-10-22
Internationalize desktop wallet amounts based on user's region or choice

nop33:
Supersedes:
- https://github.com/alephium/alephium-frontend/pull/879

Related to:
- #128

Closes https://github.com/alephium/alephium-frontend/issues/921:
- https://github.com/alephium/alephium-frontend/issues/921

This PR improves amounts display in desktop wallet:
1. Adds a new settings field called `numberFormatRegion`
2. Defaults to the user's region (using electron's `app.getSystemLocale()`)
3. Allows user to change it in Settings modal
4. Uses `Intl.NumberFormat` with `numberFormatRegion` and `fiatCurrency` to improve amounts display


apps/desktop-wallet/src/features/settings/numberFormatLocales.ts
```diff
@@ -0,0 +1,49 @@
...
+  { label: 'ES', value: 'es-ES' },
+  { label: 'FR', value: 'fr-FR' },
+  { label: 'ID', value: 'id-ID' },
+  { label: 'IT', value: 'it-IT' },
+  { label: 'PT', value: 'pt-PT' },
+  { label: 'RU', value: 'ru-RU' },
+  { label: 'TR', value: 'tr-TR' },
+  { label: 'VM', value: 'vi-VN' },
+  { label: 'GR', value: 'el-GR' },
+  { label: 'CH', value: 'de-CH' }
```
mvaivre, line 48: Nit: How is this ordered? I was surprised to see CH at the very end. Should we use alphabetic order?

apps/desktop-wallet/src/storage/settings/settingsPersistentStorage.ts
```diff
@@ -31,7 +31,8 @@ export const defaultSettings: Settings = {
     language: undefined,
     devTools: false,
     analytics: true,
-    fiatCurrency: 'USD'
+    fiatCurrency: 'USD',
```
mvaivre, line 34: This makes me think: because we know the locale, we could also set the default currency according to it. WDYT?

mvaivre (commented):
Code and functional review look good!
However I have some more general comments:
- I'm not sure about the need for a dedicated setting for this. It complicates the settings modal and I'd like to avoid this whenever possible. I'd link the formatting setting to the user "region"
- You've called the setting `numberFormatRegion` but it could be renamed as simply "formatRegion" as it could  impact more than just numbers. I'm thinking about dates, times etc. too.

Edit: I've taken a look at Ledger Live and they go with the solution I had in mind: they have a "region" setting, same as in most OSes too. This settings change the way "dates, time and currencies" are displayed.👇 

image

They have settings for currency, language, and region.

apps/desktop-wallet/src/features/settings/numberFormatLocales.ts
nop33: Let's use alphabetic, but maybe we should use the most frequent ones at the top?

apps/desktop-wallet/src/storage/settings/settingsPersistentStorage.ts
nop33: Agreed! I will look into it.

nop33:
So, in your https://github.com/alephium/alephium-frontend/pull/913#pullrequestreview-2398396114 comment do you propose renaming it to "Region" and that's it?

apps/desktop-wallet/src/storage/settings/settingsPersistentStorage.ts
nop33: Hm, not sure how this is supposed to work. Having the locale does not give us the currency people use in that region, but rather the option to format a number by providing the currency we want.

```js
1000000.toLocaleString("en-US", {style:"currency", currency:"USD"});
```

nop33: We would need an extrernal library of sorts, to provide a mapping between currencies and locales.

apps/desktop-wallet/src/features/settings/numberFormatLocales.ts
nop33: Pushed

nop33:
I renamed the settings as requested!

nop33:
btw, I found the corresponding code for the screenshot that you shared in the ledger app:
- https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/renderer/screens/settings/sections/General/RegionSelect.tsx#L41
- https://github.com/LedgerHQ/ledger-live/blob/develop/apps/ledger-live-desktop/src/renderer/screens/settings/sections/General/regions.json

mvaivre:
I get an error, probably following the rename: `Error: Error invoking remote method 'app:getSystemRegion': No handler registered for 'app:getSystemRegion'`

mvaivre:
> Hm, no, that was not renamed... I can't reproduce it, but let me push my latest changes, maybe it's fixed.
Seems to be fixed now yes!

apps/desktop-wallet/src/components/SnackbarManager.tsx
```diff
@@ -56,14 +56,10 @@ const SnackbarPopup = memo(({ message }: { message: Required<SnackbarMessage> })
 
   // Remove snackbar popup after its duration
   useEffect(() => {
-    let timer: ReturnType<typeof setTimeout>
-
     if (message && message.duration >= 0) {
-      timer = setTimeout(() => dispatch(snackbarDisplayTimeExpired()), message.duration)
-    }
+      const timer = setTimeout(() => dispatch(snackbarDisplayTimeExpired()), message.duration)
```
mvaivre, line 60: We sometimes use the custom `sleep` function from web3 (which is basically a handle for `setTimeout`), but we usually use `setTimeout` directly. 
The sleep function can be a bit confusing IMO because it feels like it could stop the whole execution (like in other languages). But I like how concise it is compared to `setTimeout`.
However the most important point:
Wit `sleep` we can't clear the timeout like we do here. This may be an issue. WDYT?

mvaivre (approved):

PR #922, opened 2024-10-23
Update release docs

### LINKED ISSUE #188: Review RC and release workflow with `changeset`

nop33:
Closes #188 
- #188

### DIFF
--- a/apps/desktop-wallet/README.md
+++ b/apps/desktop-wallet/README.md
@@ -24,3 +24,31 @@ pnpm ci:build:electron:macOS
 pnpm ci:build:electron:linux
 pnpm ci:build:electron:linux:arm64
 ```
+
+## Releasing
+
+### Releasing a release candidate version
+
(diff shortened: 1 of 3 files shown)


mvaivre (approved):
🫶

PR #925, opened 2024-10-24
Improve NFT display support (inspired by explorer)

mvaivre:


apps/desktop-wallet/src/features/thumbnails/thumbnailStorage.ts
```diff
@@ -0,0 +1,204 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+const DB_NAME = 'VideoThumbnailsDB'
+const STORE_NAME = 'thumbnails'
```
nop33, line 20: We should probably delete this cache as well when the user deletes the caches from the settings. WDYT?

mvaivre: You're right yep!

apps/desktop-wallet/src/features/thumbnails/thumbnailStorage.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  const generatedThumbnailBlob = await getOrCreateThumbnail(videoUrl)
+  const generatedThumbnailUrl = URL.createObjectURL(generatedThumbnailBlob)
+
+  saveThumbnailToDB(videoUrl, generatedThumbnailBlob)
+
+  return generatedThumbnailUrl
+}
+
+export const deleteThumbnailsDB = async () => indexedDB.deleteDatabase(DB_NAME)
```
nop33, line 206: `deleteDatabase` is not async

apps/desktop-wallet/src/modals/SettingsModal/GeneralSettingsSection.tsx
```diff
@@ -127,6 +128,7 @@ const GeneralSettingsSection = ({ className }: GeneralSettingsSectionProps) => {
   const handleClearCacheButtonPress = async () => {
     try {
       queryClient.clear()
+      deleteThumbnailsDB()
```
nop33, line 131: are we sure we want to bundle this together with the tanstack query cache? We did not add the walletconnect cache clearing together with the tanstack because one might fail whereas the other might not.

apps/desktop-wallet/src/features/thumbnails/thumbnailStorage.ts
mvaivre: Fixed

nop33 (approved):

PR #932, opened 2024-10-29
Address worth improvements/fixes

mvaivre:


apps/explorer/src/api/assets/assetsHooks.ts
```diff
@@ -172,9 +173,7 @@ export const useAssetsMetadata = (assetIds: string[] = []) => {
   } else return returnedCompleteMetadata
 }
 
-// We should get this list from the backend
-// See: https://github.com/alephium/explorer-backend/issues/512
-export const useTokensWithAvailablePrice = () => ['ALPH', 'USDT', 'USDC', 'DAI', 'WBTC', 'WETH', 'AYIN']
+export const useTokensWithAvailablePrice = () => Object.values(explorer.TokensWithPrice) as string[]
```
nop33, line 176: I don't see why this should be a hook.

On desktop wallet we do this:

```
const pricedTokens = Object.keys(explorer.TokensWithPrice)
```

Maybe we can simply export this from `@alephium/shared` instead?

nop33 (commented):
As discussed, let's make the `decimals` prop mandatory in `calculateAmountWorth` in shared

apps/explorer/src/api/assets/assetsHooks.ts
mvaivre: Yep, let's do that.

apps/explorer/src/pages/AddressInfoPage/index.tsx
```diff
@@ -197,7 +198,7 @@ const AddressInfoPage = () => {
             }
           />
           <InfoGrid.Cell
-            label={t('Fiat price')}
```
nop33, line 200: The string "Fiat price" should be removed from the translation.json file

apps/explorer/src/pages/AddressInfoPage/index.tsx
```diff
@@ -141,7 +141,8 @@ const AddressInfoPage = () => {
   const lockedBalance = addressBalance?.lockedBalance
 
   const addressWorth =
-    knownTokensWorth + (totalBalance ? calculateAmountWorth(BigInt(totalBalance), tokensPrices[ALPH.symbol] || NaN) : 0)
+    knownTokensWorth +
+    (totalBalance ? calculateAmountWorth(BigInt(totalBalance), tokensPrices[ALPH.symbol] || NaN, ALPH.decimals) : 0)
```
nop33, line 145: Let's update the `calculateAmountWorth` in @alephium/shared to require the `decimals` prop (instead of optional)

apps/explorer/src/pages/AddressInfoPage/index.tsx
mvaivre: Yeah I asked myself about this, thinking it may be good to keep translations for a pretty generic string that may be reused somewhere else. 
But I guess it's indeed better to keep the translation files clean and in sync with what is actually used in-app.

apps/explorer/src/pages/AddressInfoPage/index.tsx
mvaivre: Done

apps/explorer/src/pages/AddressInfoPage/index.tsx
mvaivre: Done

nop33: It's not only that, but it reduces the work of translators.

apps/explorer/locales/de-DE/translation.json
```diff
@@ -20,7 +20,6 @@
   "Copy token ID": "Token ID kopieren",
   "Daily": "täglich",
   "Download CSV": "CSV herunterladen",
-  "Fiat price": "Fiat Preis",
```
nop33, line 23: For future references: You only have to delete it from the English file. Our automation system will take care of deleting it from the other files.

nop33 (approved):

PR #986, opened 2024-11-20
Simplify landing screen

mvaivre:


apps/mobile-wallet/src/screens/LandingScreen.tsx
```diff
@@ -15,73 +15,41 @@ GNU Lesser General Public License for more details.
...
       storedWalletExists()
         .then((walletExists) => {
           if (walletExists) {
-            // Normally, when the app is unlocked, this screen is not in focus. However, under certain conditions we end
-            // up with an unlocked wallet and no screen in focus at all. This happens when:
-            // 1. the auto-lock is set to anything but "Fast"
-            // 2. the user manually kills the app before the auto-lock timer completes
-            // 3. the WalletConnect feature is activated
-            // Since there is no screen in focus and since the default screen set in the RootStackNavigation is this
-            // screen, we need to navigate back to the dashboard.
```
nop33, line 80: why removing all these comments? I think it's important to leave them there as it is not self-evident what is happening here by just reading the code.

apps/mobile-wallet/src/screens/LandingScreen.tsx
```diff
@@ -15,73 +15,41 @@ GNU Lesser General Public License for more details.
...
             if (isWalletUnlocked) resetNavigation(navigation)
           } else {
-            // Only display this screen's contents when we have no stored wallet. If there is a wallet, the
-            // AppUnlockModal will be displayed
```
nop33, line 84: Same for here.

apps/mobile-wallet/src/navigation/RootStackNavigation.tsx
```diff
@@ -275,14 +266,8 @@ const AppUnlockModal = ({ initialRouteName }: Required<RootStackNavigationProps>
   useAutoLock(unlockApp)
 
   return (
-    <Modal
-      visible={!!lastUsedWalletId && biometricsRequiredForAppAccess && !isWalletUnlocked}
-      onLayout={handleScreenLayoutChange}
-      animationType="none"
-    >
-      <View style={{ backgroundColor: 'black', flex: 1 }}>
-        <CoolAlephiumCanvas {...dimensions} onPress={unlockApp} />
```
nop33, line 284: The problem with removing the canvas is that we lose the feature of re-triggering the unlock flow `onPress`.

nop33: If a user cancels the authentication for whatever reason, the authentication modal goes away and they have no way of triggering it. That's why we had the `onPress` handler that apart from the cool animation of flipping the Alephium logo, it also re-triggered the authentication flow.

mvaivre: Good point, I can add the onPress handler on the black view (and add an alephium logo). But the allowed me to find a bug.. In the following modal, if I tap on try again nothing happens. If I tap on cancel, it shows an error but let me try again :D 

image

Cancel:
image

mvaivre: Tap to unlock re-add here 844069d9983fdbdcdd4538c43800de991acc8c60

nop33: I sent you a msg on Slack on how to potentially solve this build error

nop33 (commented):
Code looks fine. I'll check out the branch to test it now.

apps/mobile-wallet/src/screens/TransactionModal.tsx
```diff
@@ -74,40 +78,50 @@ const TransactionModal = ({ tx, ...props }: TransactionModalProps) => {
...
+        )}
+        {!isMoved && groupedIOAmounts.out && (
+          <Row title={t('Sent')} transparent isVertical titleColor={theme.global.send}>
+            <AmountsContainer>
+              {groupedIOAmounts.out.map(({ id, amount }) => (
+                <AssetAmountWithLogo key={id} assetId={id} amount={amount} logoSize={18} />
+              ))}
+            </AmountsContainer>
+          </Row>
+        )}
```
nop33, line 107: I would swap the order of these 2: First "Sent" and then "Received". When using dApps, it makes more sense: You send something to receive something.

Besides, if a tx includes negative amounts it means that the user has initiated the transaction. Thus, I believe, the approved amount of the user (aka: "Sent") should be displayed first, and then what they got in return for what they sent.

| I provide liquidity with these 3 tokens and I receive 0.56 LP tokens in return | I send 0.11 ALPH to swap it for 0.10 EX |
| ----- | ----- |
|  |  |

PR #993, opened 2024-11-21
Fix display of ALPH amounts

### LINKED ISSUE #981: Total balances do not reflect amounts in transactions list
@h0ngcha0 reported to me today that he and a friend of his experienced the problem that even though the desktop wallet displays a list of incoming confirmed transactions, the total balance of the wallet in the overview page remains `0`.

I've seen a similar issue on Telegram where a user complained that even though their transaction is displayed in the transaction list as well as on our explorer, the balance of their wallet has not updated, even after waiting for 45 minutes. @diomark has experienced this as well.

nop33:
Closes #981

### DIFF
--- a/.changeset/dirty-candles-flash.md
+++ b/.changeset/dirty-candles-flash.md
@@ -1,0 +1,5 @@
+---
+"alephium-desktop-wallet": patch
+---
+
+Fix displayed ALPH amounts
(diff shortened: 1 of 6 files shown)


mvaivre (approved):
👍 lgtm!

PR #995, opened 2024-11-22
#963 Add warning to notify users with zero balance.

hugoleung1225:
Feature updates for #963 point 1.

Issue #963, opened 2024-11-16
No information for users if clicked "Send" button with zero balance


hugoleung1225:
**Is your feature request related to a problem? Please describe.**
For desktop-wallet, when user clicks "Send" Shortcuts, if user has zero balance, no any warning / suggested actions to
guide user what's next.

**Describe the solution you'd like**
Suggestion :
1. For all network type, shows a popup to remind user that send action cannot perform if balance = 0.
2. For devnet  / testnet , add a button to redirect to faucet.

<img width="458" alt="shortCut_Send" src="https://github.com/user-attachments/assets/f3ae59cc-28b0-44b7-b01e-5f4ba8190517">

Issue #642, opened 2024-06-10
Alephium Wallet not sending


bullfx:
Hi, My Wallet is not openning send dialog it was working until yesterday

mvaivre:
Is this still an issue @bullfx?

bullfx:
found a way...may be because I had 2 addressess but main link is still not oppening send window


    On Monday, 15 July 2024 at 11:45:39 pm AWST, Mikaël Vaivre ***@***.***> wrote:  
 
 


Is this still an issue @bullfx?

PR #997, opened 2024-11-24
Disable send btn when wallet empty

nop33:
Closes:
- #963
- #642
- #995

## Connected to mainnet

image
image
image

## Connected to testnet/devnet

image

### DIFF
--- a/.changeset/great-gorillas-itch.md
+++ b/.changeset/great-gorillas-itch.md
@@ -1,0 +1,5 @@
+---
+"alephium-desktop-wallet": patch
+---
+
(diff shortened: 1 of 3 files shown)


mvaivre:
Thanks! Tbh I'm not sure that I rly like tooltips for long texts like this, but let's go with that for now. I'll rebase on this and see if I can find another solution!

mvaivre (approved):

nop33:
@mvaivre check the solution that a contributor proposed here: https://github.com/alephium/alephium-frontend/pull/995

Issue #999, opened 2024-11-25
Use `latest-transaction` instead of `transactions?page=1&limit=1`


nop33:
- [ ] mobile
- [x] desktop

PR #1000, opened 2024-11-25
Use `latest-transaction` endpoint in DW

nop33:
Let's wait for a new web3 version to be released where the explorer backend endpoint does not return a `404` error when the queried address does not have any transaction and instead it returns `null` or `undefined`.

cc @tdroxler 

Related to:
- #999


nop33:
Done in https://github.com/alephium/alephium-frontend/pull/1301.
Closing.

PR #1012, opened 2024-11-28
Forget address feature in mobile wallet

nop33:
- #631

### DIFF
--- a/apps/mobile-wallet/src/features/address-deletion/useCanDeleteAddress.tsx
+++ b/apps/mobile-wallet/src/features/address-deletion/useCanDeleteAddress.tsx
@@ -24,7 +24,7 @@ import { selectAddressByHash } from '~/store/addressesSlice'
 const useCanDeleteAddress = (addressHash: AddressHash) => {
   const address = useAppSelector((s) => selectAddressByHash(s, addressHash))
 
-  return address && !address.settings.isDefault && address.index !== 0
+  return address && !address.settings.isDefault
 }
 
 export default useCanDeleteAddress
--- a/apps/mobile-wallet/src/features/address-deletion/useCanDeleteAddress.tsx
+++ b/apps/mobile-wallet/src/features/address-deletion/useCanDeleteAddress.tsx
@@ -1,0 +1,30 @@
+/*
+Copyright 2018 - 2024 The Alephium Authors
+This file is part of the alephium project.
+
+The library is free software: you can redistribute it and/or modify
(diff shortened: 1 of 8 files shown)


mvaivre:
For now I've noticed one issue while testing: the delete button isn't shown after a default address change:


https://github.com/user-attachments/assets/51b1fddd-7e9e-40e4-8dae-043316382801

mvaivre:
In terms of UI, this will be redesigned so I won't insist too much, but it start to be pretty crowded up here with all those buttons:
image

Let's:
- Remove the copy address button. We can activate long press on the address badge to copy instead.
- Let's remove the buttons' background (the circle behind)

apps/mobile-wallet/src/features/address-deletion/useCanDeleteAddress.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import { AddressHash } from '@alephium/shared'
+
+import { useAppSelector } from '~/hooks/redux'
+import { selectAddressByHash } from '~/store/addressesSlice'
+
+const useCanDeleteAddress = (addressHash: AddressHash) => {
+  const address = useAppSelector((s) => selectAddressByHash(s, addressHash))
+
+  return address && !address.settings.isDefault && address.index !== 0
```
mvaivre, line 27: Ah, I now get why I saw this "bug" of not being able to delete an address. It's because the address is at index 0. As a user this is a bit confusing IMO.

mvaivre (commented):
Code looks great! 👍  Just a couple comments.

apps/mobile-wallet/src/features/address-deletion/useCanDeleteAddress.tsx
nop33: I know, we should adapt our code to allow deleting index 0, but I am afraid it will require quite some changes so for now I think it's ok to ship the feature for all the rest of the addresses. WDYT?

nop33: I remember we had a problem on DW when deleting the initial address (index 0) but tbh I can't think of why this would be a problem on mobile.

nop33: Done

nop33:
@mvaivre all your changes were implemented!

mvaivre (approved):
👏 👏 👏 👏

PR #1015, opened 2024-12-02
Convert desktop wallet Electron main process from JS to TS

nop33:
Resources I used for inspiration:
- https://www.electronjs.org/docs
- https://github.com/electron-vite/create-electron-vite
- https://github.com/vitejs/awesome-vite


apps/desktop-wallet/electron/autoUpdater.ts
nop33, line 1: This code was moved out of `electron.js` and converted to TS. No new code added.

apps/desktop-wallet/electron/menu.ts
nop33, line 1: This code was moved out of `electron.js` and converted to TS. No new code added.

apps/desktop-wallet/electron/nativeTheme.ts
nop33, line 1: This code was moved out of `electron.js` and converted to TS. No new code added.

apps/desktop-wallet/electron/electron-env.ts
nop33, line 1: The `/src/types/window.d.ts` was moved here, following the guidelines:
- https://www.electronjs.org/docs/latest/tutorial/context-isolation#usage-with-typescript
- https://github.com/vitejs/awesome-vite?tab=readme-ov-file#electron
- `pnpm create electron-vite`

apps/desktop-wallet/package.json
```diff
@@ -13,8 +13,7 @@
   "scripts": {
     "build": "tsc && cross-env INLINE_RUNTIME_CHUNK=false VITE_VERSION=$npm_package_version vite build",
     "start": "turbo start:electron",
-    "start:web": "cross-env VITE_VERSION=$npm_package_version vite",
-    "start:electron": "concurrently \"pnpm start:web\" \"electron .\"",
```
nop33, line 17: no need to run `electron .` anymore. The vite plugin takes care of it. We just need to run `vite`.

apps/desktop-wallet/src/api/apiUtils.ts
```diff
@@ -16,8 +16,7 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { NFTTokenUriMetaData } from '@alephium/web3'
-import { FungibleTokenMetadata } from '@alephium/web3/dist/src/api/api-explorer'
```
nop33, line 20: Unless I import `explorer`, the current bundling configuration cannot resolve these types. The web3 package would probably need to be updated for it to work, but I don't want to deal with it. Importing `explorer` works, so I'd rather roll with it.

apps/desktop-wallet/src/modals/SettingsModal/NetworkSettingsSection.tsx
```diff
@@ -66,9 +65,6 @@ const NetworkSettingsSection = () => {
   const { sendAnalytics } = useAnalytics()
   const theme = useTheme()
 
-  const _window = window as unknown as AlephiumWindow
-  const electron = _window.electron
-
```
nop33, line 71: No need for these ugly hacks anymore.

apps/desktop-wallet/electron/autoUpdater.ts
```diff
@@ -0,0 +1,61 @@
...
+(at your option) any later version.
+
+The library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
```
mvaivre, line 17: So much cleaner than before, to have this in TS and in a dedicated file 👏

mvaivre:
Native (system) theme change produces an error:

image

mvaivre:
Code looks good, but the error mentioned above really make me think that we should thoroughly test the auto-update.

Another thing while I'm here: buttons in the header don't work since the last electron update. I've already fixed this in my redesign branch (cf [here](https://github.com/alephium/alephium-frontend/commit/eb76046a14650d9129cc6d1404f770d13422b79b)), you need to use `app-region: no-drag;` in the header children style.

nop33:
> Native (system) theme change produces an error:

Can you plz tell me how you reproduced this?

Switching the "Appearance" in my System Settings works fine.

mvaivre (approved):

PR #1016, opened 2024-12-02
Use @reown/walletkit and update WC packages

nop33:
Possibly solves many WalletConnect connectivity issues on the mobile wallet


mvaivre (approved):
Alrighttt! Let's have our community test this 🚀

PR #1032, opened 2024-12-10
Mw redesign ilias

nop33:
1. Fix top padding of mnemonic import screen
2. Fix welcome card for bigger than 600px screens
3. Fix address & contact icons
4. Better translations for singular/plural


mvaivre (approved):
👍

PR #1034, opened 2024-12-11
Fix bottom screen padding

nop33:
On some devices, `inset.bottom` is 0 so if we want extra padding we have to provide it


mvaivre (approved):
Great! Let's merge, I may tweak it a bit (increase the size)

nop33:
@mvaivre I find it helpful to run both ios and android simulators side by side
image

PR #1045, opened 2024-12-12
Fix iOS font

mvaivre:


nop33 (commented):
Seems to work well on iOS 👏 

But now it doesn't work on Android 🤔 

image

Issue #914, opened 2024-10-22
Mobile wallet crash: `ForegroundServiceStartNotAllowedException`


nop33:
We have received the following report on [Sentry](https://alephium.sentry.io/issues/10129921/?alert_rule_id=80260&alert_timestamp=1729608451969&alert_type=email&environment=production&notification_uuid=cefd59f8-112b-426f-89bc-ecdece67d813&project=4508131917430864&referrer=alert_email).

We are using [`react-native-background-actions`](https://www.npmjs.com/package/react-native-background-actions), which also mentions that:

> [In Android 12+](https://developer.android.com/guide/components/foreground-services#background-start-restrictions) you will not be able to launch background tasks from the background

and the [link it links to](https://developer.android.com/develop/background-work/services/foreground-services#background-start-restrictions) says:

> Apps that target Android 12 or higher can't start foreground services while the app is running in the background, except for [a few special cases](https://developer.android.com/develop/background-work/services/foreground-services#background-start-restriction-exemptions). If an app tries to start a foreground service while the app runs in the background, and the foreground service doesn't satisfy one of the exceptional cases, the system throws a [ForegroundServiceStartNotAllowedException](https://developer.android.com/reference/android/app/ForegroundServiceStartNotAllowedException).

nop33:
Might be related: https://github.com/Rapsssito/react-native-background-actions/issues/242

nop33:
We get that on Sentry as well:


`ForegroundServiceStartNotAllowedException`

`Service.startForeground() not allowed due to mAllowStartForeground false: service org.alephium.wallet/com.asterinet.react.bgactions.RNBackgroundActionsTask`

`Unable to start service com.asterinet.react.bgactions.RNBackgroundActionsTask@92e97a8 with Intent { cmp=org.alephium.wallet/com.asterinet.react.bgactions.RNBackgroundActionsTask (has extras) }: android.app.ForegroundServiceStartNotAllowedException: Service.startForeground() not allowed due to mAllowStartForeground false: service org.alephium.wallet/com.asterinet.react.bgactions.RNBackgroundActionsTask`

nop33:
Looks like we need to update our  [react-native-background-actions](https://www.npmjs.com/package/react-native-background-actions), package

<img width="753" alt="image" src="https://github.com/user-attachments/assets/bda33b4d-2bba-42b6-894d-b611c2c58c84" />

[...]

PR #1060, opened 2024-12-18
Upgrade bg actions package and apply recommendations

nop33:
This is a problem we get often on Sentry. We also got several crash reports on Google in the past.

It doesn't seem to improve the dApp experience. I don't know if it makes it worse either. I also don't know if it fixes the error reports.

See:
- #914


apps/mobile-wallet/app.config.js
```diff
@@ -37,7 +37,7 @@ export default {
       supportsTablet: true,
       bundleIdentifier: 'org.alephium.mobilewallet',
       infoPlist: {
-        BGTaskSchedulerPermittedIdentifiers: ['$(PRODUCT_BUNDLE_PACKAGE_TYPE)'],
+        BGTaskSchedulerPermittedIdentifiers: ['$(PRODUCT_BUNDLE_IDENTIFIER)'],
```
nop33, line 40: This is mentioned in https://github.com/Rapsssito/react-native-background-actions/blob/master/INSTALL.md

image

patches/react-native-background-actions@4.0.1.patch
```diff
@@ -0,0 +1,15 @@
...
+diff --git a/android/src/main/AndroidManifest.xml b/android/src/main/AndroidManifest.xml
+index b67ef4d1837363489a1749acb852997a6c93c1b5..147f04a080e4faf47069c7628bb541aa439f644d 100644
+--- a/android/src/main/AndroidManifest.xml
++++ b/android/src/main/AndroidManifest.xml
+@@ -1,6 +1,6 @@
+ <manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.asterinet.react.bgactions">
+     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+     <application>
+-        <service android:name=".RNBackgroundActionsTask"/>
++        <service android:name=".RNBackgroundActionsTask" android:foregroundServiceType="dataSync" android:stopWithTask="true" />
```
nop33, line 13: Applying the same patch as the one we had before

apps/mobile-wallet/app.config.js
mvaivre: Ok, interesting

mvaivre (approved):
Good one!

PR #1076, opened 2025-01-06
Remove licence header comments from ALL files

nop33:


tdroxler:
just curious on this, do we have some inputs that it's actually unnecessary to have all those headers?

mvaivre:
💥

mvaivre (approved):

nop33:
@tdroxler I had discussed this with Cheng when I asked him why does web3 not use headers and he said it's not needed.

PR #1081, opened 2025-01-07
Redesign polish ilias

nop33:
Just fixing the lint error in your branch and a small rename for cleaner props:

image

@mvaivre I've also run the CI to build internal versions for Android and iOS from this branch to proceed with testing.


mvaivre (approved):
Good one, thanks.

Issue #67, opened 2023-05-03
Frameless window on Windows


mvaivre:

Fugashu:
should be closed:

https://github.com/alephium/alephium-frontend/pull/1104

nop33:
Completed in:
- https://github.com/alephium/alephium-frontend/pull/1104

Closing.

PR #1105, opened 2025-01-11
feat: Add title bar for frameless desktop-wallet

Fugashu:
- Added title bar
- Added ipc handlers to minimize, maximize, close desktop wallet
- Bumped electron dev tools to v4 since there were build issues on Windows
- Added exception handling when installing devtools

### DIFF
--- a/apps/desktop-wallet/electron/main.ts
+++ b/apps/desktop-wallet/electron/main.ts
@@ -140,8 +140,12 @@ app.on('ready', async function () {
       REACT_DEVELOPER_TOOLS,
       REDUX_DEVTOOLS
     } = await import('electron-devtools-installer')
-    await installExtension(REACT_DEVELOPER_TOOLS)
-    await installExtension(REDUX_DEVTOOLS)
+
+    try {
+      await installExtension([REACT_DEVELOPER_TOOLS, REDUX_DEVTOOLS])
+    } catch (e) {
+      console.error('Failed to install devtools:', e)
+    }
   }
 
   handleNativeThemeUserActions()
--- a/apps/desktop-wallet/electron/main.ts
+++ b/apps/desktop-wallet/electron/main.ts
@@ -56,6 +56,7 @@ function createWindow() {
     height: 800,
     minWidth: 1200,
     minHeight: 700,
+    frame: false,
     titleBarStyle: isWindows ? 'default' : 'hidden',
     webPreferences: {
       preload: path.join(__dirname, 'preload.js'),
@@ -219,6 +220,29 @@ app.on('ready', async function () {
     deepLinkUri = null
   })
 
+  ipcMain.handle('window:minimize', () => {
+    mainWindow?.minimize()
+  })
+
+  ipcMain.handle('window:maximize', () => {
+    if (mainWindow?.isMaximized()) {
+      mainWindow.unmaximize()
+    } else {
+      mainWindow?.maximize()
+    }
+  })
+
+  ipcMain.handle('window:close', () => {
+    mainWindow?.close()
+  })
+
+  mainWindow?.on('maximize', () => {
+    mainWindow?.webContents.send('window:maximized', true)
+  })
+
+  mainWindow?.on('unmaximize', () => {
+    mainWindow?.webContents.send('window:maximized', false)
+  })
   createWindow()
 })
 
--- a/apps/desktop-wallet/electron/main.ts
+++ b/apps/desktop-wallet/electron/main.ts
@@ -56,7 +56,7 @@ function createWindow() {
     height: 800,
     minWidth: 1200,
     minHeight: 700,
-    frame: false,
+    frame: !isWindows,
     titleBarStyle: isWindows ? 'default' : 'hidden',
     webPreferences: {
       preload: path.join(__dirname, 'preload.js'),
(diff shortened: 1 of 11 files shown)


nop33 (changes requested):
Thank you so much for the contribution @Fugashu!

To satisfy the following issue, let's make sure that the frame is removed only if the system is Windows, and that the new windows management buttons are also displayed only if the system is Windows.

- https://github.com/alephium/alephium-frontend/issues/67

I look forward to merge this soon! 👏

Fugashu:
Changed as requested!
The title bar with window management buttons are now only shown on Windows devices. 
On other platforms (Mac, Linux) the system default is used.

apps/desktop-wallet/electron/main.ts
```diff
@@ -56,7 +56,8 @@ function createWindow() {
     height: 800,
     minWidth: 1200,
     minHeight: 700,
-    titleBarStyle: isWindows ? 'default' : 'hidden',
+    frame: !isWindows,
+    titleBarStyle: isWindows ? 'hidden' : 'default',
```
nop33, line 60: This used to be `hidden` for mac and linux. I guess now it should be `hidden` in all cases?

apps/desktop-wallet/src/index.css
```diff
@@ -52,9 +52,6 @@ a {
   color: inherit;
 }
 
-#app-header {
-  -webkit-app-region: drag;
-}
```
nop33, line 57: This is required so the user can drag the window around the screen by clicking on the `AppHeader` component.

If a parent has `-webkit-app-region: drag;`, then a child can have `-webkit-app-region: no-drag;` to allow button clicks.

I suggest that:
1. For Windows, we remove the above 3 lines. For all other systems, we keep them.
2. For macOS and Linux, we add `-webkit-app-region: no-drag;` to `HeaderButtons`

apps/desktop-wallet/src/components/TitleBar.tsx
```diff
@@ -0,0 +1,105 @@
...
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+The library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
```
nop33, line 14: These lines can be removed

apps/desktop-wallet/src/components/TitleBar.tsx
```diff
@@ -0,0 +1,105 @@
...
+  color: ${({ theme }) => theme.font.secondary};
+`
+
+const DragRegion = styled.div`
+  flex: 1;
+`
+
+const WindowControls = styled.div`
+  display: flex;
+  -webkit-app-region: no-drag;
```
nop33, line 65: Nice one 👍

nop33 (changes requested):
Good job!

Just 2 small comments to fix macOS

apps/desktop-wallet/electron/main.ts
Fugashu: But if it's hidden on Mac and Linux, how can the user then minimize/maximize/close ?

nop33: Thanks for raising this concern. And I am sorry for the back and forth. I just discovered that our production app is not displayed properly on Linux. The windows management buttons are missing, along with the menu.

Another thing I became curious about is the display of the menu items. On macOS, the menu items of the currently active app are displayed on the very top OS bar and is not part of the app window. That's the default behavior. On Windows and Ubuntu, however, they are part of the app window. Some popular apps hide the menu completely. Others don't.

| | macOS | Ubuntu | Windows |
| production (v2.5.0) | `frame: true, titleBarStyle: 'hidden'` image | `frame: true, titleBarStyle: 'hidden'` image | `frame: true, titleBarStyle: 'default'` image |
| `Fugashu:feat-dw-title-bar` | `frame: true, titleBarStyle: 'default'` image | `frame: true, titleBarStyle: 'default'` image | `frame: false, titleBarStyle: 'hidden'` |

## Windows
I checked some more Windows apps and I noticed that none of them actually displays a menu:

### OneKey
image

### LedgerLive

image


### Edge browser

image


### Ubuntu
I did some research on Ubuntu apps as well. There, I see that some apps display the menu and some others don't:

> I've been trying to download OneKey and Ledger Live on my Ubuntu but because my system is arm64 I cannot run the amd64 downloadable.

### Ledger Live

> screenshot I found on Google

image

### VSCode
image

### Terminal

image

## Conclusion

My conclusion is that we should go for this setup:

### macOS

Same as in current production:
```
frame: true
titleBarStyle: 'hidden'
```

### Linux

Let's hide the menu and display your custom window management buttons (same as on Windows)

```
frame: false
titleBarStyle: 'hidden'
```

### Windows

Just like the code in your PR:

```
frame: false
titleBarStyle: 'hidden'
```

Fugashu: That's the best approach in my opinion as well. Fixed!

apps/desktop-wallet/src/components/TitleBar.tsx
Fugashu: Removed!

Fugashu:
Changed as discussed. Tested on Windows + Ubuntu + Arco Linux

nop33 (approved):
👏 Well done! Thank you!

PR #1106, opened 2025-01-13
Fix iOS crash due to SpinnerModal

nop33:
The problem is that we reset the navigation before the spinner modal is properly closed. It is fixed by rendering the SpinnerModal outside of navigatable screens and managing its state through Redux.

Closes:
- #1102


mvaivre (approved):
Good one! 👏

PR #1108, opened 2025-01-13
Fix prices on mobile wallet

nop33:
Tanstack will deprecate all this.


mvaivre (approved):
I haven't tested the Windows changes in DW, but the MW part LGTM.
I'll rebase my DW redesign on this and test then. 
Thanks guys!

PR #1125, opened 2025-01-14
Polish pre v2

mvaivre:


nop33 (commented):
Please add these commits in the new rebased PR directly: 
- https://github.com/alephium/alephium-frontend/pull/1134

I will review that PR then.

Issue #1136, opened 2025-01-15
Auto update to 2.5.1 gets stuck at 0%


nop33:
2 issues where discovered when debugging this:
1. [ ] A new version is not detected automatically because `updater.checkForUpdates()` fails due to false positive filtering in the [IPC sender validation](https://github.com/alephium/alephium-frontend/blob/next/apps/desktop-wallet/electron/autoUpdater.ts#L21) security feature that was introduced in `2.5.0` (see `isIpcSenderValid` in [source](https://github.com/alephium/alephium-frontend/blob/next/apps/desktop-wallet/electron/utils.ts#L11-L12))
   - A side-effect of this is that users that are on 2.5.0 are now stuck there.
3. [x] https://github.com/alephium/alephium-frontend/issues/1140
   - In the scenario where the auto detection fails, we have a fallback for manual download, which apparently also has a bug.

nop33:
I found the issue, but idk how to solve it yet.

The `isIpcSenderValid(senderFrame)` check is returning `false`.

nop33:
Our IPC sender validation code is this:

```ts
export const isIpcSenderValid = (frame: WebFrameMain | null) => frame && new URL(frame.url).host === 'localhost:3000'
```

https://github.com/alephium/alephium-frontend/blob/next/apps/desktop-wallet/electron/utils.ts#L11-L12

I discovered that in a production app, the `url` is this (for mac):

```
file:///Users/username/dev/alephium/alephium-frontend/apps/desktop-wallet/dist/mac/Alephium.app/Contents/Resources/app.asar/build/index.html
```

Thus, the validation fails.

nop33:
The 18th security best practice says that we should not be using the `file://` protocol for serving local pages: https://www.electronjs.org/docs/latest/tutorial/security#18-avoid-usage-of-the-file-protocol-and-prefer-usage-of-custom-protocols.

We should:
- [ ] figure out how to serve new pages from `alephium://` and then
- [ ] update our IPC sender validation to check for `alephium://`

PR #1141, opened 2025-01-16
Fix auto update

nop33:
Closes:
- https://github.com/alephium/alephium-frontend/issues/1136

This PR:
1. updates the electron app to serve its local files using the `alephium://` protocol instead of the `file://` protocol, as per [best practice](https://www.electronjs.org/docs/latest/tutorial/security#18-avoid-usage-of-the-file-protocol-and-prefer-usage-of-custom-protocols).
2. Fixes the IPC sender validation rule to properly validate interprocess communication with a production build.

See issue for more details:
- https://github.com/alephium/alephium-frontend/issues/1136

### DIFF
--- a/apps/desktop-wallet/electron/paths.ts
+++ b/apps/desktop-wallet/electron/paths.ts
@@ -1,0 +1,14 @@
+import path from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+import isDev from 'electron-is-dev'
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url))
+
+export const APP_ROOT_PATH = path.join(__dirname, '..')
+
+export const RENDERER_PATH = 'build'
+
+const VITE_PUBLIC_PATH = path.join(APP_ROOT_PATH, isDev ? 'public' : RENDERER_PATH)
+
+export const ICON_PATH = path.join(VITE_PUBLIC_PATH, 'icons', 'logo-48.png')
--- a/apps/desktop-wallet/electron/utils.ts
+++ b/apps/desktop-wallet/electron/utils.ts
@@ -1,5 +1,7 @@
 import { app, WebFrameMain } from 'electron'
 
+import { APP_PROTOCOL } from './appProtocol'
+
 export const isMac = process.platform === 'darwin'
 
 export const isWindows = process.platform === 'win32'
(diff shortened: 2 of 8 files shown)


apps/desktop-wallet/electron/utils.ts
```diff
@@ -9,4 +11,15 @@ export const CURRENT_VERSION = app.getVersion()
...
 
 // See: https://www.electronjs.org/docs/latest/tutorial/security#17-validate-the-sender-of-all-ipc-messages
-export const isIpcSenderValid = (frame: WebFrameMain | null) => frame && new URL(frame.url).host === 'localhost:3000'
+export const isIpcSenderValid = (frame: WebFrameMain | null) => {
+  if (frame) {
+    const url = new URL(frame.url)
+
+    return (
+      url.protocol === `${APP_PROTOCOL}:` || // for production builds
+      url.host === 'localhost:3000' // for dev environment
```
nop33, line 20: This is the reason why all IPC calls were filtered out. The production build was using the `file://` protocol, which was filtered out. Instead of allowing the `file://` protocol to pass, I followed the best security practices and replaced it with a custom app protocol `alephium://`.

apps/desktop-wallet/electron/walletConnect.ts
nop33, line 1: Simply grouped all WC related code into this file.

apps/desktop-wallet/src/features/autoUpdate/AutoUpdateSnackbar.tsx
```diff
@@ -27,12 +27,11 @@ const AutoUpdateSnackbar = () => {
   const [error, setError] = useState('')
 
   useEffect(() => {
-    if (!newVersion || requiresManualDownload) return
+    if (!newAutoUpdateVersion) return
```
nop33, line 30: This was the 2nd bug and the reason why the progress bar was shown with 0% instead of showing the "Download" button for manual download.

apps/desktop-wallet/electron/paths.ts
```diff
@@ -0,0 +1,14 @@
+import path from 'node:path'
+import { fileURLToPath } from 'node:url'
+
```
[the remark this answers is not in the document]
nop33: why not? The space is automatically added by our linter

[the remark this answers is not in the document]
nop33: I guess native modules and npm modules need to be separated

mvaivre (approved):
Alrighty! LGTM!

PR #1144, opened 2025-01-16
Lookout for bottom modal animation, wait before loading heavy content

mvaivre:
Perfs boost in AddressDetails modal

### DIFF
--- a/apps/mobile-wallet/src/components/AddressesTokensList.tsx
+++ b/apps/mobile-wallet/src/components/AddressesTokensList.tsx
@@ -1,8 +1,7 @@
 import { AddressHash, Asset } from '@alephium/shared'
-import { Skeleton } from 'moti/skeleton'
-import { useEffect, useMemo, useState } from 'react'
+import { ComponentProps, useEffect, useMemo, useState } from 'react'
 import { useTranslation } from 'react-i18next'
-import { ActivityIndicator } from 'react-native'
+import { ActivityIndicator, View } from 'react-native'
 import styled, { useTheme } from 'styled-components/native'
 
 import AppText from '~/components/AppText'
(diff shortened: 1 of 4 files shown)


apps/mobile-wallet/src/components/AddressesTokensList.tsx
```diff
@@ -79,35 +78,31 @@ const AddressesTokensList = ({ addressHash, isRefreshing, parentModalId }: Addre
       </EmptyPlaceholder>
     )
 
+  const LoadingSection = (props: ComponentProps<typeof View>) => (
+    <LoadingSectionContainer {...props}>
+      <ActivityIndicator size={72} color={theme.font.primary} />
```
nop33, line 83: So we are displaying a loader even though we are not loading any data, just because we are waiting for the animation to finish. Not sure I like it :/

apps/mobile-wallet/src/components/AddressesTokensList.tsx
```diff
@@ -121,30 +116,10 @@ const AddressesTokensListStyled = styled.View`
...
-
-const Loader = styled.View`
-  position: absolute;
-  top: 30px;
-  left: 0;
-  right: 0;
+const LoadingSectionContainer = styled.View`
+  width: 100%;
   align-items: center;
+  padding-top: 100px;
```
nop33, line 122: The spinner seems a bit cut-off at the bottom. I think adding a `padding-bottom: 40px` makes it look better.

```suggestion
  padding: 60px 0 40px;
```

apps/mobile-wallet/src/components/AddressesTokensList.tsx
mvaivre: Personal opinion: 
- It's much better than the UI being glitchy and laggy
- You're brain cannot parse the list of token while it's moving, during in the ~1 second it takes for the modal to open 

Instead of a spinner, we could also have a placeholder (grey boxes likes Skeleton loaders, but not animated. Tried that, but it was laggy) and fade to the actual content once the animation is done. But it takes time to do it properly, I preferred to go for the fastest implementation.

mvaivre: After testing this again, it seems like the "spinner time" could be reduced. Animation seems finished, however the spinner stays for a lil' while.

nop33: yes, that's what I experienced as well

nop33 (approved):

PR #1153, opened 2025-01-20
Fix password requirement for Ledger wallets

mvaivre:
Password should not be needed when a wallet is using Ledger.
@nop33 I don't have my ledger wallet with me so I can't test properly.

Note: I took this opportunity to optimise some Redux selectors.

### DIFF
--- a/.changeset/three-files-cross.md
+++ b/.changeset/three-files-cross.md
@@ -1,0 +1,5 @@
+---
+"alephium-desktop-wallet": patch
+---
+
+Fix password wrongly required for Ledger wallet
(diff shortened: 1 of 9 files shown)


mvaivre:
@nop33 I don't have my ledger wallet with me so I can't test properly.
There's probably more to this issue (reported by Thiago) because it seems that the password was never set in the first place (it shouldn't be possible to set a password for a ledger wallet, I've hidden the setting). 

Original issue: 
> The whole story: My regular wallet has the password-required toggle enabled. Connecting with Ledger assumes the same configuration for the Ledger Wallet.
But I didn't create any password for it. So, when I try to use the one for my other wallet, it does not work.

nop33 (approved):
Code looks good to me. I'll test with a Ledger device tomorrow.

nop33:
@mvaivre I could send a tx with my alephium ledger while the password requirement was enabled on a different wallet 👍 good to merge

mvaivre:
Thanks for testing <3

PR #1154, opened 2025-01-21
Fix i18n plurals, add analytics, improve hide tokens feature

nop33:


apps/mobile-wallet/locales/en-US/translation.json
```diff
@@ -437,5 +437,6 @@
   "nb_of_hidden_assets_one": "{{ count }} hidden asset",
   "nb_of_hidden_assets_other": "{{ count }} hidden assets",
   "Hide asset": "Hide asset",
-  "Show details": "Show details"
+  "Show details": "Show details",
+  "Hiding ALPH is not allowed": "Hiding ALPH is not allowed"
```
mvaivre, line 441: I'm not sure I like the formulation "not allowed". WDYT about simply not showing the hide option for ALPH?

nop33: agreed

nop33: fixed in 0c10dfd27

PR #1157, opened 2025-01-21
Fix hidden assets on network switch

nop33:


mvaivre:
It seems like the loading / displaying of assets when switching network is broken. Just seeing this now, but it was already the case before.

image

Also, I'm seeing now that the unknown tokens are shown in a button that does nothing, and it's a bit confusing to me. I'm thinking about two ways to go about this:
- Easy 1: go back to the previous way of displaying those
- Easy 2: use a static (non pressable) label, place it after the hidden tokens btn (cf. image below)
- More complicated: keep the button, which opens a modal with the addresses containing those unknown tokens, on press, redirect to the addresses' explorer page for those addresses. Ideally, add a URL parameter in the explorer to define which asset tab to open (in order to open the "unknown tokens" tab for this address). I can create a couple of issues for that.

image

nop33:
@mvaivre, I implemented proposal 2.

mvaivre (approved):

PR #1167, opened 2025-01-22
Add navigation buttons in ecosystem dApp WebView

nop33:
The feature is still hidden and it won't be enabled in this PR.

### DIFF
--- a/apps/mobile-wallet/src/features/ecosystem/DAppWebViewScreen.tsx
+++ b/apps/mobile-wallet/src/features/ecosystem/DAppWebViewScreen.tsx
@@ -125,9 +125,7 @@ const BrowserBottomStyled = styled.View`
   flex-direction: row;
   align-items: center;
   justify-content: space-between;
-  padding-left: ${DEFAULT_MARGIN}px;
-  padding-right: ${DEFAULT_MARGIN}px;
-  padding-bottom: 5px;
+  padding: 10px ${DEFAULT_MARGIN}px;
   gap: ${DEFAULT_MARGIN}px;
 `
 
--- a/apps/mobile-wallet/src/features/ecosystem/DAppWebViewScreen.tsx
+++ b/apps/mobile-wallet/src/features/ecosystem/DAppWebViewScreen.tsx
@@ -1,5 +1,5 @@
 import { useNavigation } from '@react-navigation/native'
-import { NativeStackScreenProps } from '@react-navigation/native-stack'
+import { NativeStackNavigationProp, NativeStackScreenProps } from '@react-navigation/native-stack'
 import * as Clipboard from 'expo-clipboard'
 import { useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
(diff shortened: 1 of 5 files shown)


apps/mobile-wallet/src/features/ecosystem/DAppWebViewScreen.tsx
```diff
@@ -1,48 +1,157 @@
...
+              onPress={onGoForward}
+              iconProps={{ name: 'arrow-right' }}
+              squared
+              type="transparent"
+              compact
+              disabled={!canGoForward}
+            />
+            <Button onPress={onReload} iconProps={{ name: 'refresh-cw' }} squared type="transparent" compact />
+          </>
+        )}
```
mvaivre, line 123: I see that this is available for android only. Why is that?

mvaivre: Ok, pull to refresh seems to be available only on iOS.

I think we could relocate the back / forward and refresh action in a bottom bar, available on Bothe platforms :)

nop33: > Ok, pull to refresh seems to be available only on iOS.

Exactly, on iOS you can use gestures and pull to refresh works. So I thought why clutter the UI?

> I think we could relocate the back / forward and refresh action in a bottom bar, available on both platforms

That would reduce the visual space available for the dApp. I think it looks good keeping everything on top! If you strongly disagree, I can implement your suggestion. But I'd like to avoid having a full row used by some small buttons. Up to you

mvaivre: Ideal would be smth like iOS, where the bottom bar is hidden when scrolling down. But may be hard / impossible to implement.

Having all the actions at the top makes it a bit crowded, hard to interact with on small screens. Buttons are super smol and only a small portion of the URL is shown.

mvaivre: > Exactly, on iOS you can use gestures and pull to refresh works. So I thought why clutter the UI?

Gestures don't work for me. Siping left goes back to the previous app screen.

nop33: OK, I'll move them to bottom and update the PR

nop33: Hm, gestures work for me

nop33: @mvaivre buttons moved to footer!

nop33:
@mvaivre I enabled again the "Coming soon" message so that we can safely merge this before it's time to release v2.1. Please review it when you have the time, thanks!

mvaivre:
Reviewed and tested! On thing: could you increase the bottom padding a little bit (maybe 5 px)? on iOS. The controls are super close to the bottom of the screen, which is visually nice but can mess up with OS gestures a bit.

nop33:
@mvaivre padding added!

mvaivre (approved):
🚀

PR #1185, opened 2025-01-23
Pod install

nop33:
@mvaivre I keep getting this diff cause I can't run `pnpm ios` without first running `pnpm dlx pod-install`. Can you please check this branch to your local machine and see if you get any problems when running `pnpm ios`?


mvaivre:
Running it just like that, I get: ```error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.```

Everything's fine after a pod install.

mvaivre (approved):

Issue #1183, opened 2025-01-23
Keyboard does not go away in new address screen on iOS


nop33:
The keyboard should collapse when clicking anywhere outside the label input field

https://github.com/user-attachments/assets/ac9afd6e-6ee6-4e22-a3e7-d187cebe2164

PR #1186, opened 2025-01-23
Alignments & Keyboard

nop33:
- Fixes misaligned address box colored badge and address label
- https://github.com/alephium/alephium-frontend/issues/1183

### DIFF
--- a/.changeset/ten-steaks-appear.md
+++ b/.changeset/ten-steaks-appear.md
@@ -1,0 +1,5 @@
+---
+"@alephium/mobile-wallet": patch
+---
+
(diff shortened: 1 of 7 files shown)


mvaivre (commented):
Thanks! Works great, two small details:

image

Bottom padding too big here.

image

Checkbox's width is bigger than the coloured badge, breaking the alignment a bit. Let's reduce it.

nop33:
The size of the check icon was adjusted ✅ 

Didn't have time to check the bottom padding problem, but on Android is not present and I don't think it's part of this PR

mvaivre (approved):

Issue #64, opened 2023-04-23
Allow "order by" in Addresses page


mvaivre:
Allow users to order addresses by:

- Last use (default)
- Worth (desc)
- Alphabetical (desc)

Use a Select component. Choice should be persisted between session, ideally per wallet. 
Other options like this one will certainly pop up in the future.

Fugashu:
I'll tackle this one next, if noone else started working on it yet

PR #1187, opened 2025-01-26
Feat address ordering

Fugashu:
Added Select component for address sorting/ordering:
- Order by Alph Value
- Order by alphabetical
- Order by last used

Store order preference in localstorage based on wallet id


apps/desktop-wallet/src/storage/addresses/addressesSlice.ts
```diff
@@ -18,18 +18,24 @@ import {
...
 const addressesSlice = createSlice({
   name: 'addresses',
   initialState,
-  reducers: {},
+  reducers: {
+    setAddressOrder: (state, action: PayloadAction<{ walletId: string; order: AddressOrder }>) => {
+      const { walletId, order } = action.payload
+      state.orderPreference[walletId] = order
+    }
+  },
```
nop33, line 38: We prefer to create reducers outside of the `slice.ts` file to avoid circular imports! Check out the `addressesActions.ts` file

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
```diff
@@ -36,6 +123,16 @@ const AddressesTabContent = memo(() => {
       onButtonClick={openNewAddressModal}
       HeaderMiddleComponent={
         <HeaderMiddle>
+          <SelectWrapper>
+            <Select<AddressOrder>
```
nop33, line 127: I didn't even know this syntax is possible 😆 TS should be smart enough to infer the type using the props of the component.

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
nop33, line 1: Address sorting is used in many more places outside the Addresses page. Here's a list where we currently use address sorting.

Wouldn't it make sense that all these places make use of the user's address sorting preference?

image

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
```diff
@@ -1,18 +1,32 @@
...
 import AddressListRow from '@/pages/UnlockedWallet/AddressesPage/addressListRow/AddressListRow'
 import AdvancedOperationsButton from '@/pages/UnlockedWallet/AddressesPage/AdvancedOperationsButton'
 import TabContent from '@/pages/UnlockedWallet/AddressesPage/TabContent'
+import { setAddressOrder } from '@/storage/addresses/addressesSlice.ts'
+import { AddressOrder } from '@/types/addresses.ts'
+
+interface OrderOption {
+  value: AddressOrder
+  label: string
+}
```
nop33, line 27: Check out the `SelectOption` and how it is used! You could use that instead of redefining the type

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
```diff
@@ -1,18 +1,32 @@
...
+
+const STORAGE_KEY = 'address-order-preferences'
```
nop33, line 29: I think this belongs to the general wallet settings. Adding it there would reduce the code needed to manage it's peristed storage and the initialization of the Redux state. It would also keep settings together in one place. WDYT?

Have a look at:
- `settingsTypes.ts`
- `settingsSlice.ts`

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
```diff
@@ -22,8 +36,81 @@ const AddressesTabContent = memo(() => {
   const [hideEmptyAddresses, setHideEmptyAddresses] = useState(false)
   const filteredByText = useFilterAddressesByText(searchInput.toLowerCase())
   const { data: filteredByToggle } = useFetchAddressesHashesWithBalance()
+  const { data: sortedWorthAlph } = useFetchAddressesHashesWithBalanceSortedByAlphWorth()
+  // const { data: sortedTotalWorth } = useFetchAddressesHashesWithBalanceSortedByTotalWorth()
```
nop33, line 40: Sorting by address ALPH balance is a great first step! I wonder, however, how useful this is for users who own many tokens. An address can contain low ALPH balance but high token balance, it would yet appear lower in the sorted list. IMO, it would be much more useful and practical to replace ALPH balance with the whole address worth. WDYT?

apps/desktop-wallet/src/types/addresses.ts
```diff
@@ -3,6 +3,12 @@ import { AddressBalancesSyncResult, AddressHash, AddressSettings, AddressTokensS
 import { explorer as e } from '@alephium/web3'
 import { EntityState } from '@reduxjs/toolkit'
 
+export enum AddressOrder {
+  LastUse = 'lastUse',
+  AlphValue = 'alphValue',
+  Alphabetical = 'alphabetical'
+}
```
nop33, line 10: I would move this to `settingsConstants.ts`

apps/desktop-wallet/src/hooks/useAddresses.ts
```diff
@@ -61,3 +61,75 @@ export const useFetchAddressesHashesWithBalance = () => {
...
+      return a.localeCompare(b)
+    })
+
+    // Then move default address to front if it exists
+    if (defaultAddress) {
+      const defaultIndex = sorted.indexOf(defaultAddress.hash)
+      if (defaultIndex > 0) {
+        sorted.unshift(sorted.splice(defaultIndex, 1)[0])
+      }
+    }
```
nop33, line 127: nit: Could be DRY'ed out into a utility function

nop33 (commented):
Amazing work @Fugashu! I love to see your progress in becoming familiar with the project!

I've left some reviews and requests. The main topics are:
1. Address order settings could be merged with the general app settings. Per-wallet address sorting is cool, but I would keep it simple for now.
2. Sorting by address worth is more useful than working by ALPH balance (which is only 1 of the many tokens and address can have)

apps/desktop-wallet/src/hooks/useAddresses.ts
Fugashu: Refactored and moved to `src/utils/addresses.ts`

apps/desktop-wallet/src/types/addresses.ts
Fugashu: done

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
Fugashu: Oh that's neat, I didn't notice it exists :D

apps/desktop-wallet/src/storage/addresses/addressesSlice.ts
Fugashu: removed the reducer and added to `addressesActions.ts`

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
Fugashu: I think maybe we should include both options. Sometimes you might want to know which address has the most $ALPH if you want to interact with a specfic dApp for example. Both options are useful imo.

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
Fugashu: As a user (at least that's my way how I interact with most apps) I mostly use the addresses, which are 'last used'. But I do have a few dormant addresses which I rarely use but store more Alph for example. Maybe I want to sort by ALPH value, but for most other interactions I normally want to see the 'last used' ones most of the time.

I guess it really depends on the use case

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
Fugashu: moved the logic to the general wallet settings. Makes much more sense!

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
nop33: fair enough 👍

apps/desktop-wallet/src/features/settings/settingsSlice.ts
```diff
@@ -81,6 +82,9 @@ const settingsSlice = createSlice({
       .addCase(fiatCurrencyChanged, (state, action) => {
         state.fiatCurrency = action.payload
       })
+      .addCase(addressOrderPreferenceChanged, (state, action) => {
+        state.addressOrderPreference = action.payload
+      })
```
nop33, line 87: To save the order preference change in local storage, the `addressOrderPreferenceChanged` action needs to be listed in the listener middleware below. See line `101`.

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
nop33: That's an interesting point. Looks like we have to differentiate between:
1. Address sorting in display of addresses tables, like in:
   - Addresses screen
   - Addresses list in Overview screen
2. Address sorting in display of addresses in "Address pickers", like:
   - Address picker in "Send funds" flow
   - Address picker in "Connect with address to dApp" flow
   - Address picker in "Receive funds" flow
   - etc

So the question is: Do we want this new "setting" to control both addresses display tables and address pickers, or only the former?

PS: For future reference, we like to follow the rule that only the creator of a PR discussion can resolve the discussion :)

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
```diff
@@ -22,8 +29,49 @@ const AddressesTabContent = memo(() => {
...
+  ])
+
+  const orderOptions: SelectOption<AddressOrder>[] = [
+    { value: AddressOrder.LastUse, label: t('Last used') },
+    { value: AddressOrder.AlphValue, label: t('ALPH value') },
+    { value: AddressOrder.Alphabetical, label: t('Alphabetical') }
+  ]
+
+  const onSelect = (value: AddressOrder) => {
+    dispatch(addressOrderPreferenceChanged(value))
```
nop33, line 73: Could you please add some analytics here? See how `const { sendAnalytics } = useAnalytics()` is used across the app.

apps/desktop-wallet/src/types/addresses.ts
```diff
@@ -2,6 +2,7 @@ import { NonSensitiveAddressData } from '@alephium/keyring'
 import { AddressBalancesSyncResult, AddressHash, AddressSettings, AddressTokensSyncResult } from '@alephium/shared'
 import { explorer as e } from '@alephium/web3'
 import { EntityState } from '@reduxjs/toolkit'
+import { AddressOrder } from '@/features/settings/settingsConstants.ts'
```
nop33, line 5: file extension can be omitted

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
Fugashu: added analytics

``` typescript
const onSelect = (value: AddressOrder) => {
    dispatch(addressOrderPreferenceChanged(value))
    sendAnalytics({ event: 'Address order changed', props: { value } })
  }
```

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
Fugashu: Personally I'd opt for option 1

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
nop33: @Fugashu are you planning on working on the second option as well?

nop33 (approved):

apps/desktop-wallet/src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
mvaivre: Just tested your work fugashu, nice! 👏 
I also think that ordering by worth is gonna be much more useful.
In the current state, ordering by Alph amount is particularly weird because the amount of ALPH is not displayed in the list. I have in mind a way to improve this in the redesign.

TLDR: Let's introduce ordering by worth, and keep ordering by ALPH as the last option in the list.

nop33: Introducing ordering by worth can come in a subsequent request. Fugashu does not have the bandwidth to work on this at the moment. We leave the issue open: 
- https://github.com/alephium/alephium-frontend/issues/64

mvaivre (approved):

PR #1197, opened 2025-01-28
Region settings

nop33:


mvaivre (commented):
Great work, thanks. A few comments:

- The number formatting isn't applied to token amounts:
image


- A search field would be great for such long lists. WDYT? Let's create an issue? 

- Not directly liked to this PR but it highlighted the issue : let's align text to left here :)
image

nop33:
> The number formatting isn't applied to token amounts:

Fixed ✅ 

> A search field would be great for such long lists. WDYT? Let's create an issue?

- https://github.com/alephium/alephium-frontend/issues/1215

> Not directly liked to this PR but it highlighted the issue : let's align text to left here :)

Fixed ✅

mvaivre (approved):
Should be good now based on our discussion!

PR #1198, opened 2025-01-28
Display tx and fee worth before sending

nop33:


mvaivre:
Added a commit to give more importance to the amounts (bold) + simplifying label (total worth vs. Worth)

image

vs 

image

mvaivre (approved):
Added a commit + approved!

Issue #1215, opened 2025-02-03
Add search bar in long select lists


nop33:
See:
- https://github.com/alephium/alephium-frontend/pull/1197#pullrequestreview-2589998705

PR #1221, opened 2025-02-04
Searchable regions modal

nop33:
Closes:
- https://github.com/alephium/alephium-frontend/issues/1215


mvaivre:
Thanks for the feedback. Because of the behaviour on Android (not tested), I'm thinking that it may be worth keeping this one for after we give a try to https://github.com/gorhom/react-native-bottom-sheet which seems to support search fields by default 🤔 

But I'll test this PR on Android ofc.

nop33:
I really hope this library will improve our modals! Nevertheless, this PR makes searching for a region easier than our current prod app. I think it still provides value. It can be improved upon with the future modal lib. WDYT?

mvaivre:
Tweaked some UI things here https://github.com/alephium/alephium-frontend/pull/1244, approving this one!

mvaivre (approved):

Issue #484, opened 2024-03-28
Display video NFTs in mobile wallet


nop33:
Our community started wanting this:
- https://github.com/alephium/alephium-frontend/issues/840

PR #1223, opened 2025-02-04
Video NFTs

nop33:
Closes: 
- #484

### DIFF
--- a/apps/mobile-wallet/src/components/NFTImage.tsx
+++ b/apps/mobile-wallet/src/components/NFTImage.tsx
@@ -16,6 +16,7 @@ export interface NFTImageProps {
   nftId: NFT['id']
   size?: DimensionValue
   play?: boolean
+  sizeLimited?: boolean
 }
 
 enum NFTDataTypes {
@@ -30,7 +31,7 @@ type NFTDataType = keyof typeof NFTDataTypes
 const maxFileSizeInMB = 5
 const maxFileSizeInBytes = 1024 * 1024 * maxFileSizeInMB
 
-const NFTImage = memo(({ nftId, size = '100%', play }: NFTImageProps) => {
+const NFTImage = memo(({ nftId, size = '100%', play, sizeLimited = true }: NFTImageProps) => {
   const nft = useAppSelector((s) => selectNFTById(s, nftId))
   const theme = useTheme()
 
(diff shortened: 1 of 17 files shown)


apps/mobile-wallet/src/components/nft/NftVideo.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+    if (play) player.play()
+  })
+
+  return (
+    <VideoViewStyled
+      style={{ width: size, height: size }}
+      player={player}
+      allowsFullscreen={false}
+      allowsPictureInPicture
```
mvaivre, line 24: Because NFTs are not long duration content, I don't really see a reason why you would like to see it play in a loop while doing another task, so I wouldn't allow picture in picture :)

apps/mobile-wallet/src/components/NFTImage.tsx
```diff
@@ -12,23 +13,51 @@ import { BORDER_RADIUS_SMALL } from '~/style/globalStyle'
...
-  return hasError ? (
-    <NFTPlaceholder size={size} />
-  ) : isDataUri ? (
-    <NFTWebView imageUri={nft.image} size={size} />
-  ) : (
+  if (!contentType) return <NFTPlaceholder size={size} showCameraOff={false} />
+
+  // Loading many videos at once is too heavy on Android.
+  // Using the image component for NFT lists on Android is sufficient.
+  if (contentType === 'video' && (Platform.OS === 'ios' || play))
```
mvaivre, line 53: Not sure this conditon is correct because the `Platform.OS === 'ios'` has no effect (because of the use of `||`)
- on iOS, this will auto play if `play` is true
- on Android, this will also auto play if `play` is true

As it is today, we can remove the `Platform.OS === 'ios'` as you use the props properly (we don't autoplay in the NFTGrid for both platform which is fine by me).
If you wanted to add a safeguard to avoid wastful autoplay on Android, we may need 2 different props (`play` & `autoPlay`).

apps/mobile-wallet/src/components/NFTImage.tsx
```diff
@@ -12,23 +13,51 @@ import { BORDER_RADIUS_SMALL } from '~/style/globalStyle'
...
+    return <NftVideo size={size} videoSource={nft.image} play={play} />
```
mvaivre, line 54: Let's overlay a subtle "play" icon on video NFT thumbnails, like we do on the explorer? WDYT?

apps/mobile-wallet/src/components/NFTImage.tsx
nop33: As the comment above describes, we use the `NftVideo` component when:
1. `contentType === 'video' && Platform.OS === 'ios'`
2. `contentType === 'video' && play`

apps/mobile-wallet/src/components/nft/NftVideo.tsx
nop33: Done ✅

apps/mobile-wallet/src/components/NFTImage.tsx
nop33: I changed the condition to make it a bit clearer, does it help?

mvaivre: Ah I didn't understand that even just loading the video component was already too much!
Thanks for answering, clearer now

mvaivre (approved):
Great work!! Works great. Pushed a tiny commit to reduce a margin.

Issue #1209, opened 2025-02-03
Fix ALPH token icon


nop33:
Use the one based on our brand guide

mvaivre:
I've updated the icon in the token list, but it's cached on device (tested with cgi). A cache clean displays the new logo. We could maybe look into caching images a bit less aggressively

PR #1229, opened 2025-02-05
Fix ALPH logo

nop33:
Closes:
- #1209


packages/shared/src/store/assets/fungibleTokensSlice.ts
```diff
@@ -47,6 +47,7 @@ const fungibleTokensSlice = createSlice({
             state,
             metadata.tokens.map((tokenInfo) => ({
               ...tokenInfo,
+              logoURI: `${tokenInfo.logoURI}${tokenInfo.id === ALPH.id ? '?v2' : ''}`,
```
mvaivre, line 50: Interesting technique!

mvaivre (approved):

PR #1234, opened 2025-02-06
Bring back lost commits and add hide/unhide token

nop33:


nop33:
Superseded by https://github.com/alephium/alephium-frontend/pull/1235

PR #1237, opened 2025-02-07
Add shortcut buttons and address list to Token Details Modal

nop33:


apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletBalancesTokensByAddress.ts
```diff
@@ -25,14 +25,17 @@ const combineBalancesByAddress = (
...
+  AddressTokensBalancesQueryFnData,
+  AddressesTokensBalances['data']
+>({
   useDataHook: useFetchWalletBalancesTokens,
   combineFn: combineBalancesByAddress,
   defaultValue: {}
 })
 
+const useFetchWalletBalancesTokensByAddress = useData
+const UseFetchWalletBalancesTokensByAddressContextProvider = DataContextProvider
```
nop33, line 38: sorry man but this code is always so hard to understand every time I want to dive into the implementation of these hooks. This tiny change helps a little bit imo

mvaivre:
Nice one!
In terms of UX, what would make the most sense IMO is to provide the list of latest transactions involving the tokens. The detailed list of addresses containing the token comes second IMO (we know that the large majority of our users only have 1 address).
Because this PR is still rather smol, would you mind implementing this? 🚀 

What I'd love to see:
- Like on MW, a box with the token description
- Under the box, a tabbar allowing to switch between "Latest transactions" and "Addresses".

WDYT? Once we have this + the last UI fixes I need to do before end of week, we'll be ready to ship 3.0!

nop33:
Sure, I'll work on these features tomorrow!

apps/desktop-wallet/package.json
```diff
@@ -131,6 +131,7 @@
     "react-confetti": "^6.0.1",
     "react-detect-click-outside": "^1.1.2",
     "react-dom": "^18.2.0",
+    "react-freeze": "^1.0.4",
```
nop33, line 134: It was introduced for 2 reasons:
1. Improve UX of tab navigation
2. Improve performance by "freezing" the hidden tab content

nop33:
@mvaivre re-requesting review

mvaivre (approved):
LGTM! 🚀

Issue #1246, opened 2025-02-10
Hide secret recovery phase words that have already been entered when restoring in mobile wallet


nop33:
### Discussed in https://github.com/alephium/alephium-frontend/discussions/816

<div type='discussions-op-text'>

<sup>Originally posted by **nop33** September  3, 2024</sup>
Feedback:

> I was in the train when verifying the seed. When i input the next word, the whole seed should not be displayed visibly. You could leave the last few words, but it’s not necessary to have all of them visible on the screen.</div>

PR #1247, opened 2025-02-10
Hide already entered words of imported mnemonic

nop33:
Closes:
- #1246


mvaivre (approved):
Really cool one! Took the opportunity to fix a top padding.
I would have liked to add a btn in the header to view the full phrase (make it visible while the button is pressed), but too complicated to implement for tonight :P

PR #1248, opened 2025-02-10
Display entered amount fiat worth

nop33:
Very useful when someone wants to send a certain amount of fiat-equivalent to someone else.

See feature request: 
- https://github.com/alephium/alephium-frontend/discussions/432

image


mvaivre (approved):
Used this opportunity to fix the amount component catching the press event when it shouldn't. LGTM

Issue #1252, opened 2025-02-11
Reduce battery usage


nop33:
Resources:
- https://reactnative.dev/docs/performance
- https://medium.com/@anisurrahmanbup/react-native-ultimate-guide-on-debugging-profiling-performance-optimization-ios-android-7e44b8690cbe
- https://medium.com/@anisurrahmanbup/react-native-app-performance-optimization-ios-android-be-a-sr-react-native-engineer-today-767cfcb0ed70
- https://docs.expo.dev/debugging/tools/
- https://github.com/anisurrahman072/React-Native-Advanced-Guide

Related issues:
- https://github.com/alephium/alephium-frontend/issues/522
- https://github.com/alephium/alephium-frontend/issues/743

nop33:
After some investigation data shows that our app has very high energy impact.

![Image](https://github.com/user-attachments/assets/c1b15a39-6ff2-486a-b775-01050da3c750)

The culprits for these are the following:
1. The display consumes a lot of energy when the light theme is on. It drastically improves when the dark theme is on.
2. The animated backgrounds bring the GPU usage and Overhead usage (device hardware) very high
3. The data polling increases the use of Network

## 💡 Ideas for optimizations

### 1. Display

1. Make the dark theme the default one.
1. Auto-switch to dark theme when battery level is low and when battery saving mode is on
1. Update light theme colors to a darker shade
1. Create new unified "gray" theme that does not consume a lot of energy (basically no dark/light anymore. Examples: Revolut, Phantom, Metamask)

### 2. GPU
1. Investigate how to "freeze" background animations when the focus is not on the screen with the animations
1. Remove animations (👎 not preferred)
1. Freeze animations when battery level low and when battery saving mode is on
1. Remove animations when battery level low and when battery saving mode is on

### 3. Network
1. Reduce data polling from 10s to 60s
1. Use Tanstack & caching to reduce number of request and reduce transferred data
1. Do not keep app active when in bg to keep WC session active now that ecosystem screen is integrated1. dApps should not use the wallet's node & explorer providers through WC to request data and should rely on web3 providers

### General
1. Create "Battery saving" mode setting that is "off" by default. Enabling it could apply several of the optimizations above.

Recommendations by Apple (through XCode)

![Image](https://github.com/user-attachments/assets/b5310aa4-2674-4b26-9acf-9e77712ffdf7)

mvaivre:
Thanks for the report! Mains things I have in mind:
[...]

PR #1257, opened 2025-02-11
Improve performance

nop33:
Related to:
- #1252


apps/mobile-wallet/index.ts
```diff
@@ -4,6 +4,9 @@ import './setupErrorHandling'
 
 import * as Sentry from '@sentry/react-native'
 import { registerRootComponent } from 'expo'
+import { enableFreeze } from 'react-native-screens'
+
+enableFreeze(true)
```
nop33, line 9: > The current version of screens/freeze integration, freezes updates only on views that are more than one level deep the stack hierarchy (the top and second to top screens are not freezing). This is necessary for slide-to-go-back functionality to work as by sliding back we may reveal the content that is displayed below. This is something we plan on improving in the future such that only the top screen is not frozen.
https://github.com/software-mansion/react-freeze?tab=readme-ov-file#quick-start-with-react-navigation-react-native-

apps/mobile-wallet/src/features/addressesManagement/AddressDetailsModalHeader.tsx
```diff
@@ -154,7 +154,7 @@ const AddressAnimatedBackground = ({ addressHash }: Pick<AddressDetailsModalHead
 
   if (!address) return null
 
-  return <AnimatedBackground shade={address.settings.color} isAnimated />
```
nop33, line 157: It's the default

apps/mobile-wallet/src/components/animatedBackground/ScreenAnimatedBackground.tsx
nop33, line 1: This is the main new component that enables animation and gyroscope only when the screen is in focus. The energy consumption remains as it was before when the user remains on the Dashboard, but it significantly reduces in every other screen 👍

apps/mobile-wallet/metro.config.js
```diff
@@ -20,4 +20,11 @@ config.resolver.nodeModulesPaths = [
 // Use turborepo to restore the cache when possible
 config.cacheStores = [new FileStore({ root: path.join(projectRoot, 'node_modules', '.cache', 'metro') })]
 
+config.transformer.minifierConfig = {
+  compress: {
+    // The option below removes all console logs statements in production.
```
mvaivre, line 25: Nice

PR #1262, opened 2025-02-14
Improve DW v3 performance

nop33:
~~I will explain in detail the thought process behind these changes~~ Done.


apps/desktop-wallet/src/api/apiDataHooks/market/useFetchTokenPrices.ts
```diff
@@ -50,7 +50,7 @@ const useFetchWalletFtsSymbols = () => {
   const {
     data: { listedFts },
     isLoading: isLoadingTokensByType
-  } = useFetchWalletTokensByType({ includeAlph: true })
+  } = useFetchWalletTokensByType({ includeHidden: true })
```
nop33, line 53: some defaults where changed, hence this diff

apps/desktop-wallet/src/api/apiDataHooks/token/useFetchToken.ts
```diff
@@ -1,54 +1,29 @@
...
-      id,
-      networkId,
-      skip: isLoadingTokenType || tokenType?.stdInterfaceId !== e.TokenStdInterfaceId.Fungible
-    })
-  )
-
-  const { data: nft, isLoading: isLoadingNft } = useFetchNft({
-    id,
-    skip: isLoadingTokenType || tokenType?.stdInterfaceId !== e.TokenStdInterfaceId.NonFungible
-  })
```
nop33, line 36: This "waterfall query" logic was moved inside a `queryFn`, read below for details

apps/desktop-wallet/src/api/apiDataHooks/utils/useFetchFtList.ts
```diff
@@ -19,22 +16,8 @@ type FTListProps = {
...
-    ...getQueryConfig({ staleTime: ONE_DAY_MS, gcTime: Infinity, networkId }),
-    queryFn:
-      !network || props?.skip
-        ? skipToken
-        : () =>
-            network === 'devnet'
-              ? [ALPH]
-              : axios.get(getTokensURL(network)).then(({ data }) => (data as TokenList)?.tokens),
-    placeholderData: network === 'mainnet' ? mainnet.tokens : network === 'testnet' ? testnet.tokens : []
-  })
```
nop33, line 37: Simply moved the code into a new `ftListQuery` function, because I wanted to reuse it in multiple places.

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletBalancesAlphByAddress.ts
nop33, line 1: This wasn't used anywhere anymore

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletBalancesByAddress.tsx
nop33, line 1: `useFetchWalletBalancesByAddress` is a merge of `useFetchWalletBalancesAlphByAddress` and `useFetchWalletBalancesTokensByAddress` because they were always used together and we currently have no use-case where one of them would be used without the other. The API contexts were also simplified with this change, see below.

apps/desktop-wallet/src/api/apiDataHooks/wallet/useFetchWalletTokensByType.tsx
nop33, line 1: the `useFetchWalletTokensByType` is being used a lot in our app. Making it an API data context improves the performance of our app a lot

apps/desktop-wallet/src/api/context/apiContext.tsx
```diff
@@ -1,18 +1,18 @@
...
+import { UseFetchWalletTokensByTypeContextProvider } from '@/api/apiDataHooks/wallet/useFetchWalletTokensByType'
 
 type ProviderProps = { children: ReactNode }
 type ProviderComponent = FC<ProviderProps>
 
 const providers: Array<ProviderComponent> = [
   UseFetchWalletBalancesTokensArrayContextProvider,
-  UseFetchWalletBalancesTokensByAddressContextProvider,
   UseFetchWalletBalancesAlphArrayContextProvider,
-  UseFetchWalletBalancesAlphByAddressContextProvider
```
nop33, line 15: These 2 contexts were merged together

apps/desktop-wallet/src/api/queries/tokenQueries.ts
```diff
@@ -126,3 +147,44 @@ export const nftDataQuery = ({ id, tokenUri, networkId, skip }: NFTQueryProps) =
...
+        if (!nftMetadata) return nst
+
+        const nftData = await queryClient.fetchQuery(nftDataQuery({ id, tokenUri: nftMetadata.tokenUri, networkId }))
+
+        return nftData ?? nst
+      }
+
+      // 5. If the type of the token cannot be determined, return the non-standard token
+      return nst
+    },
```
nop33, line 188: This is where the most important part of the PR. My investigation showed that the `useFetchToken` hook was very slow. It was using many `useQuery` calls, one after the other, with the `skip` prop controlling the order. This is apparently bad for performance. A smarter way to achieve the same result is to make use of the Tanstack query caching mechanism instead. By creating a new query (`tokenQuery`) where its `queryFn` function is composed by the all other uses of `useQuery` (by using `queryClient.fetchQuery` instead of `useQuery` since React hooks cannot be called inside a callback), we benefit not only from caching the result of the useFetchToken` hook so that the performance of the app improves, but also from reusability in other parts of our app (we can now use the `tokenQuery` inside the `useQueries` call for displaying info about the tokens of a transaction).

I believe this is an innovative idea for better app performance, it feels that this is how Tanstack is meant to be used.

mvaivre: Love it. It's concise and clear.

apps/desktop-wallet/src/types/tokens.ts
```diff
@@ -10,11 +9,13 @@ export interface NonStandardToken {
 }
 
 // To represent a token that is not in the token list but we haven't discovered its type yet
-export type UnlistedToken = { id: string }
+export type UnlistedToken = { id: TokenId }
 
 // For stricter typings in our components that handle display of multiple token types
 export type TokenDisplay = ListedFTDisplay | UnlistedFTDisplay | NFTDisplay | NonStandardTokenDisplay
 
+export type Token = ListedFT | UnlistedFT | NFT | NonStandardToken
```
mvaivre, line 17: 👍

mvaivre (approved):
Awesome work!!

PR #1280, opened 2025-02-25
Pre DW 3 fixes

nop33:


apps/desktop-wallet/src/components/Ellipsed.tsx
```diff
@@ -1,49 +1,26 @@
...
-    }
-  }, [text])
-
-  useWindowResize(handleResize)
-
-  useEffect(() => {
-    handleResize()
-  }, [handleResize, text])
+  const needsEllipsis = text.length > 11
+  const truncatedText = needsEllipsis ? text.substring(0, 4) + '...' + text.slice(-4) : text
```
mvaivre, line 11: So this isn't dynamic anymore. In case we display addresses where we have space (check TX modal for instance), the user will not be able to check the full hash?

apps/desktop-wallet/src/features/autoUpdate/AutoUpdateToastBox.tsx
```diff
@@ -12,7 +12,7 @@ import { openInWebBrowser } from '@/utils/misc'
 
 type UpdateStatus = 'download-available' | 'downloading' | 'download-finished' | 'download-failed'
 
-const AutoUpdateToastBox = () => {
+const AutoUpdateToastBox = memo(() => {
```
mvaivre, line 15: Can't wait for the react compiler :)

apps/desktop-wallet/src/components/DefaultAddressSwitch.tsx
```diff
@@ -73,7 +73,7 @@ const SelectCustomComponent = (value?: SelectOption<AddressHash>, disablePointer
       disablePointer={disablePointer}
       squared
     >
-      {value?.value && <AddressBadgeStyled addressHash={value.value} truncate />}
+      {value?.value && <AddressBadgeStyled addressHash={value.value} truncate disableCopy />}
```
mvaivre, line 76: Easy fix, but that's actually one of the practical places where to copy the main address hash 🤔 
But let's do it that way for now.

apps/desktop-wallet/src/components/Ellipsed.tsx
nop33: Correct, to avoid jumping around, and since we use monospace font for hashes, this PR proposes to unify hashes and display only the first and last 4 characters.

apps/desktop-wallet/src/features/autoUpdate/AutoUpdateToastBox.tsx
nop33: Same. I actually already integrated the eslint package of react compiler. We have quite some things to fix :) A PR will come after the release.

apps/desktop-wallet/src/components/DefaultAddressSwitch.tsx
nop33: You can still copy it once you click on it (from the list, it will be the first item)

mvaivre (approved):

PR #1281, opened 2025-02-25
Remove unused permissions

nop33:


apps/mobile-wallet/app.config.js
```diff
@@ -59,12 +59,6 @@ export default {
         foregroundImage: './assets/adaptive-icon.png',
         backgroundColor: '#000000'
       },
-      permissions: [
-        'android.permission.FOREGROUND_SERVICE',
-        'android.permission.FOREGROUND_SERVICE_DATA_SYNC',
-        'android.permission.START_FOREGROUND_SERVICES_FROM_BACKGROUND', // To potentially fix crash due to ForegroundServiceStartNotAllowedException
```
mvaivre, line 65: Are we sure it wasn't preventing the crashes?

mvaivre (approved):
One question, approving already :)

apps/mobile-wallet/app.config.js
nop33: We removed the background action for keeping wallet connect connection live in the background. So it's not needed.

nop33: We have no access to Sentry since we haven't solved the quota limit due to spamming still. So we can't know if it's solved.

PR #1282, opened 2025-02-25
Add CNY

nop33:


packages/shared/src/types/currencies.ts
```diff
@@ -1 +1,14 @@
-export type Currency = 'CHF' | 'IDR' | 'GBP' | 'EUR' | 'USD' | 'VND' | 'RUB' | 'TRY' | 'CAD' | 'AUD' | 'THB' | 'HKD'
+export type Currency =
```
tdroxler, line 1: you could also use the [Currencies](https://github.com/alephium/alephium-web3/blob/373de95c314d7c38f20c85cdaf85c2851932679f/packages/web3/src/api/api-explorer.ts#L612-L625) variable from the explorer-backend api.

PR #1286, opened 2025-02-26
Pre DW v3 fixes

nop33:


nop33:
Self-reviewed, merging since Mika's on holidays.

PR #1301, opened 2025-03-18
Dw perf

nop33:


apps/desktop-wallet/src/api/apiDataHooks/address/useFetchAddressTokensByType.ts
```diff
@@ -1,18 +1,23 @@
...
-interface UseFetchAddressTokensByType extends UseFetchAddressProps {
-  includeAlph: boolean
-}
+const useFetchAddressTokensByType = ({ addressHash }: UseFetchAddressProps) => {
+  const networkId = useCurrentlyOnlineNetworkId()
 
-const useFetchAddressTokensByType = (props: UseFetchAddressTokensByType) => {
-  const { data: allTokensBalances, isLoading: isLoadingBalances } = useFetchAddressBalances(props)
-  const { data, isLoading } = useFetchTokensSeparatedByType(allTokensBalances)
+  const { data, isLoading } = useQuery(addressTokensByTypeQuery({ addressHash, networkId }))
```
nop33, line 10: Utilizing Tanstack query cache to avoid re-calculating data by moving code into a `queryFn`, like we currently do in other areas.

apps/desktop-wallet/src/api/apiDataHooks/utils/useFetchTokensSeparatedByListing.ts
```diff
@@ -14,23 +14,7 @@ interface TokensByListing<T> {
...
-      if (listedFT) {
-        acc.listedFts.push({ ...listedFT, ...token })
-      } else {
-        acc.unlistedTokens.push(token)
-      }
-
-      return acc
-    }, initial)
-  }, [ftList, tokens])
+  const data = useMemo(() => separateTokensByListing(tokens, ftList), [ftList, tokens])
```
nop33, line 17: Moved to a separate function for reusability

apps/desktop-wallet/src/api/queries/addressQueries.ts
```diff
@@ -94,7 +97,8 @@ export type AddressTokensSearchStringQueryFnData = {
 // Generates a string that includes the names and symbols of all tokens in the address, useful for filtering addresses
 export const addressTokensSearchStringQuery = ({ addressHash, networkId }: AddressLatestTransactionQueryProps) =>
   queryOptions({
-    queryKey: ['address', addressHash, 'tokensSearchString', { networkId }],
+    queryKey: ['address', addressHash, 'computedData', 'tokensSearchString', { networkId }],
+    ...getQueryConfig({ staleTime: Infinity, gcTime: Infinity, networkId }),
```
nop33, line 101: Calculating the search string is an expensive operation (consider that the search string of an address is a string that includes token names, symbols, nft names, etc). We don't want to flush this cache until it's necessary (which is when the balances of the address change)

apps/desktop-wallet/src/api/queries/addressQueries.ts
```diff
@@ -132,3 +135,33 @@ export const addressTokensSearchStringQuery = ({ addressHash, networkId }: Addre
...
+      const tokenTypesPromiseResults = await Promise.allSettled(
+        unlistedTokens.map(({ id }) => queryClient.fetchQuery(tokenTypeQuery({ id, networkId })))
+      )
+      const tokenTypes = getFulfilledValues(tokenTypesPromiseResults)
+
+      const { fungible: unlistedFtIds, 'non-fungible': nftIds, 'non-standard': nstIds } = combineTokenTypes(tokenTypes)
+
+      return { listedFts, unlistedTokens, unlistedFtIds, nftIds, nstIds }
+    }
+  })
```
nop33, line 167: Instead of aggregating all this data in a hook (which can be expensive when there are many addresses and cannot be cached), aggregating them in a `queryFn` in combination with `fetchQuery` (important!) allows us to benefit from the Tanstack caching mechanism.

apps/desktop-wallet/src/components/NFTCard.tsx
```diff
@@ -17,36 +18,45 @@ interface NFTCardProps {
 }
 
 const NFTCard = ({ nftId }: NFTCardProps) => {
+  const ref = useRef(null)
+  const isInView = useInView(ref, { once: true })
```
nop33, line 22: Solve a major performance downgrade when there are many NFTs.

apps/desktop-wallet/src/features/dataPolling/useAddressesDataPolling.ts
nop33, line 1: One of the significant performance pain-points of our app is the frequent querying for the latest transactions of each address, especially when the wallet has numerous addresses. We currently query:
1. EVERY address once every 10 seconds
2. internal addresses involved in a sent transaction the moment the transaction confirms

Do we need such frequent polling, given that the app updates immediately when sending txs? Also, is it really necessary to query ALL addresses, even the once that are not being used frequently?

This PR proposes 2 changes:
1. Increase the polling inteval time from 10 seconds to 1 minute
1. Instead of using the same polling interval for ALL addresses, use a different one based on the frequency of use: frequently used addresses have a short polling time (1 minute) and rarely used ones have a longer polling time (5 minutes).

apps/desktop-wallet/src/features/walletConnect/walletConnectContext.tsx
nop33, line 1: Fixes a performance bug when the client gets re-initialized indefinitely when initialization is slower than the retry interval

apps/desktop-wallet/src/components/Inputs/Select.tsx
```diff
@@ -332,6 +319,51 @@ export function SelectOptionsModal<T extends OptionValue>({
...
+      onClick={() => handleOptionSelect(option.value as T)}
+      selected={isSelected}
+      focusable
+      aria-label={option.label}
+      hasCustomOptionRender={!!optionRender}
+      ref={ref}
+    >
+      {optionRender ? (
+        <CustomOptionContainer isSelected={isSelected}>
+          {(index < 20 || isInView) && optionRender(option, isSelected)}
```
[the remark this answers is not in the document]
mvaivre: Nice

mvaivre (approved):
Everything seems to work well, great work! Let's go with an RC! 🚀

PR #1304, opened 2025-03-19
Persist tanstack cache in mobile wallet

nop33:


packages/shared-react/src/api/persistQueryClientContext.tsx
```diff
@@ -45,37 +53,43 @@ export const PersistQueryClientContextProvider = ({ children }: { children: Reac
...
+    },
+    [createPersister]
+  )
 
-  const deletePersistedCache = useCallback((walletId: string) => {
-    createTanstackIndexedDBPersister('tanstack-cache-for-wallet-' + walletId).removeClient()
-  }, [])
+  const deletePersistedCache = useCallback(
+    (walletId: string) => {
+      createPersister('tanstack-cache-for-wallet-' + walletId).removeClient()
```
mvaivre, line 89: Nit: It's a bit weird to me that we run the create function again to simply remove the client. But I don't mind too much :P

mvaivre (approved):
One nit, but approving already! GJ!!

PR #1309, opened 2025-03-21
Mw tanstack 2

nop33:


apps/desktop-wallet/src/api/apiDataHooks/utils/useFetchFtList.ts
```diff
@@ -18,7 +18,6 @@ const useFetchFtList = (props?: FTListProps): FTList => {
 
   const { data, isLoading } = useQuery(ftListQuery({ networkId, skip: props?.skip }))
 
-  // TODO: Maybe return an object instead of an array for faster search?
```
nop33, line 21: I implemented this change which should improve performance

apps/desktop-wallet/src/api/queries/tokenQueries.ts
```diff
@@ -156,14 +160,15 @@ export const nftDataQuery = ({ id, tokenUri, networkId, skip }: NFTQueryProps) =
 
 export const tokenQuery = ({ id, networkId, skip }: TokenQueryProps) =>
   queryOptions({
+    // eslint-disable-next-line @tanstack/query/exhaustive-deps
```
nop33, line 163: For some reason the linter thinks that `fTList[id]` should be added as a dep, but that doesn't make sense.

apps/desktop-wallet/src/api/transactions.ts
```diff
@@ -1,13 +1,12 @@
...
 
 import { LedgerAlephium } from '@/features/ledger/utils'
-import { Address } from '@/types/addresses'
 import { CsvExportQueryParams } from '@/types/transactions'
 
-export const buildSweepTransactions = async (fromAddress: Address, toAddressHash: AddressHash) => {
+export const buildSweepTransactions = async (fromAddressHash: AddressHash, toAddressHash: AddressHash) => {
   const { unsignedTxs } = await throttledClient.node.transactions.postTransactionsSweepAddressBuild({
-    fromPublicKey: fromAddress.publicKey,
+    fromPublicKey: keyring.exportPublicKeyOfAddress(fromAddressHash),
```
nop33, line 9: In my goal to unify the `Address` types between the desktop wallet and the mobile wallet I realized that the only property that these 2 apps will eventually have different (after the mw migration to tanstack) is the `publicKey`. Desktop wallet includes it whereas mobile wallet doesn't. The reason mobile wallet doesn't is that it stores it in secure storage to avoid regenerating them on wallet unlock which is expensive for the mobile env. By relying on the keyring for public key access in the desktop wallet instead of the redux state we remove the duplicated state (until now the public key existed in both the keyring and the redux state) and also prepare for the unification of the 2 Address types between the wallets.

apps/desktop-wallet/src/storage/addresses/addressesSlice.ts
```diff
@@ -1,47 +1,43 @@
...
   walletSaved,
   walletSwitched,
   walletUnlocked
 } from '@/storage/wallets/walletActions'
-import { Address, AddressBase, AddressesState } from '@/types/addresses'
 import { UnlockedWallet } from '@/types/wallet'
 import { getInitialAddressSettings } from '@/utils/addresses'
 
-const initialState: AddressesState = addressesAdapter.getInitialState({
-  isRestoringAddressesFromMetadata: false
```
nop33, line 26: This was not used anywhere

apps/desktop-wallet/src/storage/addresses/addressesSlice.ts
```diff
@@ -1,47 +1,43 @@
...
-      .addCase(addressSettingsSaved, (state, action) => {
-        const { addressHash, settings } = action.payload
-
-        if (settings.isDefault) updateOldDefaultAddress(state)
-
-        addressesAdapter.updateOne(state, {
-          id: addressHash,
-          changes: settings
-        })
-      })
```
nop33, line 44: Moved to shared to be reused by the mobile wallet

apps/desktop-wallet/src/types/addresses.ts
nop33, line 1: All these types have been moved to shared

apps/mobile-wallet/src/api/transactions.ts
```diff
@@ -1,17 +1,14 @@
-import { AddressHash, AssetAmount, client } from '@alephium/shared'
+import { AddressHash, AssetAmount, client, DEPRECATED_Address as Address } from '@alephium/shared'
```
nop33, line 1: I moved the mobile wallet `Address` type to shared and I renamed it to `DEPRECATED_Address`. By importing it and renaming it with `DEPRECATED_Address as Address` makes it easy for migration since I'll just need to remove the `DEPRECATED_Address as ` part once migration is done.

apps/mobile-wallet/src/components/AddressBadge.tsx
```diff
@@ -53,10 +53,10 @@ const AddressBadge = ({
       {address ? (
         <AddressBadgeContainer>
           {!hideSymbol && <AddressColorSymbol addressHash={addressHash} size={16} />}
-          {address.settings.label ? (
+          {address.label ? (
```
nop33, line 56: The address settings in the desktop wallet are stored flat in the `Address` type, so I changed this in the mobile wallet as well to prepare for the migration.

apps/mobile-wallet/src/contexts/walletConnect/WalletConnectContext.tsx
```diff
@@ -745,9 +745,6 @@ export const WalletConnectContextProvider = ({ children }: { children: ReactNode
         }
       }
     },
-    // The `addresses` dependency causes re-rendering when any property of an Address changes, even though we only need
-    // the `hash` and the `publicKey`. Creating a selector that extracts those 3 doesn't help.
-    // Using addressIds fixes the problem, but now the api/transactions.ts file becomes dependant on the store file.
```
nop33, line 750: This comment was not representing reality anymore.

apps/mobile-wallet/src/components/AddressBadge.tsx
mvaivre: sounds reasonable

packages/shared/src/types/addresses.ts
```diff
@@ -20,10 +21,33 @@ export type AddressSettings = {
...
+  hash: AddressHash
+}
+
+// hash: AddressHash
+// index: number
+// group: number
+// isDefault: boolean
+// color: string
+// label?: string
+export type Address = AddressBase & {
```
mvaivre, line 47: Does it make sense to have both `AddressBase` and `Address` when only one field is different? 
Would it make sense to simply have: 

```ts 
type Address = AddressMetadata & { 
  hash: AddressHash
  group: number
}
```

mvaivre: Yes we'd need to derive the group by default, but it would prevent confusing types.

apps/desktop-wallet/src/api/queries/tokenQueries.ts
nop33: I noticed that in my next PR where I move this code to the shared package, the linter doesn't complain anymore..!

PR #1313, opened 2025-03-25
Fix dw crash

nop33:
- Fix crash due to uninitialized address state
- Fix NFTs loading by infinite loading pagination
- Fix cached data loading on app launch
- Fix tabs glitch

### DIFF
--- a/apps/desktop-wallet/src/components/tabs/TabBar.tsx
+++ b/apps/desktop-wallet/src/components/tabs/TabBar.tsx
@@ -15,8 +15,8 @@ export interface TabBarProps<T extends string> {
   justifyTabs?: 'left' | 'center'
   TabComponent?: typeof Tab
   className?: string
-  onMouseOverTabBar?: (tabValue: T) => void
-  onMouseLeaveTabBar?: (tabValue: T) => void
+  onMouseEnterTab?: (tabValue: T) => void
+  onMouseLeaveTab?: (tabValue: T) => void
 }
 
 const TabBar = <T extends string>({
@@ -28,8 +28,8 @@ const TabBar = <T extends string>({
   TabComponent = Tab,
   justifyTabs = 'center',
   className,
-  onMouseOverTabBar,
-  onMouseLeaveTabBar
+  onMouseEnterTab,
+  onMouseLeaveTab
 }: TabBarProps<T>) => {
   const { t } = useTranslation()
 
@@ -53,8 +53,8 @@ const TabBar = <T extends string>({
               tabIndex={0}
               aria-selected={isActive}
               isActive={isActive}
-              onMouseEnter={() => onMouseOverTabBar?.(item.value)}
-              onMouseLeave={() => onMouseLeaveTabBar?.(item.value)}
+              onMouseEnter={() => onMouseEnterTab?.(item.value)}
+              onMouseLeave={() => onMouseLeaveTab?.(item.value)}
             >
               <TabLabel isActive={isActive}>
                 {item.Icon && (
--- a/apps/desktop-wallet/src/components/tabs/TabBar.tsx
+++ b/apps/desktop-wallet/src/components/tabs/TabBar.tsx
@@ -1,3 +1,4 @@
+import { motion } from 'framer-motion'
 import { ChevronRight } from 'lucide-react'
 import { useTranslation } from 'react-i18next'
 import styled, { css } from 'styled-components'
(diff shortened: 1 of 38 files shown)


apps/desktop-wallet/src/features/assetsLists/NFTsGrid.tsx
```diff
@@ -53,18 +58,59 @@ interface NFTsGridProps {
...
-        <Grid columns={columns}>
-          {isLoading ? <NFTsLoader /> : nftIds.map((nftId) => <NFTCard key={nftId} nftId={nftId} />)}
-        </Grid>
-      ))}
-  </motion.div>
-)
+const NFTsGrid = ({ columns, nftIds, isLoading, placeholderText }: NFTsGridProps) => {
+  const { data, fetchNextPage, hasNextPage } = useFetchPaginatedNFTs({ nftIds })
+
+  const ref = useRef(null)
```
mvaivre, line 64: Nit: I find it useful to name the ref, especially when it doesn't reference the main container component. Here it could be `gridBottomRef`

apps/desktop-wallet/src/storage/addresses/addressesSelectors.ts
```diff
@@ -30,9 +30,11 @@ export const makeSelectAddresses = () =>
 
 export const selectDefaultAddress = createSelector(
   selectAllAddresses,
-  (addresses) => addresses.find((address) => address.isDefault) || addresses[0]
+  (addresses) => addresses.find((address) => address.isDefault) || addresses.at(0)
```
mvaivre, line 33: Side note about this (don't remember if we talked about that already), but there's a lint rule to output a warning when we do things like addresses[0]. I tried it, but it's pretty strict / restrictive / dumb. Never thought about using `addresses.at(0)` because I didn't know it'd return either the type or `undefined`, nice finding!

apps/desktop-wallet/src/components/tabs/TabBar.tsx
```diff
@@ -48,6 +53,8 @@ const TabBar = <T extends string>({
               tabIndex={0}
               aria-selected={isActive}
               isActive={isActive}
+              onMouseEnter={() => onMouseOverTabBar?.(item.value)}
```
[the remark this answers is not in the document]
nop33: ok!

apps/desktop-wallet/src/features/assetsLists/NFTsGrid.tsx
nop33: ok!

apps/desktop-wallet/src/features/assetsLists/NFTsGrid.tsx
```diff
@@ -53,18 +58,59 @@ interface NFTsGridProps {
...
+    </motion.div>
+  )
+}
+
+interface UseFetchPaginatedNFTsProps {
+  nftIds: TokenId[]
+  pageSize?: number
+}
+
+const useFetchPaginatedNFTs = ({ nftIds, pageSize = PAGE_SIZE }: UseFetchPaginatedNFTsProps) =>
```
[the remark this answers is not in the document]
nop33: sure

apps/desktop-wallet/src/storage/addresses/addressesSelectors.ts
nop33: Yeah, I think the rule is the one that we currently have an open PR about and indeed it's very restrictive and we'd need to make a ton of changes.

apps/desktop-wallet/src/features/assetsLists/NFTsGrid.tsx
```diff
@@ -53,18 +58,59 @@ interface NFTsGridProps {
...
-  </motion.div>
-)
+const NFTsGrid = ({ columns, nftIds, isLoading, placeholderText }: NFTsGridProps) => {
+  const { data, fetchNextPage, hasNextPage } = useFetchPaginatedNFTs({ nftIds })
+
+  const ref = useRef(null)
+  const isInView = useInView(ref)
+
+  useEffect(() => {
+    if (isInView && hasNextPage) fetchNextPage()
```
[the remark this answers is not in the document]
mvaivre: Alrighty!

nop33:
@mvaivre all comments were addressed and the branch was rebased to solve the conflicts ✅ 

Re-requested review.

PR #1314, opened 2025-03-25
Move Tanstack and Redux code from DW to shared

nop33:
This PR moves a lot of code from DW to shared:
- API hooks
- API queries
- Types
- Redux slices:
  - Hidden tokens
  - Sent transactions
  - Addresses
  - Shared settings
    - fiat currency

The majority of the diff is:
- Import path changes
- Change from default to named exports (so that importing from shared works)

Next PR will finally focus on updating the MW code to use the code from shared.


mvaivre (approved):
Awesome conscientious work 👏

PR #1318, opened 2025-04-03
Store pubKey in Redux to fix Ledger unlocks

### LINKED ISSUE #1315: Cannot use Ledger in 3.0.1-rc.5
The issue is that the public key cannot be retried from keyring since ledger doesn't use keyring.

nop33:
Closes #1315


mvaivre (approved):
LGTM!

Issue #20, opened 2023-09-05
Add notification bubble to transfers tab & app icon


mvaivre:
Add a notification bubble with a number when:

- There are some pending motifs
- Or there was new incoming transactions since the last time the app was open.

The number of new incoming transactions could be shown in the App icon notification bubble.

nop33:
This was implemented. Closing issue.

PR #1322, opened 2025-04-07
Fix tx caching and add activity indicator

nop33:
While working on using Tanstack for fetching transaction data in the mobile wallet I had the idea of improving the UX in the desktop wallet, which led to killing 2 🐦 with 1 🪨 :

The problem was that 5 minutes after the user navigates away from the Activity screen the transactions cache data were deleted, because we want to avoid bloating the cache with useless data when addresses are deleted/created (the `queryKey` changes with changes in the list of addresses).

The solution is to keep an active instance of the same query that the Activity page is using, somewhere else. That's when I thought of implementing an Activity Indicator. This way, when addresses are deleted/created, the old cache will be deleted after 5 minutes, but the new cache will be persisted, which leads to the same behavior for the Activity page as for any other page of our app.

Related to:
- #20

### DIFF
--- a/.changeset/fruity-bars-sniff.md
+++ b/.changeset/fruity-bars-sniff.md
@@ -1,0 +1,5 @@
+---
+'alephium-desktop-wallet': patch
+---
+
+Display activity indicator when new transactions arrive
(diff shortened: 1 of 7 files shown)


mvaivre:
As I said on the call: really nice idea!

image

Let's go for this look:

image

And then do this to attract attention on new TXs: 
image

WDYT?

nop33:
I would avoid merging this before https://github.com/alephium/alephium-frontend/pull/1325 is merged since it touches code that has been moved to the shared packages.

PR #1326, opened 2025-04-10
WIP: MW Tanstack migration [5/5]

nop33:


apps/desktop-wallet/src/features/ledger/useInitializeAppWithLedgerData.tsx
```diff
@@ -1,4 +1,5 @@
 import { NonSensitiveAddressData } from '@alephium/keyring'
+import { walletUnlockedDesktop } from '@alephium/shared'
```
nop33, line 2: The reason of this naming will be explained below, at the definition of the action.

apps/desktop-wallet/src/modals/SettingsModal/EditWalletNameModal.tsx
```diff
@@ -1,4 +1,4 @@
-import { getHumanReadableError } from '@alephium/shared'
+import { ActiveWalletDesktop, getHumanReadableError } from '@alephium/shared'
```
nop33, line 1: Same for this one.

packages/shared/src/store/wallets/walletActionsDesktop.ts
nop33, line 1: As I mentioned the previous week on Slack:

> A tricky problem that I don’t have a very satisfactory solution yet for is how to share a redux slice between 2 apps by placing it in the shared package, but still be able to listen to actions that are project-specific. To give an example, the address slice is placed in shared so that all Tanstack hooks that use addresses can also be placed in shared. At the same time, the address slice should react to actions dispatched from the apps. However, the 2 wallets don’t necessarily share the same actions. The “walletUnlock” of mobile wallet is very different from the one in the dw, etc. I’ll figure it out, just wanted to give you a heads up




The solution I found is to move the app-specific actions (and corresponding types) to shared, but name them accordingly so that it's clear in which app they are used.

I am open to other ideas.

mvaivre (approved):
Code looks good! Shared the testing feedback on Slack.

PR #1328, opened 2025-04-15
Tanstack & Expo [6/8]: Upgrade to Expo 52

nop33:
🐛 Issues I discovered when testing:
- [x] ~~There are no addresses when importing a wallet unless the app is refreshed. Probably related to Tanstack?~~
- [x] ~~The screen titles are a bit faded on the top, most noticable on Android, but also ios~~
- [x] Glitch on Android when importing wallet: address discovery screen and biometrics activation screen flashes up and down non-stop. `KeyboardAvoidingView` must be the reason"
    - related: https://github.com/facebook/react-native/issues/45192
    - check if solved after this is merged and a new version released: https://github.com/facebook/react-native/pull/48181
    - I fixed it in 5ec732b4c27acf13d6c8fb9f60bfab435d7a3a49 by adding the `hasKeyboard` prop
- [ ] The onScroll callback of ScrollView and FlashList doesn’t not always fire on the first scroll leading to the header not being faded in properly.
- [x] ImportWalletAddressDiscoveryScreen:  The action 'GO_BACK' was not handled by any navigator.

✚ Bonus:
- [x] Upgrade to [React Navigation 7](https://reactnavigation.org/docs/upgrading-from-6.x/):
   - [x] Navigating to ActivityScreen after sending tx does not work
   - [x] After creating a new contact the screen freezes


apps/mobile-wallet/android/app/build.gradle
nop33, line 1: All the changes in the `/android` folder happened after I had to delete it and recreate it with `pnpm prebuild`.

apps/mobile-wallet/ios/Alephium.xcodeproj/project.pbxproj
nop33, line 1: I also had to delete the `/ios` folder and recreate it.

apps/mobile-wallet/src/components/animatedBackground/AnimatedBackground.tsx
nop33, line 1: This was a difficult bug to find because the error was not helping a lot. I manage to find the root cause (which was the wrapping of `SkiaCanvas` with `createAnimatedComponent`. This solution seems to fix the problem.

mvaivre:
>  The onScroll callback of ScrollView and FlashList doesn’t not always fire on the first scroll leading to the header not being faded in properly.

nop33:
I agree

nop33:
> I'm in favour of merging this

I'm merging this then.

PR #1332, opened 2025-04-24
Tanstack & Expo [8/8]: Replace AsyncStorage with MMKV

### LINKED ISSUE #1305: Consider migrating from AsyncStorage to MMKV
https://github.com/mrousavy/react-native-mmkv/blob/main/docs/MIGRATE_FROM_ASYNC_STORAGE.md

nop33:
- Closes #1305

### DIFF
--- a/apps/mobile-wallet/src/persistent-storage/storage.ts
+++ b/apps/mobile-wallet/src/persistent-storage/storage.ts
@@ -1,0 +1,39 @@
+import AsyncStorage from '@react-native-async-storage/async-storage'
+import { MMKV } from 'react-native-mmkv'
+
+export const storage = new MMKV({ id: 'mmkv.alephium-mobile-wallet' })
+
+// TODO: Delete after everyone has migrated to MMKV
+export const hasMigratedFromAsyncStorage = storage.getBoolean('hasMigratedFromAsyncStorage')
+
+// TODO: Delete after everyone has migrated to MMKV
+export async function migrateFromAsyncStorage(): Promise<void> {
+  console.log('Migrating from AsyncStorage -> MMKV...')
+  const start = global.performance.now()
+
+  const keys = await AsyncStorage.getAllKeys()
+
+  for (const key of keys) {
+    try {
+      const value = await AsyncStorage.getItem(key)
+
+      if (value != null) {
+        if (['true', 'false'].includes(value)) {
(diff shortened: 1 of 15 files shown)


apps/mobile-wallet/src/persistent-storage/storage.ts
nop33, line 1: Followed the migration guide: https://github.com/mrousavy/react-native-mmkv/blob/main/docs/MIGRATE_FROM_ASYNC_STORAGE.md

apps/mobile-wallet/src/persistent-storage/storage.ts
```diff
@@ -0,0 +1,39 @@
+import AsyncStorage from '@react-native-async-storage/async-storage'
+import { MMKV } from 'react-native-mmkv'
+
+export const storage = new MMKV({ id: 'mmkv.alephium-mobile-wallet' })
+
+// TODO: Delete after everyone has migrated to MMKV
+export const hasMigratedFromAsyncStorage = storage.getBoolean('hasMigratedFromAsyncStorage')
+
+// TODO: Delete after everyone has migrated to MMKV
+export async function migrateFromAsyncStorage(): Promise<void> {
```
mvaivre, line 10: I'm wondering if we should maybe keep a "backup" of the AsyncStorage for a while instead of removing everything, until we make sure most of users migrated safely. WDYT?

mvaivre (commented):
Looks good and seems functional!! Good job 👏 
Just one comment.

apps/mobile-wallet/src/persistent-storage/storage.ts
nop33: so you are just proposing to delete line 27

```ts
AsyncStorage.removeItem(key)
```

but keep line 35

```ts
  storage.set('hasMigratedFromAsyncStorage', true)
```

so that we don't accidentally re-migrate on every app launch?

mvaivre: Yeah exactly. So in case anything goes wrong, we still have the old pre-migration data available in asyncStorage.

nop33: Done ✅

mvaivre (approved):

PR #1337, opened 2025-05-15
Bump explorer to 1.10.12

mvaivre:


nop33 (approved):

nop33:
@mvaivre would you also like to fix the audit? Looking at the error: https://github.com/alephium/alephium-frontend/actions/runs/15048712604/job/42297816649?pr=1337

1. axios need to be updated
2. base-x is a dep of bip32 which is a dep of @alephium/web3-wallet and other packages, let's check if bip32 can be updated
3. I find this `@alephium/shared@link:../../packages/shared >` confusing... why `@link:`? Could that be a mistake from local linking of packages?
4. @babel/runtime: Dont' worry about it, it's fixed after updating react native (through expo)
5. @babel/helpers : same

PR #1340, opened 2025-05-19
Fix cache clearing

nop33:


apps/desktop-wallet/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # alephium-desktop-wallet
 
+## 3.0.1
```
nop33, line 3: Looks like we forgot to push the bump commit to `next`

mvaivre (approved):

PR #1352, opened 2025-05-28
Allow custom dApp URLs

nop33:
It's been requested again and again from our community, very useful for people who invite users to test their dApps on testnet, while the dapp is not yet on alph land.

- See: https://github.com/alephium/alephium-frontend/discussions/1266


mvaivre:
It's weird to me to tap on a random dapp to then be able to set the custom URL. I'll try to come up with an alternative solution :)

mvaivre (approved):
Good job! 
Added a couple of PRs on top (#1365 and #1366).

Issue #1338, opened 2025-05-16
Pre v2.X Tanstack & Expo migration release


nop33:
- [ ] The onScroll callback of ScrollView and FlashList doesn’t not always fire on the first scroll leading to the header not being faded in properly.

PR #1353, opened 2025-05-28
Fix input glitch in modals

nop33:
- Related to #1338


mvaivre (approved):
LGTM! Seems to work all fine!
Will keep an eye on inputs when we'll fix the remaining glitches and optimize.

PR #1356, opened 2025-06-02
Fix explorer prices

nop33:


apps/explorer/src/api/client.ts
nop33, line 1: code simply moved to the next file

apps/explorer/src/components/Amount.tsx
```diff
@@ -75,14 +75,15 @@ const Amount = ({
 
   const [integralPart, fractionalPart] = amount.split('.')
 
-  const RawAmountComponent = () => (
-    <>
-      <RawAmount data-tooltip-id="default" data-tooltip-content={convertToPositive(value as bigint).toString()}>
```
nop33, line 80: type casting was causing unexpected errors

apps/explorer/src/pages/AddressInfoPage/index.tsx
nop33, line 1: I took the opportunity to optimize the page rendering by moving code to subcomponents.

Issue #390, opened 2024-02-29
Group abstraction round 1 🥊 


mvaivre:
When users interact with dapps, sometimes they don’t have enough coins in group X, but they have enough in group Y. 
In this case, the wallet could try to do two transactions. With the first one move down coins from group Y to group X, and the second one do the target dApp transaction. 
If no address was derived in group Y, we derive the address first, then do the transaction.

nop33:
Agreed!

We still have this problem which will prob affect the solution of group abstraction: https://github.com/alephium/alephium-frontend/issues/36

PR #1357, opened 2025-06-02
Support groupless addresses and chained transactions across all apps

nop33:
- Closes https://github.com/alephium/alephium-frontend/issues/390


mvaivre (approved):
Smol lil PR. LGTM.

PR #1364, opened 2025-06-05
Improve desktop wallet CPU usage

nop33:


polarker:
IIRC, we fixed a CPU usage issue with animation - is this a new problem?

PR #1365, opened 2025-06-08
MW quick visual update

mvaivre:
Quick icons & font update on top of the custom dApp PR.


apps/mobile-wallet/android/app/src/main/assets/fonts/Geist-Bold.ttf
nop33, line 1: Are these really the icons that you like? :/

Issue #1367, opened 2025-06-10
Wrong transaction data right after confirmation


nop33:
The `getAddressesAddressTransactions` endpoint returns incomplete transaction data right after a transaction moves out of the mempool and becomes confirmed. The missing data are in the `inputs` of the transaction. A few moments later, the return results change and the `inputs` get updated. Here is an example diff: https://www.diffchecker.com/JxxgcNqi/

This creates problems in the UI:
1. We cannot calculate the amount diff of the transaction, since `inputs` are incomplete, leading to displaying wrong data.
2. The mechanism to reduce explorer backend API calls is affected: The `getAddressesAddressLatestTransaction` query does not include any info whether the tx `inputs` have been initialized or not. Since we only refetched wallet data when this endpoing returns a new hash, our UI is broken. We need to implement new logic for preventing refetching on a new hash and instead:
   1. Query `getAddressesAddressTransactions` with `{page: 1,limit: 1}`
   2. Check if all inputs have been initialized
   3. Refetch

An intermediate solution until the backend solves this issue is to display transactions with missing `input` properties as pending. This results in slower perceived transaction speed.

nop33:
Thomas is working on a fix for this, we should then remove the fix we added in #1368

PR #1368, opened 2025-06-10
Display incomplete confirmed txs as pending

nop33:
- Related to #1367

### DIFF
--- a/apps/explorer/src/pages/AddressPage/AddressTransactions.tsx
+++ b/apps/explorer/src/pages/AddressPage/AddressTransactions.tsx
@@ -1,7 +1,5 @@
 import { useFetchAddressTransactionsCount } from '@alephium/shared-react'
-import { MempoolTransaction } from '@alephium/web3/dist/src/api/api-explorer'
 import { keepPreviousData, useQuery } from '@tanstack/react-query'
-import { useEffect, useRef } from 'react'
 import { useTranslation } from 'react-i18next'
 import { usePageVisibility } from 'react-page-visibility'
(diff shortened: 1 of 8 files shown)


apps/explorer/src/pages/AddressPage/AddressTransactions.tsx
```diff
@@ -27,32 +25,22 @@ const AddressTransactions = ({ addressStr }: AddressTransactionsProps) => {
...
     ...queries.address.transactions.confirmed(addressStr, pageNumber, numberOfTxsPerPage),
-    enabled: !!addressStr,
+    refetchInterval,
     placeholderData: keepPreviousData
   })
 
   const { data: addressMempoolTransactions = [] } = useQuery({
     ...queries.address.transactions.mempool(addressStr),
-    refetchInterval: isAppVisible && pageNumber === 1 ? 10000 : undefined
+    refetchInterval
```
nop33, line 41: We need to poll both queries to avoid cases where a tx goes in and out of the mempool during the interval time (10s)

apps/explorer/src/pages/TransactionInfoPage.tsx
```diff
@@ -340,9 +337,6 @@ const TransactionInfoPage = () => {
   )
 }
 
-const isConfirmedTx = (tx?: Transaction | AcceptedTransaction | PendingTransaction): tx is Transaction =>
-  !!tx && (tx as Transaction).blockHash !== undefined
-
```
nop33, line 345: Deleted to reuse the shared one

PR #1376, opened 2025-06-12
Redirect to 404 page if address isn't correct

mvaivre:


apps/explorer/src/pages/AddressPage/AddressPage.tsx
```diff
@@ -28,10 +30,16 @@ const AddressPage = ({ addressStr }: AddressPageProps) => {
   const { t } = useTranslation()
   const theme = useTheme()
   const dispatch = useAppDispatch()
+  const navigate = useNavigate()
 
   useEffect(() => {
+    if (!isValidAddress(addressStr)) {
+      navigate('/404')
+      return
+    }
```
nop33, line 39: I would put this not in this component but in the one that instantiates it in the existing validation without the need of a useEffect, as I wrote in the previous comment,

apps/explorer/src/pages/AddressInfoPage/index.tsx
```diff
@@ -19,7 +19,7 @@ const AddressInfoPage = () => {
 
   if (!addressHash) {
     displaySnackbar({ text: t('The address format seems invalid'), type: 'alert' })
-    navigate('404')
+    navigate('/404')
   }
```
nop33, line 23:
```suggestion
  if (!addressHash || !isValidAddress(addressStr)) {
    displaySnackbar({ text: t('The address format seems invalid'), type: 'alert' })
    navigate('/404')
  }
```

mvaivre: Done

apps/explorer/src/pages/AddressPage/AddressPage.tsx
mvaivre: Done

PR #1379, opened 2025-06-16
Simplify modal dismissal

nop33:


apps/mobile-wallet/src/components/BiometricsWarningModal.tsx
```diff
@@ -29,7 +29,7 @@ const BiometricsWarningModal = memo<BiometricsWarningModalProps & ModalBaseProp>
   }
 
   return (
-    <BottomModal2 onDismiss={onDismiss} notScrollable modalId={id} title={`⚠️ ${t('Are you sure?')}`} noPadding>
+    <BottomModal2 notScrollable modalId={id} title={`⚠️ ${t('Are you sure?')}`} noPadding>
```
mvaivre, line 32: Awesome, already so much cleaner 👏 
The context could also contain the `modalId` no? 
`BottomModal2` would then be super clean, no need to pass down props anymore. We may even get rid of the `ModalBaseProp` type I think?

nop33: Let me have a look into it! Thanks for the feedback!

mvaivre (approved):
That's clean! ✨ 👏

PR #1396, opened 2025-06-24
Groupless 2

nop33:


apps/desktop-wallet/src/features/send/SendModal.tsx
nop33, line 1: This fine now only handles the normal send flow (by "normal" I mean the flow when the user clicks a "Send" button from the app UI) and doesn't handle the UI for WalletConnect requests anymore. The changes seen here are basically transfering the code from the `SendTransferModal` and its related files.

apps/desktop-wallet/src/modals/AppModals.tsx
nop33, line 1: Maybe we can look into implementing the same idea with the modal context in the desktop wallet as well, but for now it's not a priority imo

PR #1397, opened 2025-06-24
Fix scan modal opening, dry up useScreenOrModaIsFocused

mvaivre:
The scan modal isn't a "proper" modal (not stored in redux), hence the previous bug.

### DIFF
--- a/apps/mobile-wallet/src/features/qrCodeScan/CameraScanButtonBase.tsx
+++ b/apps/mobile-wallet/src/features/qrCodeScan/CameraScanButtonBase.tsx
@@ -10,7 +10,7 @@ import QRCodeScannerModal from '~/components/QRCodeScannerModal'
 import { useAppDispatch, useAppSelector } from '~/hooks/redux'
 import { cameraToggled } from '~/store/appSlice'
 import { showToast } from '~/utils/layout'
-import { useAppScreenIsFocused } from '~/utils/navigation'
+import { useScreenIsFocused } from '~/utils/navigation'
 
 interface CameraScanButtonBaseProps extends ButtonProps {
   onValidAddressScanned: (addressHash: AddressHash) => void
@@ -28,7 +28,7 @@ const CameraScanButtonBase = ({
   generateInvalidDataText,
   ...props
 }: CameraScanButtonBaseProps) => {
-  const isFocused = useAppScreenIsFocused()
+  const isFocused = useScreenIsFocused()
   const isCameraOpen = useAppSelector((s) => s.app.isCameraOpen)
   const dispatch = useAppDispatch()
   const { t } = useTranslation()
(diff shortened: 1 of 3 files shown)


apps/mobile-wallet/src/utils/navigation.ts
```diff
@@ -27,25 +26,12 @@ export const resetNavigation = (
   navigation.reset(getInitialNavigationState(initialRouteName))
 }
 
-export const useAppScreenIsFocused = () => {
-  const navigation = useNavigation()
+export const useScreenIsFocused = useIsFocused
+
+export const useScreenOrModaIsFocused = () => {
```
nop33, line 31: nit: rename to `useIsScreenOrModaFocused`

apps/mobile-wallet/src/utils/navigation.ts
```diff
@@ -27,25 +26,12 @@ export const resetNavigation = (
   navigation.reset(getInitialNavigationState(initialRouteName))
 }
 
-export const useAppScreenIsFocused = () => {
-  const navigation = useNavigation()
+export const useScreenIsFocused = useIsFocused
```
nop33, line 29: nit: rename: `useIsScreenFocused`

apps/mobile-wallet/src/features/qrCodeScan/CameraScanButtonBase.tsx
```diff
@@ -28,7 +28,7 @@ const CameraScanButtonBase = ({
   generateInvalidDataText,
   ...props
 }: CameraScanButtonBaseProps) => {
-  const isFocused = useAppScreenIsFocused()
+  const isFocused = useScreenIsFocused()
```
nop33, line 31: I think this component is also rendered inside a modal, so shouldn't we use the other hook?

apps/mobile-wallet/src/utils/navigation.ts
```diff
@@ -27,25 +26,12 @@ export const resetNavigation = (
...
+
+export const useScreenOrModaIsFocused = () => {
   const isAnyModalOpened = useAppSelector(selectIsAnyModalOpened)
   const isTopModal = useIsTopModal()
+  const isScreenFocused = useIsFocused()
```
nop33, line 34: I read some warnings in the docs about this hook, are we sure we need it instead of any other of the proposed solutions?

nop33: > Note that using this hook triggers a re-render for the component when the screen it's in changes focus. This might cause lags during the animation if your component is heavy. You might want to extract the expensive parts to separate components and use [React.memo](https://react.dev/reference/react/memo) or [React.PureComponent](https://react.dev/reference/react/PureComponent) to minimize re-renders for them.

https://reactnavigation.org/docs/use-is-focused/

nop33: Let's just roll with it

apps/mobile-wallet/src/utils/navigation.ts
nop33: Added in separate PR: https://github.com/alephium/alephium-frontend/pull/1399

PR #1406, opened 2025-06-30
WIP: Chained txs

nop33:


mvaivre (approved):
I love how you iterated by reusing the same logic you used before. Clean!
We start to have a lot of TX utilities though, but I guess it's not avoidable.

LGTM, great work!

PR #1418, opened 2025-07-09
Fix UTXO consolidation in MW and DW

nop33:
1. The bug in the mobile wallet is that we changed the props of the `buildSweepTransactions` to accept the hash instead of the public key, but since both are strings TypeScript didn't complained and we forgot to update the code.
2. The bug in the desktop wallet was introduced with the new modal system. Even though we were setting the transaction data to the component state, the callback of the consoildation modal was not being updated.


mvaivre (approved):
LGTM!

PR #1422, opened 2025-07-10
Groupless tests fixes

nop33:


packages/shared-react/src/api/queryClient.ts
```diff
@@ -2,7 +2,7 @@ import { FIVE_MINUTES_MS, ONE_MINUTE_MS } from '@alephium/shared'
 import { QueryClient, QueryClientConfig } from '@tanstack/react-query'
 import { AxiosError } from 'axios'
 
-const MAX_RETRIES = 3
+const MAX_RETRIES = 10
```
nop33, line 5: Since our API produces many 429's now, we want to avoid having incomplete data because all 3 retries have failed

packages/shared-react/src/api/queries/addressQueries.ts
```diff
@@ -105,7 +105,7 @@ export const addressTokensBalancesQuery = ({ addressHash, networkId, skip }: Add
 
 export const addressBalancesQuery = ({ addressHash, networkId, skip }: AddressLatestTransactionQueryProps) =>
   queryOptions({
-    queryKey: ['address', addressHash, 'balance', { networkId }],
+    queryKey: ['address', addressHash, 'level:1', 'balances-all', { networkId }],
```
mvaivre, line 108: I may need some explanation regarding those keys.

Two questions popping in my mind:
- How are those "levels" defined?
- Why are we sometimes using camel case (`searchString`, `tokensByType`), and other times dashes (`balances-all`, `tokens-search-strings`)?

packages/shared-react/src/api/queryInvalidation.ts
```diff
@@ -0,0 +1,14 @@
+import { queryClient } from '@/api/queryClient'
+
+// Queries need to be invalidated in order of dependency
+export const invalidateAddressQueries = async (address: string) => {
+  await queryClient.invalidateQueries({ queryKey: ['address', address, 'level:0'] })
+  await queryClient.invalidateQueries({ queryKey: ['address', address, 'level:1'] })
+  await queryClient.invalidateQueries({ queryKey: ['address', address, 'level:2'] })
+  await queryClient.invalidateQueries({ queryKey: ['address', address, 'level:3'] })
+  await queryClient.invalidateQueries({ queryKey: ['address', address, 'level:4'] })
+}
```
mvaivre, line 10: Could use a loop

```ts
export const invalidateAddressQueries = async (address: string) => {
  for (let level = 0; level <= 4; level++) {
    await queryClient.invalidateQueries({ queryKey: ['address', address, `level:${level}`] });
  }
}
```

mvaivre (commented):
My main question is about those invalidation levels. I'm wondering about maintainability, how to know what is in each level?

packages/shared-react/src/api/queries/addressQueries.ts
nop33: level:1 needs to be invalidated before level:2, because the queryFn of level:2 is calling queries that have level:1.

I did not find a better way to solve this. Using numbers creates some conceptual idea of dependency.

As for the camel case, I just didn't notice it. I'll fix that.

PR #1430, opened 2025-07-14
Fix amount calculation

mvaivre:
Fixes ##1429

### DIFF
--- a/packages/shared/src/utils/transactions.ts
+++ b/packages/shared/src/utils/transactions.ts
@@ -174,7 +174,7 @@ export const getTransactionAssetAmounts = (assetAmounts: AssetAmount[]) => {
     .map((asset) => ({ id: asset.id, amount: asset.amount.toString() }))
 
   const minAlphAmountRequirement = DUST_AMOUNT * BigInt(tokens.length)
-  const minDiff = minAlphAmountRequirement - alphAmount
+  const minDiff = alphAmount - minAlphAmountRequirement
   const totalAlphAmount = minDiff > 0 ? alphAmount + minDiff : alphAmount
 
   return {


packages/shared/src/utils/transactions.ts
```diff
@@ -174,7 +174,7 @@ export const getTransactionAssetAmounts = (assetAmounts: AssetAmount[]) => {
     .map((asset) => ({ id: asset.id, amount: asset.amount.toString() }))
 
   const minAlphAmountRequirement = DUST_AMOUNT * BigInt(tokens.length)
-  const minDiff = minAlphAmountRequirement - alphAmount
+  const minDiff = alphAmount - minAlphAmountRequirement
   const totalAlphAmount = minDiff > 0 ? alphAmount + minDiff : alphAmount
```
nop33, line 178: I don't think this change is correct.

Say the user selects to send:
- 1 USDT
- 1 EX

That means we have

```ts
const alphAmount = 0
const tokens = [{ id: 'USDT_ID', amount: 1 }, { id: 'EX_ID', amount: 1 }]

const minAlphAmountRequirement = 0.001 * 2 = 0.002
const minDiff = 0 - 0.002 = -0.002
const totalAlphAmount = 0
```

which is incorrect. We want the `totalAlphAmount` to be `0.002` in this case.

That's why it was set to:

```ts
const minDiff = minAlphAmountRequirement - alphAmount
```

which would result to 

```ts
const minDiff = 0.002 - 0 = 0.002
```

The commit that added this function to the shared package is this: 6461ba6b28│

This function was then removed from the mobile wallet and the desktop wallet. 

In the mobile wallet it looked like this, which is exactly the same as the current shared:

```ts
export const getTransactionAssetAmounts = (assetAmounts: AssetAmount[]) => {
  const alphAmount = assetAmounts.find((asset) => asset.id === ALPH.id)?.amount ?? BigInt(0)
  const tokens = assetAmounts
    .filter((asset): asset is Required<AssetAmount> => asset.id !== ALPH.id && asset.amount !== undefined)
    .map((asset) => ({ id: asset.id, amount: asset.amount.toString() }))

  const minAlphAmountRequirement = DUST_AMOUNT * BigInt(tokens.length)
  const minDiff = minAlphAmountRequirement - alphAmount
  const totalAlphAmount = minDiff > 0 ? alphAmount + minDiff : alphAmount

  return {
    attoAlphAmount: totalAlphAmount.toString(),
    tokens

  }
}
```

In the DW it looked a bit different. The difference was that the DW was adding the dust amount NO MATTER WHAT the user chose as the ALPH amount. In both cases though, the dust amount (or the diff between the dust amount and the user's selected ALPH) is added on top of the user's selected ALPH amount.

```ts
export const getTransactionAssetAmounts = (assetAmounts: AssetAmount[]) => {
  const alphAmount = assetAmounts.find((asset) => asset.id === ALPH.id)?.amount ?? BigInt(0)
  const tokens = assetAmounts
    .filter((asset): asset is Required<AssetAmount> => asset.id !== ALPH.id && asset.amount !== undefined)
    .map((asset) => ({ id: asset.id, amount: asset.amount.toString() }))

  const minAlphAmountRequirement = DUST_AMOUNT * BigInt(tokens.length)

  return {
    attoAlphAmount: (alphAmount + minAlphAmountRequirement).toString(),
    extraAlphForDust: minAlphAmountRequirement,
    tokens
  }
}
```

PR #1437, opened 2025-07-22
Fix transactionDetailsModal's TokenAmountsBox amounts

### LINKED ISSUE #1429: Amount details on DW are wrong
On DW, displayed amounts in the TX details modal are wrong for some TXs. Probably the case on MW too.
The culprit seems to be in `getTransactionAssetAmounts` that was recently updated.

mvaivre:
Fixes #1429

### DIFF
--- a/apps/desktop-wallet/src/components/TokenAmountsBox.tsx
+++ b/apps/desktop-wallet/src/components/TokenAmountsBox.tsx
@@ -24,45 +24,62 @@ import { useAppDispatch } from '@/hooks/redux'
 import { links } from '@/utils/links'
 import { openInWebBrowser } from '@/utils/misc'
 
-interface CheckAmountsBoxProps extends AmountBaseProps, BoxProps {
+interface TokenAmountsBoxProps extends AmountBaseProps, BoxProps {
   assetAmounts: AssetAmount[]
+  shouldAddAlphForDust?: boolean
   className?: string
 }
 
-const CheckAmountsBox = ({
+const TokenAmountsBox = ({
   assetAmounts,
   className,
   hasBg,
   hasHorizontalPadding,
   hasVerticalPadding,
   hasBorder,
+  shouldAddAlphForDust,
   ...props
-}: CheckAmountsBoxProps) => {
-  const userSpecifiedAlphAmount = assetAmounts.find((asset) => asset.id === ALPH.id)?.amount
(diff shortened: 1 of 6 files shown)


apps/desktop-wallet/src/components/TokenAmountsBox.tsx
```diff
@@ -24,45 +24,62 @@ import { useAppDispatch } from '@/hooks/redux'
...
+  let alphForDust = BigInt(0)
+
+  if (shouldAddAlphForDust) {
+    const userSpecifiedAlphAmount = assetAmounts.find((asset) => asset.id === ALPH.id)?.amount
+    const { attoAlphAmount, tokens, extraAlphForDust } = getTransactionAssetAmounts(assetAmounts)
+    const alphAsset = { id: ALPH.id, amount: attoAlphAmount }
+
+    alphForDust = extraAlphForDust
+    amounts = userSpecifiedAlphAmount ? [alphAsset, ...tokens] : [...tokens, alphAsset]
+  }
```
nop33, line 57: Let's put all that into a `useMemo` call, it feels wrong running all this logic on every re-render.

apps/desktop-wallet/src/features/transactionsDisplay/transactionDetailsModal/FTAmounts.tsx
```diff
@@ -11,7 +11,9 @@ const FTAmounts = ({ tx, refAddressHash }: TransactionDetailsModalTxProps) => {
 
   const isMoved = infoType === 'move'
 
-  return <CheckAmountsBox assetAmounts={fungibleTokens} highlight={!isMoved} showPlusMinus={!isMoved} />
+  const assetAmounts = fungibleTokens.map(({ id, amount }) => ({ id, amount }))
```
nop33, line 14: Is this rly necessary? Typescript shouldn't really complain since the props are already there.

apps/desktop-wallet/src/components/TokenAmountsBox.tsx
mvaivre: Done

apps/desktop-wallet/src/features/transactionsDisplay/transactionDetailsModal/FTAmounts.tsx
mvaivre: Removed

PR #1444, opened 2025-07-29
Groupless fixes

nop33:


packages/shared/src/utils/transactions.ts
```diff
@@ -92,7 +92,7 @@ export const getTransactionInfoType = (
       return 'swap'
     } else {
       const alphAmountReduced = alphAmount < 0 // tokenAmounts is checked in the swap condition
-      const isInternalTransfer = isInternalTx(tx, internalAddresses)
+      const isInternalTransfer = isWalletSelfTransfer(tx, internalAddresses)
```
nop33, line 95: this func will eventually become obsolete so I wouldn't bother renaming things in it for now.

packages/shared/src/transactions.ts
```diff
@@ -151,32 +151,31 @@ export const isSentTx = (
...
+  walletAddresses: AddressHash[]
+): boolean => getTxAddresses(tx).every((address) => address && walletAddresses.includes(getBaseAddressStr(address)))
 
 export const isContractTx = (tx: e.Transaction | e.PendingTransaction | e.MempoolTransaction): boolean =>
   !!tx.outputs?.some(inputOutputIsContractAddress) || !!tx.inputs?.some(inputOutputIsContractAddress)
 
 const inputOutputIsContractAddress = (io: e.Input | e.Output): boolean => !!io.address && isContractAddress(io.address)
 
+const getTxAddresses = (tx: e.Transaction | e.PendingTransaction | e.MempoolTransaction): (AddressHash | undefined)[] =>
+  [...(tx.outputs ?? []), ...(tx.inputs ?? [])].map((io) => io.address)
```
mvaivre, line 165: Should we dedupe this list? Or maybe add a param to let us do it? 🤔 
I'm wondering, because the result of this function is often re-mapped, like above at line 34 and 157. 
In line 157, we actually map twice, so duplicated addresses probably have an impact on perfs.

packages/shared/src/transactions.ts
```diff
@@ -154,15 +154,34 @@ export const isSentTx = (
...
 const inputOutputIsContractAddress = (io: e.Input | e.Output): boolean => !!io.address && isContractAddress(io.address)
 
-const getTxAddresses = (tx: e.Transaction | e.PendingTransaction | e.MempoolTransaction): (AddressHash | undefined)[] =>
-  [...(tx.outputs ?? []), ...(tx.inputs ?? [])].map((io) => io.address)
+const getTxAddresses = (tx: e.Transaction | e.PendingTransaction | e.MempoolTransaction): AddressHash[] => {
+  const addresses = new Set<AddressHash>()
+
+  if (tx.outputs) {
+    for (const output of tx.outputs) {
+      if (output.address && !addresses.has(output.address)) {
```
mvaivre, line 169: A `Set` ensures unique values by default, so I don't think this check is needed. If I'm not mistaken, I think this could be simplify quite a bit, and only use a single loop:

```ts
const getTxAddresses = (
  tx: e.Transaction | e.PendingTransaction | e.MempoolTransaction
): AddressHash[] => {
  const addresses = new Set<AddressHash>()
  const entries = [...(tx.outputs ?? []), ...(tx.inputs ?? [])]

  for (const entry of entries) {
    if (entry.address) {
      addresses.add(entry.address)
    }
  }

  return Array.from(addresses)
}

```

mvaivre (commented):
One more suggestion, otherwise LGTM!

packages/shared/src/transactions.ts
nop33: You are right, the check is not needed.

I want to avoid deconstructing and creating new objects unnecessarily. But I measured the difference and even though your version is slightly worse I think it's acceptable cause the difference is small. Updating...

packages/shared/src/transactions.ts
nop33: Discussed and resolved on slack.

packages/shared/src/transactions.ts
nop33: Pushed

PR #1478, opened 2025-08-25
Improve performance when sending txs

### LINKED ISSUE #1475: Fix performance issues when any address gets a new latest tx hash while wallet tx list has many page preloaded
## Problem

When a new transaction is sent or detected via polling, the latest tx hash of the involved addresses changes. This invalidates all address-related queries and the wallet transactions infinite query.

- If the user has previously scrolled deep into their activity history (pages 1…X), all those pages are stored in TanStack’s cache and persisted storage.
- Upon invalidation, all loaded pages are refetched.
- With many addresses (e.g. 10) and multiple loaded pages, this can result in dozens of requests at once (15–50+), causing major performance lags, especially on Android.
- The logic is correct (since each page shifts when new txs appear), but the behavior leads to excessive refetching, bloated cache/storage, and UI freezes.

nop33:
- Closes #1475

### DIFF
--- a/apps/mobile-wallet/src/features/ecosystem/dAppMessaging/DappBrowserContext.tsx
+++ b/apps/mobile-wallet/src/features/ecosystem/dAppMessaging/DappBrowserContext.tsx
@@ -456,7 +456,7 @@ export const DappBrowserContextProvider = ({ children, dAppUrl, dAppName }: Dapp
         })
       )
     },
-    [dAppUrl, dispatch, replyToDapp]
+    [dAppUrl, dispatch, replyToDapp, currentlyOnlineNetworkId]
   )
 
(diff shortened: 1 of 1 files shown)


apps/mobile-wallet/src/components/RefreshSpinner.tsx
```diff
@@ -18,3 +19,25 @@ const RefreshSpinner = (props: Partial<RefreshControlProps>) => {
 }
 
 export default RefreshSpinner
+
+const useRefreshAddressesBalances = () => {
```
mvaivre, line 23: Do you plan to leave it in the RefreshSpinner component? Or have you a masterplan and you want to move this later on?

PR #1527, opened 2025-09-26
Fix cache issues

nop33:
- Hopefully addresses https://github.com/alephium/alephium-frontend/issues/1518 which I didn't manage to reproduce

### DIFF
--- a/apps/mobile-wallet/package.json
+++ b/apps/mobile-wallet/package.json
@@ -24,7 +24,7 @@
     "@alephium/shared": "workspace:*",
     "@alephium/shared-crypto": "workspace:*",
     "@alephium/shared-react": "workspace:*",
-    "@alephium/token-list": "^0.0.20",
+    "@alephium/token-list": "^0.0.21",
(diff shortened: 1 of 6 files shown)


apps/mobile-wallet/package.json
```diff
@@ -97,7 +98,7 @@
     "react-hook-form": "^7.42.1",
     "react-i18next": "^13.2.1",
     "react-native": "0.76.9",
-    "react-native-aes-crypto": "^2.1.0",
+    "react-native-aes-crypto": "2.1.1",
```
nop33, line 101: I locked the version of this package for safety

apps/mobile-wallet/src/App.tsx
```diff
@@ -93,31 +93,50 @@ const AppContent = () => {
   return (
     <Provider store={store}>
       <PersistQueryClientContextProvider createPersister={createTanstackAsyncStoragePersister}>
-        <ApiContextProvider>
```
nop33, line 96: I moved this provider so that API requests fire only after the persisted cache has been restored.

PR #1591, opened 2026-03-17
MW Staking

mvaivre:
- Moved NFT FlashList to Dashboard, removed the NFT screen
- Refactored TabBar & tab content mgmt
- Init staking screen, built around the currently selected default address (for now)


packages/shared-react/src/features/dataPolling/usePendingTxPolling.ts
```diff
@@ -18,12 +18,22 @@ import { useCurrentlyOnlineNetworkId } from '../../network/useCurrentlyOnlineNet
...
   const dispatch = useSharedDispatch()
   const networkId = useCurrentlyOnlineNetworkId()
   const sentTx = useSharedSelector((s) => selectSentTransactionByHash(s, txHash))
   const allAddressHashes = useUnsortedAddressesHashes()
   const isExplorerOffline = useIsExplorerOffline()
+  const onConfirmedRef = useRef(options?.onConfirmed)
+
+  useEffect(() => {
+    onConfirmedRef.current = options?.onConfirmed
+  }, [options?.onConfirmed])
```
nop33: This feels unnecessary. The consumer of `usePendingTxPolling` should make sure `useCallback` is used for `onConfirmed`, if it's rerenders we worry about. Or why else do we need a ref here?

apps/mobile-wallet/metro.config.js
```diff
@@ -50,11 +50,11 @@ config.resolver.resolveRequest = (context, moduleName, platform) => {
 }
 
 // Polyfill Node core modules used by other deps (e.g. axios, crypto). Not needed for ws when aliased above.
-// fs: @alephium/web3's contract.ts has a dynamic import('fs') that Metro resolves statically.
 config.resolver.extraNodeModules = {
   ...config.resolver.extraNodeModules,
   crypto: require.resolve('react-native-quick-crypto'),
-  fs: path.resolve(__dirname, 'shims/fs.js')
+  stream: require.resolve('readable-stream'),
+  events: require.resolve('events')
```
nop33: wow, why are these needed? We just got rid of all of these polyfills. Which library needs these?

packages/shared-react/src/features/dataPolling/usePendingTxPolling.ts
mvaivre: useCallback only stabilizes until its deps change; when deps change, identity still changes, and it would retrigger polling effects. So a simple ref like that is a guard to make the hook robust even if one consumer forgets or can’t keep callback stable.

apps/mobile-wallet/metro.config.js
mvaivre: Mhh that's weird, I'm pretty sure I only deleted `fs`, and according to `blame` those were added in feb. https://github.com/alephium/alephium-frontend/blame/mw-staking/apps/mobile-wallet/metro.config.js

nop33: Then why do they appear in this diff? 😅

nop33: Solved in my PR

nop33 (approved):

Issue #51, opened 2023-10-18
Import watch-only accounts


nit27:
Some people would like to add watch-only accounts due to many purposes.  
### Example of another app is using:
> My Ether Wallet app (for iOS)
### Screenshot
![IMAGE 2023-10-18 11:10:09 PM](https://github.com/alephium/mobile-wallet/assets/95131051/51c4eb36-2566-4155-9c75-dc0fc4b2d04b)
![IMAGE 2023-10-18 11:18:23 PM](https://github.com/alephium/mobile-wallet/assets/95131051/16e5ea1a-5c2a-4e1c-8c62-aab0e6d8fcd6)
### Reference link 
https://www.mewwallet.com/

nit27:
can't wait for this feature 🔥

nop33:
Introduced in mobile wallet. Closing issue.

PR #1614, opened 2026-04-18
Add support for multiple wallets in mobile app

### LINKED ISSUE #24: Add support for multiple wallets
It was there before, but we wanted to make sure we store and secure wallets the right way. 

The main question is how do we efficiently secure all wallets with biometrics, without having to encrypt each wallet one by one.
The most sensible solution as of now would be to :

- First generate a random key, and use that random Key to encrypt the original wallet data “Atad = encrypt(Data, Key)”. 
- We store digest in regular (non secure) storage, and store the Key with biometrics encryption.
- To decrypt, we first unlock the key with biometrics, and load the Atad from regular storage. Then “Data = decrypt(Atad, key)”

This would allow us to work around the 2048 bytes limitation of the Secure Store.

For ref, here's a commit on master with the last state of multiple wallets support: https://github.com/alephium/mobile-wallet/commit/055e841e2b24ee23ab54b38cb419b6331355c89e

nop33:
- Closes #24
- Related to #51


apps/mobile-wallet/src/features/ecosystem/authorizedConnections/authorizedConnectionsSlice.ts
```diff
@@ -22,7 +19,7 @@ const initialState = connectionsAdapter.getInitialState()
 
 const authorizedConnectionsSlice = createSlice({
   name: sliceName,
-  initialState: connectionsAdapter.setAll(initialState, loadAuthorizedConnections()),
+  initialState,
```
mvaivre, line 25: Seems like we don't reload previous connection on app start anymore?

apps/mobile-wallet/src/features/ecosystem/authorizedConnections/authorizedConnectionsSlice.ts
```diff
@@ -54,8 +51,12 @@ export const authorizedConnectionsListenerMiddleware = createListenerMiddleware(
 
 authorizedConnectionsListenerMiddleware.startListening({
   matcher: isAnyOf(connectionAuthorized, connectionRemoved, connectionsCleared, hostConnectionRemoved, appReset),
-  effect: async (_, { getState }) =>
-    persistAuthorizedConnections(selectAllAuthorizedConnections(getState() as RootState))
+  effect: async (_, { getState }) => {
+    const rootState = getState() as RootState
+    const walletId = rootState.wallet.id
+
+    persistAuthorizedConnections(walletId, selectAllAuthorizedConnections(rootState))
```
mvaivre, line 64: Somehow linked to my previous comment: isn't there a risk that we end up persisting a stale state from another wallet when switching to another wallet (new `walletId`, old rootState?). Couldn't test it yet because of issues with my simulator / device.

apps/mobile-wallet/src/persistent-storage/wallet.ts
```diff
@@ -579,8 +360,324 @@ const generateAndStoreAddressKeypairForIndex = async (
...
+          // This shouldn't happen post-migration, but handle gracefully
+          sendAnalytics({ type: 'error', message: 'Post-migration wallet has unmigrated metadata' })
+        }
+
+        return 'valid'
+      }
+    }
+
+    // Wallet list exists but no valid wallet found - shouldn't happen
+    return 'valid'
```
mvaivre: Not sure it's related to this PR, but is it a good idea to return `valid` in that case?

mvaivre (commented):
Awesome initiative! 👏 
A couple of observations:
- Wallet content takes a while to load after wallet switch
- Let's make sure the wallet switch is the same size as adjacent buttons. Let's take the opportunity to increase the WC btn size as well (it's hard to press right now).

Image

nop33:
> Wallet content takes a while to load after wallet switch

- Make sure you're on mainnet, we don't persist cache on testnet.
- Also, keep in mind that the cache is persisted when the app goes to the bg
- If the 2 above are met, then the experience should be the same as when unlocking the wallet (cache gets restored)

apps/mobile-wallet/src/features/ecosystem/authorizedConnections/authorizedConnectionsSlice.ts
nop33: Correct, now that we don't have a single seed anymore, we need to first identify which is the wallet for which we have to load the authorized connections. Great catch, cause I forgot to move the initialization, I just removed it from here but did not add a reducer for wallet unlock and wallet switch. Added in a new commit. Thanks 🙏

apps/mobile-wallet/src/features/ecosystem/authorizedConnections/authorizedConnectionsSlice.ts
nop33: The way this middleware works is:
1. If any of the following actions gets dispatched: connectionAuthorized, connectionRemoved, connectionsCleared, hostConnectionRemoved, appReset
    1. Then persist data in storage

When we switch a wallet, this middleware does not get trigger.
Why then would there be a risk?

apps/mobile-wallet/src/persistent-storage/wallet.ts
nop33: You have a good point. Even though we are never supposed to end up to this scenario, it's not a great idea to assume a valid state. It won't be the first time that we end up to a scenario we can't explain. I'll change that to not swallow the invalid state and log things to our analytics. Thanks for this catch!

nop33: Fixed

nop33:
> Let's make sure the wallet switch is the same size as adjacent buttons. Let's take the opportunity to increase the WC btn size as well (it's hard to press right now).

Done!

nop33:
Conflicts solved ✅

apps/mobile-wallet/src/features/ecosystem/authorizedConnections/authorizedConnectionsSlice.ts
mvaivre: That's the problem I had in mind, `authorizedConnections ` isn't refreshed on wallet switch, so I wonder if the next "persistence-triggering" actions after switching to a new wallet will write on the previous wallet’s in-memory connections, under the new wallet’s storage key. But I guess you tested it so I'm probably wrong!

packages/shared-react/src/api/balanceUtils.ts
nop33, line 1: solving circular import warnings

apps/mobile-wallet/src/components/buttons/HighlightButton.tsx
nop33, line 1: solving console warnings

mvaivre (approved):

PR #1616, opened 2026-04-19
Add watch-only address

nop33:
- Related to #51


mvaivre (commented):
I don't know yet if this is related to this PR; but just had this bug while tapping on an address:

Image

mvaivre:
- We still see the main actions in a watch only address: 

image

---

I think this is because when you switch from a traditional `seed` wallet to a watch-only address, the `watch-only` type isn't properly updated in the store (check the payload here, type isn't included):

```ts
builder.addMatcher(
      isAnyOf(walletUnlockedMobile, newWalletGenerated, newWalletImportedWithMetadata, appLaunchedWithLastUsedWallet),
      (state, { payload: { name, id, isMnemonicBackedUp } }) => ({
        ...state,
        id,
        name,
        isMnemonicBackedUp
      })
```

---

- We also see the "address tab" which doesn't make sense anymore. 
- Same for the Scan and WC buttons. 
- In the settings, we should replace mentions of "wallet" by "watch-only address".

PR #1617, opened 2026-04-20
Improve staking actions UX

nop33:


apps/mobile-wallet/src/screens/Staking/UnstakingRequestItem.tsx
```diff
@@ -1,29 +1,50 @@
...
-const UnstakingRequestItem = ({ request }: UnstakingRequestItemProps) => {
+const UnstakingRequestItem = ({ request, addressHash }: UnstakingRequestItemProps) => {
   const { t } = useTranslation()
+  const dispatch = useAppDispatch()
   const { claimUnstaked, cancelUnstake } = useAlphStaking()
   const [isClaiming, setIsClaiming] = useState(false)
   const [isCancelling, setIsCancelling] = useState(false)
 
+  const vaultIndex = request.vaultIndex.toString()
+  const pendingVaultAction = useAppSelector((s) => s.staking.pendingVaultActions[vaultIndex])
```
mvaivre, line 36: Mhh, this is an edge case, but what happens if the user has multiple pending actions in the same vault index but for different addresses? 
They'll see the pending request corresponding to another staking address?

If yes, we should maybe add the `address` in the pending action, and filter by address in the UI.

nop33: Each unstake request has its own vault index I believe

nop33: I confirm that every unstake request has its own vaultIndex

image

mvaivre (approved):

PR #1621, opened 2026-04-22
Integrate staking APR

nop33:


apps/mobile-wallet/package.json
```diff
@@ -22,16 +22,18 @@
   },
   "dependencies": {
     "@alephium/keyring": "workspace:*",
-    "@alephium/powfi-sdk": "0.0.1-rc.24",
+    "@alephium/powfi-backend": "0.0.1-rc.16",
+    "@alephium/powfi-sdk": "npm:@eyettea/powfi-sdk@0.0.1-rc.32",
```
mvaivre: still using "your" sdk

apps/mobile-wallet/src/screens/Staking/StakingApr.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const region = useAppSelector((s) => s.settings.region)
+
+  const percentFormatter = useMemo(
+    () => new Intl.NumberFormat(region, { style: 'percent', maximumFractionDigits: 2 }),
+    [region]
+  )
+
+  if (isLoading) return <AmountSkeleton height={25} />
+
+  return <StakingAprStyled>{data?.apr ? percentFormatter.format(Number(data.apr) / 100) : '0.00%'}</StakingAprStyled>
```
mvaivre: Should we use '-' rather that 0% if value isn't available?

PR #1628, opened 2026-04-30
Fix wallet auto-connect in mobile in-app browser

nop33:
The in-app browser's auto-connect was broken by 3 issues:

1. The injected provider's `attach()` was deferred to the window load event, which fires after React's `useEffect`. By the time `attach()` ran, the dApp's auto-connect had already failed and given up.
2. Even with the provider available, `document.hasFocus()` returns `false` in React Native WebView during page load, causing `#checkTabFocused()` to throw "Unsolicited request" and abort the connection handshake.
3. Groupless addresses where not taken into consideration

Fix by calling `attach()` immediately after the UMD bundle and removing the `document.hasFocus()` check from `#unsafeEnable`. The check was designed to prevent background browser tabs from initiating connections (extension wallet), but `wallet-dapp-provider` is only used in the mobile in-app browser where the user explicitly opened the dApp.

### DIFF
--- a/.changeset/witty-trains-spend.md
+++ b/.changeset/witty-trains-spend.md
@@ -1,0 +1,5 @@
+---
+'@alephium/mobile-wallet': patch
+---
+
+Fix in-app browser auto-connect
--- a/apps/mobile-wallet/src/features/ecosystem/dAppMessaging/injectedJs.ts
+++ b/apps/mobile-wallet/src/features/ecosystem/dAppMessaging/injectedJs.ts
@@ -17,11 +17,9 @@ ${windowMessagePolyfill}
 
 ${alephiumProvider.code}
 
-window.addEventListener("load", () => {
-  if (typeof AlephiumWalletProvider !== 'undefined') {
-    AlephiumWalletProvider.attach();
-  }
-});
+if (typeof AlephiumWalletProvider !== 'undefined') {
+  AlephiumWalletProvider.attach();
+}
 
 // window.onerror = function(message, source, lineno, colno, error) {
 //   window.ReactNativeWebView.postMessage(JSON.stringify({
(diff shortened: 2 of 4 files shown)


.changeset/witty-trains-spend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@alephium/mobile-wallet': patch
+---
+
+Fix in-app browser auto-connect
```
tdroxler, line 5: `witty-trains-spend.md` 😅  
what's the usage of that file? I don't see any otherfile beside `config.json` in `.changeset`

apps/mobile-wallet/src/features/ecosystem/dAppMessaging/injectedJs.ts
```diff
@@ -36,7 +34,13 @@ window.addEventListener("load", () => {
...
 //     data: args
 //   }));
-//   return true;
+// };
+
+// window.console.error = function(...args) {
+//   window.ReactNativeWebView.postMessage(JSON.stringify({
+//     type: 'CONSOLE_ERROR',
+//     data: args.map(a => a instanceof Error ? a.message : a)
+//   }));
```
tdroxler, line 43: dead code

tdroxler (approved):
Minor comments, otherwise LGTM

.changeset/witty-trains-spend.md
nop33: It's a temporary file created by running the `changeset` tool. It's accommulates the changes while developing. Then, when we are ready to release a new version, we run `changeset version` and it compiles a CHANGELOG entry, and deletes these temp files.

apps/mobile-wallet/src/features/ecosystem/dAppMessaging/injectedJs.ts
nop33: yes, but useful for development. I'd like to keep it there.

PR #1629, opened 2026-04-30
Fix memory leaks and auto-connect

nop33:
- **Fix memory leak by removing event listeners on reject**
- **Fix in-app page navigation for Android devices**
- **Fix duplicate event handlers in in-app browser**
- **Wrap message handlers in try/catch to avoid stale state**
- **Add tests**
- **Remove unused ALPH_DISCONNECT_ACCOUNT**
- **Fix orphaned event listeners in Promise.race waitForMessage calls**
- **Fix infinite render loop in fund password feature**


apps/mobile-wallet/src/features/ecosystem/dAppMessaging/DappBrowserContext.tsx
nop33, line 1: just wrapping in try/catch

tdroxler (approved):
Thx for the nice commit history.
I must say I will approve because I haven't see anything weird and changes seem logical, but I'll have to trust you as I can't test the mobile wallet from a branch. (One day I'll update my stack to be able to)

nop33:
@tdroxler Thanks man! I can tell you when you can open your Google Play Store to update the app and test there, together with Diomark and CGI

PR #1630, opened 2026-04-30
Improve bottom modals dynamic height

nop33:


apps/mobile-wallet/src/features/modals/BottomModal.tsx
```diff
@@ -126,8 +132,6 @@ const BottomModal2 = <T,>(props: BottomModal2Props<T>) => {
 
 export default BottomModal2
 
-const SNAP_POINTS = ['50%', '100%']
```
tdroxler, line 129: why not keeping it as you are definiing twice `snapPoints: ['50%', '100%']` in this PR?
moving it in a sharable util?

apps/mobile-wallet/src/features/assetsDisplay/nftsDisplay/NftGridModal.tsx
```diff
@@ -11,7 +11,12 @@ interface NftGridModalProps {
 const NftGridModal = memo<NftGridModalProps>(({ nftsData }) => {
   const flashListProps = useNftsGridFlashListProps({ nfts: nftsData })
 
-  return <BottomModal2 flashListProps={flashListProps} />
+  return (
+    <BottomModal2
```
tdroxler: I know it's out of scope of this PR, but `BottomModal2` doesn't sounds very good, what does it do differently than `BottomModal`?

apps/mobile-wallet/src/features/modals/BottomModal.tsx
nop33: 1. Laziness
2. Allowing more customizable config
3. But really, laziness

apps/mobile-wallet/src/features/assetsDisplay/nftsDisplay/NftGridModal.tsx
nop33: Leftovers from the migration :D Will update, thanks!

apps/mobile-wallet/src/features/modals/BottomModal.tsx
tdroxler: Fine with it then :)

tdroxler (approved):

Issue #1622, opened 2026-04-22
Integrate swapping in mobile wallet


nop33:

nop33:
Closed by above PRs. Closing.

PR #1632, opened 2026-05-04
Add swap feature through Powfi redirect

nop33:
First iteration to implement the following issue:
- #1622

Next iteration will be a native swap support (similar to the staking feature)

### DIFF
--- a/.changeset/brave-tools-search.md
+++ b/.changeset/brave-tools-search.md
@@ -1,0 +1,5 @@
+---
+'@alephium/mobile-wallet': patch
+---
+
+Add basic swap support for testnet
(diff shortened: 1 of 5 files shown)


mvaivre:
All good! But it'd probably be better to ship the native integration directly. Better, cleaner UX. I imaging it wouldn't take us too long to implement, let's discuss when you're back!

nop33:
Isn't it better to allow the users already to swap, while we work on the native integration which can take up to a month to be released? True iterative releases :) My 2 cents!

mvaivre:
I guess! And it's true that the wallet connexion UX isn't that bad. 

One comment: for read-only addresses, let's remove all the call to action buttons:
image

nop33:
> One comment: for read-only addresses, let's remove all the call to action buttons:

Why remove "Receive" and "Buy"? They can be useful. No need for signing for any of these 2 operations.

I'll remove "Swap" since that requires signing.

nop33:
Done

PR #1641, opened 2026-06-16
Cleanup internal packages

nop33:
Less code -> Less attack surface

JS bundle sizes reduced across wallets:

| app | before | after | reduction | 
| desktop wallet |4.7MB | 4.2MB | `9.75%` |
| mobile wallet | 26.3MB | 25.6MB | `2.66%` |

Tested mobile wallet upgrade from version `1.0.7` (2 years old) to current ✅


apps/desktop-wallet/src/utils/misc.ts
```diff
@@ -25,17 +26,8 @@ export const openInWebBrowser = (url: string) => {
   }
 }
 
-export const stringToDoubleSHA256HexString = (data: string): string => {
```
tdroxler, line 28: so naming was wrong previously? it was not a double sha256, already a double sha512

apps/desktop-wallet/package.json
```diff
@@ -67,6 +67,7 @@
     "@ledgerhq/devices": "^8.13.0",
     "@ledgerhq/hw-transport-webhid": "^6.34.0",
     "@ledgerhq/hw-transport-webusb": "^6.33.0",
+    "@noble/hashes": "^1.8.0",
```
tdroxler, line 69: is it really safe? used by know crypto project?

CLAUDE.md
```diff
@@ -65,21 +65,23 @@ pnpm --filter @alephium/mobile-wallet exec jest path/to/file.test.ts
 
 ### Shared Packages
 
-- **`packages/shared`** — Core utilities, Redux store slices, types, constants, API clients. Built with tsup (ESM). Most packages depend on this
-- **`packages/shared-react`** — React hooks and components shared between apps. Built with tsup
-- **`packages/shared-crypto`** — Cryptographic utilities (BIP39, key derivation). Built with webpack into a single bundle (`dist/alephium.min.js`)
-- **`packages/keyring`** — Wallet key management. Built with tsup
-- **`packages/encryptor`** — Encryption utilities. Built with tsup
-- **`packages/wallet-dapp-provider`** — WalletConnect dApp provider. Built with Rollup
+- **`packages/shared`** — Core utilities, Redux store slices, types, constants, API clients. Ships raw TypeScript source (no build step). Most packages depend on this
```
tdroxler: Should we name this file rather `AGENTS.md` so we don't restrict to claude only? it seems a standard too. We could also use a symlink to keep them in sync:

```
Name your primary context file AGENTS.md at the project root. 
Then, create a relative symbolic link pointing CLAUDE.md to it.

ln -s AGENTS.md CLAUDE.md

Every major agent (like Codex) reads AGENTS.md by default, while Claude Code 
safely follows the symlink to read the exact same instructions.
```

tdroxler (approved):
Minor comments, otherwise LGTM

apps/desktop-wallet/src/utils/misc.ts
nop33: yes

apps/desktop-wallet/package.json
nop33: yes, we use it in web3 too, recommended by Hongchao

CLAUDE.md
nop33: good idea

nop33: done and pushed

PR #1653, opened 2026-07-11
Replace react-native-background-timer with AppState-based timestamp check

### LINKED ISSUE #1532: Replace `react-native-background-timer` with a better maintained package
When running

```
npx expo-doctor
```

we get this warning:

```
The following issues were found when validating your dependencies against React Native Directory:
  Untested on New Architecture: react-native-background-timer
  Unmaintained: react-native-background-timer
```

and from the looks of it this package hasn't been updated for 5 years. Metamask still uses it nevertheless

Labi-Joy:
Hi alephium team, i worked on this issue. Closes #1532.

Removes `react-native-background-timer` (unmaintained, flagged by `expo-doctor` as "Untested on New Architecture") and its `@types/react-native-background-timer` companion, and replaces the timer-driven auto-lock in `apps/mobile-wallet/src/features/auto-lock/useAutoLock.tsx` with an `AppState`-based timestamp check. On background we record `Date.now()` into a `useRef`; on foreground we compare against `autoLockSeconds` and dispatch `appBecameInactive` if the elapsed threshold was crossed. All existing branches are preserved: immediate lock when `autoLockSeconds` is `0`, no-op when it is `-1`, the camera-open guard, and the wallet-already-locked path.

One side effect worth flagging. With the old timer, on Android the lock could dispatch while the app was still backgrounded because the timer fired in-process; on iOS the OS suspended the timer within about 180s of backgrounding, so anything past that never locked until foreground anyway. The new code always locks on the next foreground transition once the threshold has elapsed. The user-visible experience of opening the app after a long time and seeing a locked wallet with a biometric prompt is identical. The only observable diff is for code that reads `state.wallet.isUnlocked` while the app is in background, for example handlers running from push notifications or WalletConnect socket callbacks; that state will now stay unlocked until foreground, whereas previously on Android it could flip to locked mid-background. I raised this exact scenario on the issue thread and am proceeding on the assumption it's acceptable since the foreground UX is unchanged.

No tests reference `useAutoLock` or `BackgroundTimer`. Ran `pnpm --filter @alephium/mobile-wallet lint` and `pnpm --filter @alephium/mobile-wallet exec tsc --noEmit`, both green. Changeset included, patch bump on `@alephium/mobile-wallet`.

Small note on `ios/Podfile.lock`: leaving as-is since the alephium workflow regenerates it via `pnpm prebuild` (commit `1e39eecba` did that most recently). The stale entries for `react-native-background-timer` will drop out on the next prebuild.


Labi-Joy:
Hi @nop33, gentle bump whenever you get a moment. This is the `AppState`-based timestamp approach discussed on #1532, with `backgroundedAt` in a `useRef` so nothing lives in module scope. CI is waiting on approval to run the workflow. Behavior notes plus the iOS >180s edge case i asked about are covered in the PR body, and i'm proceeding on the assumption that diff is acceptable since the foreground UX is unchanged. Thanks.

apps/mobile-wallet/src/features/auto-lock/useAutoLock.tsx
```diff
@@ -27,17 +25,18 @@ const useAutoLock = (unlockApp: () => Promise<void>) => {
...
+            backgroundedAt.current !== undefined && Date.now() - backgroundedAt.current >= autoLockSeconds * 1000
+          backgroundedAt.current = undefined
+
+          if (backgroundedForLong && isWalletUnlocked) {
+            dispatch(appBecameInactive())
+          }
 
-          if (!isWalletUnlocked && !isCameraOpen) {
+          if ((backgroundedForLong || !isWalletUnlocked) && !isCameraOpen) {
             unlockApp()
```
nop33: Blocking: the lock-then-unlock in the same tick leaves the app stuck

```ts
  if (backgroundedForLong && isWalletUnlocked) {
    dispatch(appBecameInactive())
  }
  if ((backgroundedForLong || !isWalletUnlocked) && !isCameraOpen) {
    unlockApp()
  }
```

  `unlockApp` opens with `if (isWalletUnlocked) return`, and it's a `useCallback` with `isWalletUnlocked` in its deps. Both calls above run in the same synchronous `AppState` callback, so React has not re-rendered yet. The `unlockApp` in scope is still the one captured while `isUnlocked === true`. It returns immediately and does nothing.

  So after a long background, the user gets: wallet locked in the store, no biometric prompt, and a black modal with the Alephium logo that they have to tap to authenticate. Worse, if usesBiometrics is off, that modal isn't rendered at all (visible requires `biometricsRequiredForAppAccess`), so the wallet UI stays mounted with `isUnlocked === false`. The effect re-runs after the re-render, but `isAppStateChangeCallbackRegistered` is already true, so `handleAppStateChange('active')` is never replayed. 

  The old code dodged this: the timer dispatched the lock while backgrounded, so by the time the app foregrounded, React had re-rendered and unlockApp was the fresh closure.

  The clean fix is to stop chaining the two calls: dispatch the lock on foreground, and let a separate effect keyed on `isWalletUnlocked === false` drive `unlockApp()`. That decouples them and removes the ordering assumption entirely.

apps/mobile-wallet/src/features/auto-lock/useAutoLock.tsx
```diff
@@ -27,17 +25,18 @@ const useAutoLock = (unlockApp: () => Promise<void>) => {
...
-              if (lockTimer) {
-                dispatch(appBecameInactive())
-              }
-            }, autoLockSeconds * 1000)
+            backgroundedAt.current = Date.now()
           }
         } else if (nextAppState === 'active') {
-          clearBackgroundTimer()
+          const backgroundedForLong =
+            backgroundedAt.current !== undefined && Date.now() - backgroundedAt.current >= autoLockSeconds * 1000
```
nop33: nit: Minor suggestion, let's protect ourselves against negative time in case the clock of the devices changes for whatever reason.

```suggestion
const elapsed = backgroundedAt.current !== undefined ? Date.now() - backgroundedAt.current : undefined
const backgroundedForLong = elapsed !== undefined && (elapsed < 0 || elapsed >= autoLockSeconds * 1000)
```

nop33 (commented):
Thanks for the contribution!

Don't worry about `ios/Podfile.lock`, I'll take care of it.

I have a few comments.

Labi-Joy:
Happy to help! Both addressed in a087c013f.

For the blocking one, you're right, that chained call was invoking a stale `unlockApp` closure. I moved `unlockApp` into a separate `useEffect` keyed on `appStateStatus`, `isWalletUnlocked`, and `isCameraOpen`, so it fires on the next render after the lock dispatch with a fresh closure. To keep that effect from firing while the app is backgrounded, I lifted the app-state tracking from a `useRef` into `useState`. That let me drop the `isAppStateChangeCallbackRegistered` flag and the synthetic `handleAppStateChange('active')` on mount too, since the new effect covers initial unlock through its deps naturally. I walked through each path (long background then foreground, wallet already locked at foreground, initial mount with locked wallet, `autoLockSeconds` -1, `autoLockSeconds` 0, camera-open) and the ordering works out in all of them.

For the nit, I applied your suggestion verbatim. Prettier reflowed the two lines into single-line ternaries but the logic is identical.

I ran lint and typecheck at the workspace root, both green.

nop33 (approved):
LGTM!

PR #1659, opened 2026-07-14
Wire React Query focus and online managers to React Native

### LINKED ISSUE #1655: Wire React Query focusManager and onlineManager to AppState and NetInfo
## Problem

React Query's `focusManager` and `onlineManager` are **never wired to React Native's `AppState` / NetInfo**. A grep for `focusManager|onlineManager` across `apps/mobile-wallet/src` and `packages/shared*` returns **zero hits**.

Both managers ship with **browser-only** defaults:

- `focusManager.isFocused()` falls back to `globalThis.document?.visibilityState !== 'hidden'`. In React Native `document` is `undefined`, so this evaluates to `undefined !== 'hidden'` → **permanently `true`**.
- `onlineManager` initialises `#online = true` and only ever updates via `window.addEventListener('online'|'offline')`, which never fires in RN → **permanently `true`**.

React Query therefore believes the app is **always focused and always online**, for the entire lifetime of the process.

## Consequences

1. **There is no refresh-on-foreground path at all.** `refetchOnWindowFocus` defaults to `true`, but it can never fire because no `visibilitychange` event exists in RN. When a user backgrounds the wallet and comes back to it, nothing triggers a refetch. They wait for the OS to release the suspended `setInterval` timers.

   This is the case users actually hit: *open the wallet to check if the money arrived*. Today that shows stale cached data until a 60s timer (that was suspended while backgrounded) happens to come around.

2. **`refetchOnReconnect` never fires.** Losing and regaining Wi-Fi / cell data does not trigger any refetch. The only network-recovery mechanism is the 2s retry loop in `packages/shared-react/src/useInitializeThrottledClient.ts:13,63` which flips `network.nodeStatus` / `explorerStatus` in Redux, which in turn un-`skipToken`s queries. That is an indirect side effect, not a deliberate reconnect strategy.

3. **`refetchInterval` timers are never deliberately suspended in the background.** React Query only pauses them when `focusManager` reports unfocused (and `refetchIntervalInBackground !== true`). Since it always reports focused, the intervals stay armed. They stop only because iOS happens to suspend the JS thread - an implicit, platform-dependent accident rather than a decision. **On Android the JS thread can keep running for a while, so polling requests can keep firing in the background**, burning battery and mobile data.

   Note the address list survives auto-lock (`appBecameInactive` only sets `wallet.isUnlocked = false` in `apps/mobile-wallet/src/store/wallet/walletSlice.ts:43-45`; the addresses slice is only reset on `walletLocked` / `walletSwitchedMobile` / `appReset`). So `useAddressesDataPolling` keeps its intervals armed even while the wallet is locked in the background.

This is also a **prerequisite** for any future transport work (batched polling, WebSockets, SSE): all of them need the app to have a correct notion of "am I in the foreground" and "am I online".

## Relevant code

- `packages/shared-react/src/api/queryClient.ts:14-25` - the `QueryClientConfig`. No `refetchOnWindowFocus` / `refetchOnReconnect` / `refetchIntervalInBackground` overrides.
- `packages/shared-react/src/features/dataPolling/useAddressesDataPolling.ts:16-29` - the per-address `refetchInterval` (60s frequent / 300s dormant) that should be suspended in the background.
- `apps/mobile-wallet/src/App.tsx:92` - where `useAddressesDataPolling()` is mounted.
- `apps/mobile-wallet/src/features/auto-lock/useAutoLock.tsx:24-55` - an existing `AppState` listener (auto-lock). **Do not** bolt this onto it; the managers should be registered once, outside React.
- `apps/mobile-wallet/src/features/persistQueryCache/usePersistQueryCacheOnBackground.ts:12-34` - the other existing `AppState` listener (cache persistence). Same note.

## Suggested implementation

`@react-native-community/netinfo` is **already a direct dependency** of `apps/mobile-wallet` (`11.5.2`), so no new package is needed.

Follow the [TanStack RN guide](https://tanstack.com/query/latest/docs/framework/react/react-native). Register both managers **once, at module scope** (not in a hook, not in a component), e.g. in a new `apps/mobile-wallet/src/api/reactQueryPlatform.ts` imported from `App.tsx` before the provider mounts:

```ts
import NetInfo from '@react-native-community/netinfo'
import { focusManager, onlineManager } from '@tanstack/react-query'
import { AppState, Platform } from 'react-native'

onlineManager.setEventListener((setOnline) =>
  NetInfo.addEventListener((state) => {
    setOnline(!!state.isConnected)
  })
)

AppState.addEventListener('change', (status) => {
  if (Platform.OS !== 'web') {
    focusManager.setFocused(status === 'active')
  }
})
```

Then decide the refetch policy explicitly in `queryClientConfig` rather than inheriting browser defaults. Suggested starting point:

- `refetchOnWindowFocus: true` - this is the whole point; foregrounding should refresh.
- `refetchOnReconnect: true`.
- `refetchIntervalInBackground: false` (the default) - now that `focusManager` is honest, this will genuinely suspend the 60s/300s pollers on background instead of relying on the OS to do it by accident.

### Watch out for

- **`staleTime` gates `refetchOnWindowFocus`.** Most of the wallet's data (balances, tokens, NFTs) is `staleTime: Infinity` and is **invalidation-driven**, not poll-driven - it only refetches when `addressLatestTransactionQuery` notices a new tx hash and fires the `invalidateAddressQueries` cascade (`packages/shared-react/src/api/queries/transactionQueries.ts:54-60`). So on foreground, the query that must go stale and refetch is **`addressLatestTransactionQuery`** (`staleTime: ONE_MINUTE_MS`). Verify that foregrounding after >1 min actually re-runs it and that the cascade fires. Foregrounding after <1 min will legitimately not refetch.
- **Don't double-fire on launch.** `focusManager.setFocused(true)` on the initial `AppState` event plus the query mount can cause a duplicate initial fetch. Check for a burst against the 10 req/s client throttle (`packages/shared/src/api/throttledClient.ts:8-11`), especially on wallets with many addresses.
- **Auto-lock interaction.** When the app is locked, we probably do *not* want to refetch on foreground before the user authenticates. Confirm the behaviour: `useAddressesDataPolling` stays mounted while locked because the addresses slice survives `appBecameInactive`.
- `Platform.OS !== 'web'` guard matters because `focusManager` already has a correct DOM implementation on web.

## Acceptance criteria

- [ ] Backgrounding the app suspends the address/price pollers (verify: no network requests fired while backgrounded, on **Android** in particular, where the JS thread survives).
- [ ] Foregrounding the app after >1 minute triggers an immediate `addressLatestTransactionQuery` refetch, and a new incoming tx cascades to fresh balances **without** the user pulling to refresh.
- [ ] Turning airplane mode off triggers a refetch (`refetchOnReconnect`).
- [ ] No duplicate request burst on cold start.
- [ ] Refetch policy is set **explicitly** in `queryClientConfig` rather than inherited from browser defaults.

## Context

Found while evaluating whether WebSockets/SSE are worth supporting (see `REALTIME_TRANSPORT_ANALYSIS.md`). The conclusion of that analysis is that **this fix improves the latency users actually perceive more than any push transport would**, because the dominant user journey is "open the wallet to check", not "sit in the open app and wait".

nop33:
- Fixes #1655

React Query's `focusManager`/`onlineManager` were never wired to React Native, so the app was permanently treated as focused and online. Consequences: no refresh when you reopen the app, no refetch on reconnect, and pollers that never suspend in the background (wasted battery and mobile data, especially on Android).

This registers both managers against `AppState`/NetInfo and states `refetchOnWindowFocus`/`refetchOnReconnect` explicitly.

**Reviewer notes**
- The one shared-react change (`queryClient.ts`) is a no-op for desktop and explorer; both were checked.
- `onlineManager` now pauses queries when NetInfo reports offline. It uses `isConnected`, not the probe-based `isInternetReachable`, so VPNs are the thing to watch; `networkMode: 'offlineFirst'` is available as optional insurance.

### DIFF
--- a/apps/mobile-wallet/__tests__/reactQueryPlatform.test.ts
+++ b/apps/mobile-wallet/__tests__/reactQueryPlatform.test.ts
@@ -1,0 +1,170 @@
+import '~/api/reactQueryPlatform'
+
+import { FREQUENT_ADDRESSES_TRANSACTIONS_REFRESH_INTERVAL, ONE_MINUTE_MS } from '@alephium/shared'
+import { throttledClient } from '@alephium/shared/api'
+import { addressLatestTransactionQuery, queryClientConfig } from '@alephium/shared-react'
+import { explorer as e } from '@alephium/web3'
+import { focusManager, onlineManager, QueryClient, QueryObserver } from '@tanstack/react-query'
+import { AppState, AppStateStatus } from 'react-native'
+
+// React Native defines a global `window` (react-native/Libraries/Core/setUpGlobals.js). Without it, React Query
+// considers itself to be running on a server and never schedules refetch intervals, which would make the polling
+// assertions below vacuous.
+const { netInfoListeners } = vi.hoisted(() => {
+  globalThis.window = globalThis as unknown as Window & typeof globalThis
+
+  return { netInfoListeners: [] as Array<(state: { isConnected: boolean | null }) => void> }
+})
+
+vi.mock('@react-native-community/netinfo', () => ({
+  default: {
+    addEventListener: (listener: (state: { isConnected: boolean | null }) => void) => {
+      netInfoListeners.push(listener)
+
+      return () => netInfoListeners.splice(netInfoListeners.indexOf(listener), 1)
+    }
+  }
+}))
+
+vi.mock('@alephium/shared/api', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('@alephium/shared/api')>()),
+  throttledClient: {
+    explorer: {
+      addresses: {
+        getAddressesAddressLatestTransaction: vi.fn()
+      }
+    }
+  }
+}))
+
+const ADDRESS_HASH = '1DrDyTr9RpRsQnDnXo2YRiPzPW4ooHX5LLoqXrqfMrpQH'
+
+const fetchLatestTransaction = vi.mocked(throttledClient.explorer.addresses.getAddressesAddressLatestTransaction)
+
+const emitAppStateChange = (status: AppStateStatus) => {
+  const calls = vi.mocked(AppState.addEventListener).mock.calls as unknown as Array<
+    [string, (status: AppStateStatus) => void]
+  >
+
+  calls.filter(([event]) => event === 'change').forEach(([, handler]) => handler(status))
+}
+
+const emitConnectivityChange = (isConnected: boolean) =>
+  netInfoListeners.forEach((listener) => listener({ isConnected }))
+
+// Mirrors what useAddressesDataPolling mounts: the latest transaction query of an address, optionally polled
+const withMountedObserver = async (refetchInterval: number | undefined, assert: () => Promise<void>) => {
+  vi.useFakeTimers()
+
+  const queryClient = new QueryClient(queryClientConfig)
+  queryClient.mount()
+
+  const observer = new QueryObserver(queryClient, {
+    ...addressLatestTransactionQuery({ addressHash: ADDRESS_HASH, networkId: 0, isExplorerOnline: true }),
+    refetchInterval,
+    notifyOnChangeProps: []
+  })
+  const unsubscribe = observer.subscribe(() => null)
+
+  try {
+    await vi.advanceTimersByTimeAsync(0)
+    expect(fetchLatestTransaction).toHaveBeenCalledTimes(1)
+
+    await assert()
+  } finally {
+    unsubscribe()
+    queryClient.unmount()
+    queryClient.clear()
+    vi.useRealTimers()
+  }
+}
+
+describe('reactQueryPlatform', () => {
+  beforeEach(() => {
+    focusManager.setFocused(true)
+    onlineManager.setOnline(true)
+    fetchLatestTransaction.mockReset()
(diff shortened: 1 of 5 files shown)


apps/mobile-wallet/__tests__/reactQueryPlatform.test.ts
```diff
@@ -0,0 +1,170 @@
...
+// React Native defines a global `window` (react-native/Libraries/Core/setUpGlobals.js). Without it, React Query
+// considers itself to be running on a server and never schedules refetch intervals, which would make the polling
+// assertions below vacuous.
+const { netInfoListeners } = vi.hoisted(() => {
+  globalThis.window = globalThis as unknown as Window & typeof globalThis
+
+  return { netInfoListeners: [] as Array<(state: { isConnected: boolean | null }) => void> }
+})
+
+vi.mock('@react-native-community/netinfo', () => ({
```
tdroxler, line 19: `vi` like the editor? 😆 😜

tdroxler (approved):

apps/mobile-wallet/__tests__/reactQueryPlatform.test.ts
nop33: vi like vitest :P

PR #1664, opened 2026-07-14
Make explorer homepage chart requests cacheable

### LINKED ISSUE #1656: Explorer homepage: 9 uncacheable requests every 10s, per viewer
## Problem

The explorer homepage fires **9 uncacheable HTTP requests every 10 seconds, per open tab**, to render data that changes at best once per block and at worst **once per day**.

The polling driver is a raw `setInterval`, `apps/explorer/src/pages/HomePage/HomePage.tsx:63-73`:

```tsx
useInterval(
  () => {
    fetchStatistics()
    if (currentPageNumber === 1) {
      getBlocks(currentPageNumber, false)
    }
  },
  10 * 1000,
  !isAppVisible
)
```

`fetchStatistics()` (`apps/explorer/src/pages/HomePage/useStatisticsData.ts:66-155`) fans out into **8 concurrent, unbatched requests** on every tick:

| # | Endpoint | Line |
|---|---|---|
| 1 | `charts/hashrates` (24h, hourly) | `useStatisticsData.ts:75-83` |
| 2 | `infos/heights` | `useStatisticsData.ts:85-92` |
| 3 | `infos/average-block-times` | `useStatisticsData.ts:94-98` |
| 4 | `charts/transactions-count` | `useStatisticsData.ts:100-121` |
| 5 | `charts/hashrates` (1 month) | `useStatisticsData.ts:123-146` |
| 6 | `infos/supply/total-alph` | `useStatisticsData.ts:152` |
| 7 | `infos/supply/circulating-alph` | `useStatisticsData.ts:153` |
| 8 | `infos/total-transactions` | `useStatisticsData.ts:154` |

Plus `GET /blocks?page=1&limit=10` (`useBlockListData.ts:20`) = **9 requests / 10s**.

### Why none of it can be cached

`getTimeIntervals()` recomputes `Date.now()` on **every single call** (`useStatisticsData.ts:31-40`), so `fromTs` / `toTs` are **different on every poll**. Every chart URL is therefore unique.

**A CDN or reverse proxy in front of explorer-backend cannot collapse or cache any of these.** Requests 1, 4 and 5 are the expensive chart-aggregation endpoints, and every single one is a guaranteed cache miss.

Meanwhile, on the default `Daily` interval, the tx-count and hashrate series **change once per day** - and we refetch them **8,640 times per day, per viewer**.

### The cost

| | |
|---|---|
| Per viewer | **0.9 req/s sustained** (54 req/min) |
| 1,000 concurrent viewers | **~900 req/s**, of which **~300 req/s** hit the uncacheable chart aggregations |
| Per day at that level | ~78M requests |

Every viewer receives **byte-identical** data. The useful origin work is `O(1)`; we are doing `O(viewers × 9)`.

The explorer is fully anonymous and public (no auth anywhere in `apps/explorer/src`), so this scales with traffic and with bots/crawlers, unbounded.

## Fix

1. **Bucket the chart timestamps instead of using `Date.now()`.** Round `fromTs`/`toTs` down to the interval granularity (e.g. the current hour for `Hourly`, the current day for `Daily`) in `getTimeIntervals()` (`useStatisticsData.ts:31-40`). Identical URLs across viewers and across polls → immediately cacheable and dedupable.
2. **Add `Cache-Control` to the explorer-backend responses** for `charts/*` and `infos/*` (a short TTL, e.g. 5-10s, is enough for the `infos/*` ones; the chart endpoints can take much longer once bucketed). Coordinate with the backend.
3. **Stop refetching the chart series on the 10s tick.** They change once per interval bucket, not every 10 seconds. Only the block list and the cheap `infos/*` scalars need the 10s cadence.
4. **Move the homepage onto React Query** so it gets request dedup, `staleTime`, and stale-while-revalidate for free. The homepage is currently the **only** part of the explorer that bypasses React Query entirely - it is hand-rolled `useState` + `useCallback` + `setInterval`, so the `QueryClient` defaults in `apps/explorer/src/App.tsx:44-60` (`staleTime: 10000`, `gcTime: ONE_DAY_MS`) **do not apply to the highest-traffic page in the app**. Tracked under #1657.

Items 1-3 are worth doing on their own and do not depend on item 4.

## Also worth fixing while in here

- `console.log('Fetching blocks...')` fires on every poll (`apps/explorer/src/api/useBlockListData.ts:11`).
- Pagination mismatch: `getBlocks` requests `limit: 10` (`useBlockListData.ts:20`) but `PageSwitch` is told `elementsPerPage={8}` (`HomePage.tsx:201`).
- On tab refocus, `setInterval` does not fire immediately, so the user stares at stale data for a full 10s. `refetchOnWindowFocus: false` (`App.tsx:45`) means React Query won't rescue it either. Moving to React Query with `refetchOnWindowFocus: true` fixes this.

## Acceptance criteria

- [ ] Two viewers loading the homepage at the same time issue **identical** chart URLs.
- [ ] Chart requests are not re-issued on every 10s tick.
- [ ] Chart and `infos/*` responses carry a `Cache-Control` header and are served from cache on repeat.
- [ ] Homepage request rate per viewer drops substantially from the current 0.9 req/s.

## Context

Found while evaluating whether WebSockets/SSE are worth supporting (see `REALTIME_TRANSPORT_ANALYSIS.md`). Conclusion there: **this is the single biggest cost win available in the explorer, and it needs no new protocol.** It is strictly cheaper than pushing the same data over a socket, and unlike persistent connections it degrades gracefully under scrapers.

nop33:
- Fixes #1656

The homepage fired 9 requests every 10s per viewer, and `getTimeIntervals` used `Date.now()` on every poll, so every chart URL was unique and uncacheable. At ~1k viewers that is ~900 req/s of origin work serving byte-identical data.

Fix: floor the chart timestamps to a 60s bucket, so every viewer within the same minute emits an identical URL. An edge cache then collapses them to one origin request per minute regardless of viewer count, and nothing is ever more than ~70s stale. Also drops a debug log and fixes a pagination off-by-two.

Note that the edge cache only kicks in once explorer-backend adds `Cache-Control` to these endpoints (tracked in alephium/explorer-backend#670). This PR makes the URLs cacheable; it does not add the caching itself.

### DIFF
--- a/apps/explorer/src/components/Table/TableRow.tsx
+++ b/apps/explorer/src/components/Table/TableRow.tsx
@@ -1,17 +1,27 @@
 import { motion } from 'framer-motion'
 import { Children } from 'react'
 import { Link } from 'react-router-dom'
-import styled from 'styled-components'
+import styled, { css, keyframes } from 'styled-components'
 
 import { deviceBreakPoints } from '@/styles/globalStyles'
 
 interface RowProps {
   isActive?: boolean
   highlight?: boolean
+  isNew?: boolean
   onClick?: React.MouseEventHandler<HTMLTableRowElement>
   linkTo?: string
   className?: string
 }
+
+const newRowFlash = keyframes`
+  from {
+    background-color: var(--new-row-flash-color);
+  }
+  to {
+    background-color: transparent;
+  }
+`
 
 const rowVariants = {
   hidden: { opacity: 0 },
@@ -36,9 +46,18 @@ const TableRow: FC<RowProps> = ({ children, onClick, linkTo, className }) => (
 )
 
 export default styled(TableRow)`
+  --new-row-flash-color: ${({ theme }) => theme.bg.accent};
   background-color: ${({ theme, isActive }) => (isActive ? theme.bg.primary : '')};
   border: none;
   cursor: ${({ onClick }) => (onClick ? 'pointer' : 'auto')};
+
+  ${({ isNew }) =>
+    isNew &&
+    css`
+      @media (prefers-reduced-motion: no-preference) {
+        animation: ${newRowFlash} 1.6s ease-out;
+      }
+    `}
 
   td:first-child {
     display: flex;
--- a/.changeset/lucky-bats-cache.md
+++ b/.changeset/lucky-bats-cache.md
@@ -1,0 +1,5 @@
+---
+'@alephium/explorer': patch
+---
+
+Reduce the number of requests the homepage makes and let its chart data be cached
--- a/.changeset/warm-lions-blink.md
+++ b/.changeset/warm-lions-blink.md
@@ -1,0 +1,5 @@
+---
+'@alephium/explorer': patch
+---
+
+Refresh the homepage's latest blocks every 5s and briefly highlight newly-arrived ones
--- a/apps/explorer/src/pages/HomePage/HomePage.tsx
+++ b/apps/explorer/src/pages/HomePage/HomePage.tsx
@@ -28,7 +28,7 @@ import { useWindowSize } from '@/hooks/useWindowSize'
 import { deviceBreakPoints, deviceSizes } from '@/styles/globalStyles'
 import { formatNumberForDisplay } from '@/utils/strings'
 
-import useBlockListData, { BLOCKS_PER_PAGE } from './useBlockListData'
+import useBlockListData, { BLOCKS_PER_PAGE, BLOCKS_REFRESH_INTERVAL } from './useBlockListData'
 import useStatisticsData from './useStatisticsData'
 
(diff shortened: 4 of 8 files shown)


apps/explorer/src/components/Table/TableRow.tsx
```diff
@@ -1,18 +1,28 @@
...
   highlight?: boolean
+  isNew?: boolean
   onClick?: React.MouseEventHandler<HTMLTableRowElement>
   linkTo?: string
   className?: string
 }
 
+const newRowFlash = keyframes`
+  from {
+    background-color: var(--new-row-flash-color);
```
tdroxler, line 19: niiiiice can't wait to see that flash thing, good idea

tdroxler (approved):

PR #1689, opened 2026-07-23
Integrate creation of Schnorr addresses in desktop and mobile wallets

nop33:


packages/shared/src/types/addresses.ts
```diff
@@ -48,3 +48,7 @@ export type GrouplessAddress = AddressBase & {
 export type Address = AddressWithGroup | GrouplessAddress
 
 export type AddressesState = EntityState<Address>
+
+// The user-facing address kind chosen when creating an address, mapped to a key type at generation
+// time: 'groupless' -> gl-secp256k1, 'grouped' -> default, 'schnorr' -> bip340-schnorr.
+export type NewAddressType = 'groupless' | 'grouped' | 'schnorr'
```
h0ngcha0, line 54: We might need to consider key type [here](https://github.com/alephium/alephium-frontend/blob/alephium-frontend_schnorr/apps/mobile-wallet/src/features/ecosystem/modals/SignMessageTxModal.tsx#L35) and [here](https://github.com/alephium/alephium-frontend/blob/alephium-frontend_schnorr/packages/shared/src/transactions/transactionParams.ts#L137) ?

nop33: Good points! Pushed a fix!

nop33:
Fixed conflicts

apps/mobile-wallet/src/signer.ts
```diff
@@ -14,8 +14,12 @@ export class SecureStoreSigner extends AlephiumWalletSigner {
   public getPublicKey = async (addressStr: string): Promise<string> =>
     getAddressAsymetricKey(this.getWalletId(), getBaseAddressStr(addressStr), 'public')
 
-  public signRaw = async (addressStr: string, tx: string): Promise<string> =>
-    transactionSign(tx, await getAddressAsymetricKey(this.getWalletId(), getBaseAddressStr(addressStr), 'private'))
+  public signRaw = async (addressStr: string, tx: string): Promise<string> => {
+    const baseAddressStr = getBaseAddressStr(addressStr)
+    const keyType = store.getState().addresses.entities[baseAddressStr]?.keyType
```
h0ngcha0: When `keyType` is undefined here it would default to `default`. Now when we have Schnorr it might produce the wrong signature. Perhaps we should throw when keyType is undefined here.

nop33: Good idea to throw, although we should never have the case that the address won't exist in the Redux store. But in case of corrupted data for whatever reason, it's a good idea. Fixed

h0ngcha0 (approved):
LGTM 👍

PR #1715, opened 2026-08-03
Fix transactions loading

nop33:
The main goal of this PR is to fix a bug that exists in production that was revealed during the lower rate limiting phase: the transactions infinite queries were using a `Promise.all` with the list of all addresses. When the list of address was large, and at least 1 of them would get rate limited, the whole list would be retried. This PR fixes this issue by not retrying requests that have already returned successfully.

While fixing this I discovered a few more bugs and improvements which I addressed. Each commit is clean.


h0ngcha0 (approved):
LGTM 👍

PR #1716, opened 2026-08-03
Fix turborepo config & CI workflows

nop33:
After the upgrade of Turborepo from v1 to v2 some CI configuration was not updated, specifically the path to the turbo cache had changed. This PR fixes that.

In the process another bug was discovered around built outputs of the desktop wallet which were not included in the cache. That's fixed too.

I also collapsed some checks into one job to avoid setting up the monorepo, and installing deps many times in parallel. It's pretty fast already so no need for parallelization.


h0ngcha0 (approved):
LGTM 👍

